PR #55, opened 2018-04-19
Have CI do a fmt as well

Hoverbear:
Currently we're not using `rustfmt` via CI for Raft, and we should be! This configures it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 env:
   global:
     - RUST_BACKTRACE=1
-    - CLIPPY_VERSION=0.0.179
+    - CLIPPY_VERSION=0.0.192
     - RUSTFLAGS="-D warnings"
 cache: cargo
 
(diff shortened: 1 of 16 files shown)


Hoverbear:
This is failing due to the Rust version being too old to have the `rustfmt-preview` component. I'll do this in another PR.

Hoverbear:
Based this off #56 so it will be green. Please merge #56 first though.

.travis.yml
```diff
@@ -3,21 +3,26 @@ sudo: false
...
+  # This build uses stable and checks rustfmt (clippy is not on stable).
   - rust: stable
-  - rust: nightly-2018-01-12
+    install:
+      - rustup component add rustfmt-preview
+  # This build uses the nightly used by TiKV and checks both rustfmt & clippy.
+  - rust: nightly-2018-04-06
     install:
       - export PATH="$PATH:$HOME/.cargo/bin"
+      - rustup component add rustfmt-preview
```
BusyJay: What if the format is different from stable channel?

Hoverbear: :+1: Great catch, you're totally right. We should just run `rustfmt` on `stable`.

BusyJay:
LGTM

PR #58, opened 2018-05-09
Learner needs to respond vote requests.

### LINKED ISSUE #57: Learner should respond to request vote but don't accept votes
Consider following case, a cluster has three nodes A, B, C, and A is Leader. And then add learner D, E to the cluster. D, E is promoted to voter then, but D is isolated from A, so D doesn't know himself is promoted. When A, B both crash, the cluster still has three voters C, D, E, which should be able to form a quorum. However, D can't vote in the current implementation, since it doesn't know it's not learner anymore.

One solution to this problem is let learner respond to vote request and let candidate check if the response is from a valid peer. Because we won't let a voter step back as learner, so when candidate consider a node is a voter, then it's definitely a voter from that point to infinite future. So the candidate check should be safe. In the situation described above, the quorum can be formed again, and the cluster can recover automatically.

/cc @xiang90 @siddontang any thoughts?

hicqu:
So that if some voters fail, rest voters and learners can still form a quorum. Fix #57 .

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -27,16 +27,16 @@
 
 use std::cmp;
 
-use rand::{self, Rng};
 use eraftpb::{Entry, EntryType, HardState, Message, MessageType, Snapshot};
 use fxhash::FxHashMap;
 use protobuf::repeated::RepeatedField;
+use rand::{self, Rng};
 
-use super::storage::Storage;
-use super::progress::{Inflights, Progress, ProgressSet, ProgressState};
 use super::errors::{Error, Result, StorageError};
+use super::progress::{Inflights, Progress, ProgressSet, ProgressState};
 use super::raft_log::{self, RaftLog};
 use super::read_only::{ReadOnly, ReadOnlyOption, ReadState};
+use super::storage::Storage;
 
 // CAMPAIGN_PRE_ELECTION represents the first phase of a normal election when
 // Config.pre_vote is true.
@@ -828,6 +828,7 @@ impl<T: Storage> Raft<T> {
             return;
         }
 
+        // Only send vote request to voters.
         let prs = self.take_prs();
         prs.voters()
             .keys()
@@ -1016,24 +1017,6 @@ impl<T: Storage> Raft<T> {
                 debug!("{} ignoring MsgHup because already leader", self.tag);
             },
             MessageType::MsgRequestVote | MessageType::MsgRequestPreVote => {
-                if self.is_learner {
-                    // TODO: learner may need to vote, in case of node down when confchange.
-                    info!(
-                        "{} [logterm: {}, index: {}, vote: {}] ignored {:?} from {} \
-                         [logterm: {}, index: {}] at term {}: learner can not vote",
-                        self.tag,
-                        self.raft_log.last_term(),
-                        self.raft_log.last_index(),
-                        self.vote,
-                        m.get_msg_type(),
-                        m.get_from(),
-                        m.get_log_term(),
-                        m.get_index(),
-                        self.term,
-                    );
-                    return Ok(());
-                }
-
                 // We can vote if this is a repeat of a vote we've already cast...
                 let can_vote = (self.vote == m.get_from()) ||
                     // ...we haven't voted and we don't think there's a leader yet in this term...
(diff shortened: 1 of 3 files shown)


hicqu:
@BusyJay @Hoverbear @overvenus PTAL thanks.

src/raft.rs
```diff
@@ -1083,9 +1066,10 @@ impl<T: Storage> Raft<T> {
 
     fn log_vote_approve(&self, m: &Message) {
         info!(
-            "{} [logterm: {}, index: {}, vote: {}] cast {:?} for {} [logterm: {}, index: {}] \
+            "{}({}) [logterm: {}, index: {}, vote: {}] cast {:?} for {} [logterm: {}, index: {}] \
              at term {}",
             self.tag,
+            if self.is_learner { "learner" } else { "voter" },
```
BusyJay: This is redundant.

tests/cases/test_raft.rs
```diff
@@ -3978,3 +3949,47 @@ fn test_remove_learner() {
...
+
+    let timeout = n1.get_election_timeout();
+
+    let do_campaign = |nw: &mut Network| {
+        for _ in 0..timeout << 1 {
+            nw.peers.get_mut(&1).unwrap().tick();
+            nw.peers.get_mut(&3).unwrap().tick();
+        }
+
+        // MsgRequeestVote should only come from 1.
```
Hoverbear: Too many `e`s in `MsgRequeestVote`

Hoverbear:
@hicqu I see that in https://github.com/pingcap/raft-rs/issues/57#issue-321442130 a candidate check was written about. In this PR I don't see an added candidate check. Was this already implemented, or found to be unnecessary?

hicqu:
@Hoverbear I  omitted the check, because voters only send `MsgRequetsVote` to known voters. So if a voter receive a `MsgRequestVoteResponse`, it must come from a peer in the voter's `voter_prs`, not `learner_prs`.

tests/cases/test_raft.rs
```diff
@@ -3978,3 +3949,47 @@ fn test_remove_learner() {
...
+        }
+
+        // MsgRequestVote should only come from 1.
+        let msgs = read_messages(nw.peers.get_mut(&1).unwrap());
+        msgs.iter().for_each(|m| assert_eq!(m.get_from(), 1));
+        nw.send(msgs);
+    };
+
+    let mut network = Network::new(vec![Some(n1), None, Some(n3)]);
+    network.isolate(2);
```
BusyJay: `None` means it won't respond to requests already.

tests/cases/test_raft.rs
```diff
@@ -3978,3 +3949,47 @@ fn test_remove_learner() {
...
+    let mut n3 = new_test_learner_raft(3, vec![1, 2], vec![3], 10, 1, new_storage());
+    n3.become_follower(1, INVALID_ID);
+    n3.reset_randomized_election_timeout();
+
+    let timeout = n1.get_election_timeout();
+
+    let do_campaign = |nw: &mut Network| {
+        for _ in 0..timeout << 1 {
+            nw.peers.get_mut(&1).unwrap().tick();
+            nw.peers.get_mut(&3).unwrap().tick();
```
BusyJay: What's the point of ticking node 3?

tests/cases/test_raft.rs
```diff
@@ -3978,3 +3949,47 @@ fn test_remove_learner() {
...
+        }
+
+        // MsgRequestVote should only come from 1.
+        let msgs = read_messages(nw.peers.get_mut(&1).unwrap());
```
BusyJay: Why not send a `MsgHup` to node 1 directly?

tests/cases/test_raft.rs
```diff
@@ -3978,3 +3949,47 @@ fn test_remove_learner() {
...
+    };
+
+    let mut network = Network::new(vec![Some(n1), None, Some(n3)]);
+    network.isolate(2);
+
+    // Can't elect new leader because 1 won't send MsgRequestVote to 3.
+    do_campaign(&mut network);
+    assert_eq!(network.peers[&1].state, StateRole::Candidate);
+
+    match network.peers.get_mut(&1) {
```
BusyJay: `unwrap` is OK.

tests/cases/test_raft.rs
hicqu: To show all `Vote` messages come from store 1 even if 3 will also `step(MsgHup)`.

BusyJay: I don't get it. Where are all those messages showed? How does it help with the test case?

BusyJay: No, they are still here.

BusyJay:
LGTM

PR #66, opened 2018-05-28
raft: fix quorum math for abnormal sizes

lucab:
This updates quorum math for cluster of abnormal size (< 3 members).
It also adds a test to ensure that quorum number is bound between 1 and
the size of the cluster.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -317,7 +317,10 @@ pub fn vote_resp_msg_type(t: MessageType) -> MessageType {
 
 // Calculate the quorum of a Raft cluster with the specified total nodes.
 pub fn quorum(total: usize) -> usize {
-    total / 2 + 1
+    match total {
+        0...2 => total,
+        _ => total / 2 + 1,
+    }
 }
 
 impl<T: Storage> Raft<T> {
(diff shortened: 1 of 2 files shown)


src/raft.rs
```diff
@@ -317,7 +317,10 @@ pub fn vote_resp_msg_type(t: MessageType) -> MessageType {
 
 // Calculate the quorum of a Raft cluster with the specified total nodes.
 pub fn quorum(total: usize) -> usize {
-    total / 2 + 1
+    match total {
```
BusyJay: I think the original implement is almost OK. For 0, it should panic instead.

lucab: Indeed you are right, only the 0-case was off. Perhaps I should just close this PR and wait to use [std::num::NonZeroUsize](https://doc.rust-lang.org/1.26.0/std/num/struct.NonZeroUsize.html) once it reaches stable (in 1.28.0)?

BusyJay: Interesting. `NonZeroUsize` sounds like a good idea to me.

lucab: Ack. I'm closing this, I'll keep a note to revisit after 1.28.0 release.

PR #74, opened 2018-06-08
downgrade protobuf to 1.x

BusyJay:
protobuf 2.0 changes the way it handles unknown enum. This can cause compatibility issues and there is no clean and efficient way to get around it yet. So let's stick to version 1.x until it's fixed.


Cargo.toml
```diff
@@ -13,7 +13,7 @@ categories = ["algorithms", "database-implementations"]
 
 [dependencies]
 log = "0.4.1"
-protobuf = "2.0"
+protobuf = "1.7"
```
hicqu: `eraftpb.rs` generated by `rust-protobuf:1.7` still has the problem. `read_proto3_enum_with_unknown_fields_into` is still called in `eraftpb.rs`. So maybe we can generate the file by `cargo install protobuf --vers=1.4.1 --force`.

hicqu:
rest LGTM.

BusyJay:
Found a solution to resolve this without downgrade the dependency, sent a new pr as #75.

PR #75, opened 2018-06-08
Use the old way to read enum

BusyJay:
We will add compatibility check once there is a clean and efficient way.


overvenus:
Could you add a test, if it is not urgent?
Do we need to update the kvproto with the same way?

BusyJay:
> Could you add a test, if it is not urgent?

The CI pass means that the generated code is correct already.

> Do we need to update the kvproto with the same way?

Yes.

Issue #70, opened 2018-05-31
Tested examples in src/lib.rs Rustdocs


Hoverbear:
https://github.com/pingcap/raft-rs/pull/69 introduces examples for usage to src/lib.rs. However most of them  are not actually compiled or tested.

We would like them to remove the `ignore` from the code fences so the examples are tested.

Please use [hiding](https://doc.rust-lang.org/1.24.1/rustdoc/documentation-tests.html#hiding-portions-of-the-example) in the examples to keep the redundant setup code from getting in the way.

@hoverbear is very happy to help mentor this issue. Just ping. :)

KeeganMyers:
@Hoverbear this will be my first PR in the Rust ecosystem, but I would be willing to give this a shot over the coming weekend 02/16-02/17.

Hoverbear:
@KeeganMyers Oh, amazing! What an honor. 🙇‍♀️ 

I'll make sure to keep an eye on this issue over the weekend. Let me know how I can help. Progress one or more examples would be a great PR. :)

The API can be a bit confusing, feel encouraged to ask questions.

KeeganMyers:
Sorry for the delay I have added hidden boilerplate to make all the examples pass. I did note that in the integration tests there are several errors:

test integration_cases::test_raft_paper::test_leader_election_in_one_round_rpc ... ok
ERROR 2019-02-21T23:05:57Z: raft::raft:  invalid format of MsgReadIndexResp from 0, entries count: 0

test integration_cases::test_raft_snap::test_snapshot_succeed ... ok
ERROR 2019-02-21T23:05:57Z: raft::raft: The node 1 already exists in the voters set.
ERROR 2019-02-21T23:05:57Z: raft::raft: The node 1 already exists in the voters set.

test integration_cases::test_raft::test_restore_learner_promotion ... ok
ERROR 2019-02-21T23:05:57Z: raft::raft:  can't become learner when restores snapshot [index: 11, term: 11]

test integration_cases::test_raft::test_add_node_check_quorum ... ok
ERROR 2019-02-21T23:05:57Z: raft::raft: The node 1 already exists in the voters set.

Should I explore and resolve these or are they outside of the scope of this issue?

Hoverbear:
@KeeganMyers Those are expected errors where we are testing to make sure Raft properly reports the error. It's ok. :) If you see the `cargo test` output report at the end there was a failure, then the tests failed, otherwise you don't need to worry.

PR #80, opened 2018-06-18
Use consistent naming throughout example code

djc:


djc:
Looking at it a bit more, I'm not really sure having the code from the example in the documentation root is worth it as it is right now: it is at the same time pretty elaborate but also leaves out some essential details that you need to understand to be able to grok the code. Also, given that it's ignored by rustc, it will likely grow more outdated and/or inconsistent over time. Given all this, it might be better to redirect to the `single_mem_node` example for actual code.

Hoverbear:
@djc Hi! Yes you may notice #70 has this same concern.

PR #93, opened 2018-07-12
Resolves issue #88: Insert functions don't panic now

### LINKED ISSUE #88: Insert Learner/Voter should not panic.
Currently these two functions have no return, yet they panic on an otherwise reasonable mistake which could be handled by application code.

https://github.com/pingcap/raft-rs/blob/1e0741a9aea3b242d33dadc46245ad611e7e7aa4/src/progress.rs#L104-L120

This issue would be solved with them returning a `Result<()>` with an error condition stating the voter/learner is already present, or some other `#[must_use]` value so that the user knows they need to handle this return value.

MusiKid:
These functions return now () or a custom Error enum.

Closes #88.


src/raft.rs
```diff
@@ -411,12 +411,16 @@ impl<T: Storage> Raft<T> {
         };
         for p in peers {
             let pr = new_progress(1, r.max_inflight);
-            r.mut_prs().insert_voter(*p, pr);
+            if let Err(e) = r.mut_prs().insert_voter(*p, pr) {
+                println!("{}", e);
```
siddontang: this error is serious and is not expected, I prefer panic direct or use fatal!

Hoverbear:
Thank you for tackling this! :bowing_woman:

MusiKid:
You're welcome !

src/raft.rs
MusiKid: Thanks for the review !
So it should panic in the raft mod ?

tests/cases/test_raft.rs
```diff
@@ -209,12 +209,16 @@ impl Interface {
                         is_learner: true,
                         ..Default::default()
                     };
-                    self.mut_prs().insert_learner(*id, progress);
+                    if let Err(e) = self.mut_prs().insert_learner(*id, progress) {
+                        println!("{}", e);
```
Hoverbear: You can make it `panic` here as well since it is tests.

MusiKid: ok

src/raft.rs
```diff
@@ -439,12 +439,16 @@ impl<T: Storage> Raft<T> {
         };
         for p in peers {
             let pr = new_progress(1, r.max_inflight);
-            r.mut_prs().insert_voter(*p, pr);
+            if let Err(e) = r.mut_prs().insert_voter(*p, pr) {
+                panic!("{}", e);
```
siddontang: em, I can't see the difference between panic inside the `insert_voter` and here, they are all in the Raft module and outside user can't know it.

Hoverbear: Later we can go and make `new()` return a `Result` when it fails instead of having a panic. It will be a follow up ticket.

src/progress.rs
```diff
@@ -119,27 +120,29 @@ impl ProgressSet {
     /// # Panics
```
Hoverbear: Please remove this comment now. :)

src/progress.rs
```diff
@@ -119,27 +120,29 @@ impl ProgressSet {
...
         if self.voters.insert(id, pr).is_some() {
-            panic!("insert voter {} twice", id);
+            return Err(AlreadyInVoters);
         }
+        Ok(())
     }
 
     /// Adds a learner to the cluster
     ///
     /// # Panics
```
Hoverbear: Please remove this comment now. :)

src/raft.rs
siddontang: @Hoverbear 

This function is not only used in the new function, other places also use it in depth, if we want to return error, we need to change a lot for many functions.

Hoverbear:
Closing in favor of #102

MusiKid:
Argh sorry I forgot about this issue !

Hoverbear:
No worries! Thanks for your work anyways. Really appreciate it!

PR #101, opened 2018-08-08
Joint Consensus

Hoverbear:
This PR enables Joint Consensus as described by the Raft paper (https://raft.github.io/raft.pdf).

As of this commit it is possible to undergo arbitrary peer membership changes in a safe way.

Notably, this gives us a resilient "Replace Node" functionality, which is able to progress in the situation of a loss of both the old (removed) and new (added) peers go down. This is not possible with our previous one-at-a-time strategy.

Unfortunately, this feature is fairly large in scope. Thankfully it's mostly testing code!

There is some new API surface, notably the `Raft::begin_membership_change` function.

There is also some moved API surface. The `RaftLog::applied_to` function has been moved to `Raft::commit_apply`. The old function, while it still works, carries a deprection warning. In the future we should make it a function available to only `Raft::applied_to`.

**Several APIs have changed** to return errors instead of panic or otherwise ignore non-critical errors. In most cases you can just do `old_call().ok()` to have the same behavior.

Then, in order to facilitate the needs in testing, some additional API was added to the `Network` harness, which now can dispatch messages and not automatically send responses, allowing time to inspect the state of peers.

Finally, many tests were added to ensure Joint Consensus can work.


Hoverbear:
(Whoops, wrong button)

src/progress.rs
```diff
@@ -48,9 +54,81 @@ impl Default for ProgressState {
...
+            learners: HashSet::with_capacity_and_hasher(learners, FxBuildHasher::default()),
+        }
+    }
+
+    // Validates that the configuration not problematic.
+    //
+    // Namely:
+    // * There can be no overlap of voters and learners.
+    // * There must be at least one voter.
+    pub fn valid(&self) -> Result<(), Error> {
```
hicqu: How about `validate`? It's a verb.

src/progress.rs
```diff
@@ -141,99 +252,283 @@ impl ProgressSet {
...
+        debug!("Inserting voter with id {}.", id);
+
+        if self.learner_ids().contains(&id) {
+            Err(Error::Exists(id, "learners"))?;
+        } else if self.voter_ids().contains(&id) {
+            Err(Error::Exists(id, "voters"))?;
         }
-        pr.is_learner = false;
         self.configuration.voters.insert(id);
+        self.next_configuration
```
hicqu: Can we call `insert_voter` if the progress currently in transision?

src/progress.rs
```diff
@@ -141,99 +252,283 @@ impl ProgressSet {
...
             self.progress.len()
         );
     }
+
+    /// Returns the minimal committed index for the cluster.
+    ///
+    /// Eg. If the matched indexes are [2,2,2,4,5], it will return 2.
+    pub fn minimum_committed_index(&self) -> u64 {
+        let mut matched = self
+            .voters()
```
hicqu: Is it correct when `progress.next_configuration` is not none? Suppose configuration contains voters `(1, 2, 3)`, and next_configuration contains voters `(2, 3, 4, 5)`, `voters` will be `(1, 2, 3, 4, 5)`, so the quorum is 3. However maybe we need acks from `(1, 2)` and `(3, 4, 5)` to commit a proposal.

src/progress.rs
Hoverbear: The user can `insert_voter` whenever they want, so it's never safe. They *should* only call it when they find a `EntryConfChange` `Entry`.

Right now the protection for doing so is here:

https://github.com/pingcap/raft-rs/blob/52856d1cdc3d65ec33539a7c9f9cffdd34529e1a/src/raft.rs#L1395-L1401

I am not satisfied with this though, and I definitely support this change.

src/progress.rs
Hoverbear: I think `config.valid()?;` reads like a question, a bit nicer than `config.validate()?`, what do you think?

src/progress.rs
Hoverbear: In this case the  quorum is 2 from the old, and 3 from the next, not 3. So we so acks from `(2,3,4)` is enough to commit, since both quorums are met.

I updated the code a bit (and I think fixed this that you were pointing it out).

src/progress.rs
hicqu: Got it! :)

Hoverbear:
@hicqu I've done some updating to make things cleaner, more understandable, and more consistent. In addition I fixed some bugs. PTAL. :)

src/progress.rs
```diff
@@ -275,61 +450,167 @@ impl ProgressSet {
...
+    ///
+    /// Valid transitions:
+    /// * Non-existing -> Learner
+    /// * Non-existing -> Voter
+    /// * Learner -> Voter
+    /// * Learner -> Non-existing
+    /// * Voter -> Non-existing
+    ///
+    /// Errors:
+    /// * Voter -> Learner
```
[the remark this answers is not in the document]
Hoverbear: There is nothing that dictates that we cannot do this, however this is not defined as part of the algorithm. (In fact, the learner process is intended to have automated promotion in the paper's description. Unfortunately we need a manual learner process due to how we are currently using learners.)

So this simply creates more surface to prove stability. While I'm not against supporting this feature, I'd prefer to leave this for future work, as I'd like to differentiate between learners (which are intended to become voters) and learners (which are intended to just be replica followers) in the future, which would definitely enable this demotion feature.

src/lib.rs
```diff
@@ -275,6 +364,8 @@ extern crate quick_error;
 #[cfg(test)]
 extern crate env_logger;
 extern crate rand;
+#[macro_use]
+extern crate getset;
```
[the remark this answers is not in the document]
Hoverbear: I'd like to use this library for most getters and setters (we have several it can replace.) So we could maybe do it now, but I'd remove it later and always use it.

tests/test_util/mod.rs
```diff
@@ -321,6 +335,19 @@ impl Network {
...
     }
 
+    /// Dispatches the given messages to the appropriate peers.
+    ///
+    /// Unlike `send` this does not gather and send any responses. It also does not ignore errors.
+    pub fn dispatch(&mut self, messages: impl IntoIterator<Item = Message>) -> Result<()> {
+        for message in self.filter(messages) {
+            let to = message.get_to();
+            let peer = self.peers.get_mut(&to).unwrap();
+            let result = peer.step(message);
```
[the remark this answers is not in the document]
Hoverbear: Oh, I meant to. :see_no_evil:

src/lib.rs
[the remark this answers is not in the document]
overvenus: Oh, I thought it was a test crate.

src/progress.rs
```diff
@@ -88,70 +181,99 @@ pub struct ProgressSet {
...
+            next_configuration: Option::default(),
         }
     }
 
     /// Returns the status of voters.
+    ///
+    /// **Note:** Do not use this for majority/quorum calculation. The Raft node may be
+    /// transitioning to a new configuration and have two qourums. Use `has_quorum` instead.
     #[inline]
     pub fn voters(&self) -> impl Iterator<Item = (&u64, &Progress)> {
```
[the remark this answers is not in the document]
overvenus: What about define a struct that impls `Iterator`? It is verbose though. Anyway, it is out of scope of this PR.

```rust
struct Voters<'p, 's> {
    progress: slice::Iter<'p, Progress>,
    set: &'s HashSet<u64>,
}

impl<'p, 's> Iterator for Voters<'p, 's> {
    type Item = &'p Progress;
    fn next(&mut self) -> Option<Self::Item> {
        loop {
            let pr = self.progress.next()?;
            if self.set.contains(&pr.id) {
                return Some(pr);
            }
        }
    }
}
```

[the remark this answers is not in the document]
Hoverbear: Ok, resolving this. :)

src/progress.rs
```diff
@@ -54,12 +55,98 @@ impl Default for ProgressState {
...
+        learners: impl IntoIterator<Item = u64>,
+    ) -> Self {
+        Self {
+            voters: voters.into_iter().collect(),
+            learners: learners.into_iter().collect(),
+        }
+    }
+}
+
+impl<I> From<(I, I)> for Configuration
```
hicqu: Maybe `(Iter1, Iter2)` is better? We don't need they are same type.

src/progress.rs
hicqu: Do we need to restrict the method to `pub(crate) insert_voter`? However, personally I still think just panic here if `next_configuration` is not none is the most simple solution. :)

src/progress.rs
```diff
@@ -259,13 +417,30 @@ impl ProgressSet {
...
+        let mut mci = matched[matched.len() / 2];
+
+        if let Some(next) = &self.next_configuration {
+            let mut matched = next
+                .voters()
+                .iter()
+                .map(|id| {
+                    let peer = &self.progress[id];
+                    peer.matched
+                }).collect::<Vec<_>>();
```
hicqu: Seems you collect a vector again. how about re-use `matched`?

src/progress.rs
```diff
@@ -275,61 +450,167 @@ impl ProgressSet {
...
+        let next = next.into();
+        next.valid()?;
+        // Demotion check.
+        if let Some(&demoted) = self
+            .configuration
+            .voters
+            .intersection(&next.learners)
+            .next()
+        {
+            Err(Error::Exists(demoted, "learners"))?;
```
hicqu: No return? Amazing it can work.

src/raft.rs
```diff
@@ -572,6 +627,54 @@ impl<T: Storage> Raft<T> {
...
+    ///
+    /// # Hooks
+    ///
+    /// * Post: Checks to see if it's time to finalize a Joint Consensus state.
+    pub fn commit_apply(&mut self, applied: u64) {
+        #[allow(deprecated)]
+        self.raft_log.applied_to(applied);
+
+        // Check to see if we need to finalize a Joint Consensus state now.
+        if let Some(index) = self.began_conf_change_at {
```
hicqu: What if the peer get down just after `Configuraion(old,new)` is just committed? Seems `began_conf_change_at` is not persisted with `Configuration(old,new)`.

src/progress.rs
Hoverbear: Thanks, I was lazy it seems. :(

src/progress.rs
Hoverbear: The `?` makes it return. Maybe I can make it more clear.

src/raft.rs
Hoverbear: It's a great catch. You're right. If the Raft node goes down right now it will forget when it `began_conf_change_at`. We'll need to persist it.

src/raft.rs
```diff
@@ -1039,7 +1152,123 @@ impl<T: Storage> Raft<T> {
...
+    ///
+    /// We apply the change when a node *applies* the entry, not when the entry is received.
+    ///
+    /// # Contracts
+    ///
+    /// * The `entry: Entry` is of type `EntryType::ConfChange` and the `data` field is a
+    /// `ConfChange` serialized by `protobuf::Message::write_to_bytes`.
+    /// * The `ConfChange` must be of type `BeginConfChange` and contain a `configuration` value.
+    #[inline(always)]
+    pub fn begin_membership_change(&mut self, entry: &Entry) -> Result<()> {
```
hicqu: Personally I prefer `ConfChange` instead of `Entry`. One reason is about style: `add_learner` and `and_node` didn't use `Entry`. And the another reason is if the method is called out of this crate, the `Entry` has been already parsed generally.

Hoverbear: Part of the reason for this is so we can do debug assertions to make sure the log index (found in the entry) actually contains the given ConfChange.

src/raw_node.rs
```diff
@@ -342,6 +350,13 @@ impl<T: Storage> RawNode<T> {
             ConfChangeType::AddNode => self.raft.add_node(nid),
             ConfChangeType::AddLearnerNode => self.raft.add_learner(nid),
             ConfChangeType::RemoveNode => self.raft.remove_node(nid),
+            ConfChangeType::BeginConfChange => self
+                .raft
+                .propose_membership_change(cc.get_configuration().clone())
```
hicqu: Is it not `begin_membership_change`?

src/raft.rs
```diff
@@ -1937,6 +2237,15 @@ impl<T: Storage> Raft<T> {
         self.raft_log.committed = hs.get_commit();
         self.term = hs.get_term();
         self.vote = hs.get_vote();
+        // HACK: Our current Protobuf uses `0` as `None`.
+        // Good thing we know that index 0 of the raft log is always not important in this context.
+        let began_conf_change_at = hs.get_began_conf_change_at();
+        if began_conf_change_at != 0 {
+            let entry = &self.get_store().entries(began_conf_change_at, began_conf_change_at + 1, NO_LIMIT)
```
hicqu: What if the raft log is compacted?
How about remove `began_conf_change_at` from `HardState`, but add `nodes_next_configuration` and `learners_next_configuration` in `ConfState`? So we can construct a ConfChange entry from it.

hicqu: There is an another problem about `began_conf_change_at`: in the current implement snapshots won't touch this field. For snapshot machanism, you can check `Raft::restore_raft` in `raft.rs`.

hicqu:
Nice test cases. Rest LGTM.

src/raft.rs
Hoverbear: Oh, a good point. I'll investigate this.

Hoverbear:
@hicqu Can you think of any other cases which we should definitely include?

src/raw_node.rs
Hoverbear: Yes. 🤦‍♀️

src/raft.rs
Hoverbear: This should be resolved, a really great catch. Added tests to cover it, too.

src/progress.rs
Hoverbear: @hicqu How about we make `insert_voter` and such return a Result instead?

Hoverbear: This would be a quite large change, actually. :( We have a lot of places where we swallow errors.

src/progress.rs
```diff
@@ -54,12 +55,99 @@ impl Default for ProgressState {
...
+    }
+
+    fn has_quorum(&self, potential_quorum: &FxHashSet<u64>) -> bool {
+        self.voters.intersection(potential_quorum).count() >= majority(self.voters.len())
+    }
+
+    /// Returns whether or not the given `id` is a member of this configuration.
+    // Allowed to maintain API consistency. See `HashSet::contains(&u64)`.
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub fn contains(&self, id: &u64) -> bool {
```
hicqu: How about pass `u64`?

proto/eraftpb.proto
```diff
@@ -91,11 +94,20 @@ enum ConfChangeType {
...
 }
 
 message ConfChange {
     uint64 id = 1;
     ConfChangeType change_type = 2;
+    // Used in `AddNode`, `RemoveNode`, and `AddLearnerNode`.
     uint64 node_id = 3;
     bytes context = 4;
+    // Used in `BeginMembershipChange` and `FinalizeMembershipChange`.
+    ConfState configuration = 5;
```
hicqu: Seems in `FinalizeMembershipChange` it must be blank?

hicqu:
Hi @Hoverbear I think the PR is too large. Could we seperate it to some little PRs? If you agree it, maybe I can make some help if you don't mind.

benches/suites/progress_set.rs
```diff
@@ -146,14 +147,32 @@ pub fn bench_progress_set_iter(c: &mut Criterion) {
...
-pub fn bench_progress_set_nodes(c: &mut Criterion) {
+pub fn bench_progress_set_voters(c: &mut Criterion) {
     let bench = |voters, learners| {
         move |b: &mut Bencher| {
             let set = quick_progress_set(voters, learners);
             b.iter(|| {
                 let set = set.clone();
-                let agg = set.iter().all(|_| true);
-                agg
+                set.voters().for_each(|_| {});
```
brson: This line looks like the kind of code that could be optimized away completely. Are you confident it's not?

benches/suites/progress_set.rs
```diff
@@ -146,14 +147,32 @@ pub fn bench_progress_set_iter(c: &mut Criterion) {
...
+    });
+}
+
+pub fn bench_progress_set_learners(c: &mut Criterion) {
+    let bench = |voters, learners| {
+        move |b: &mut Bencher| {
+            let set = quick_progress_set(voters, learners);
+            b.iter(|| {
+                let set = set.clone();
+                set.voters().for_each(|_| {});
```
brson: Ditto.

src/lib.rs
```diff
@@ -258,6 +284,74 @@ The `Ready` state contains quite a bit of information, and you need to check and
...
+* Add peer (learner or voter) *n* to the group.
+* Remove peer *n* from the group.
+* Remove a leader (unmanaged, via stepdown)
+* Promote a learner to a voter.
+
+It (currently) does not:
+
+* Allow control of the replacement leader during a stepdown.
+* Optionally roll back a change during a peer group pause where the new peer group configuration
+fails.
```
brson: Does this mean it leaves the cluster in a failure state?

proto/eraftpb.proto
```diff
@@ -80,6 +81,8 @@ message HardState {
     uint64 term = 1;
     uint64 vote = 2;
     uint64 commit = 3;
+    // If the peer goes down we need to be aware of a partially completed membership change.
```
hicqu: I prefer put it into `ConfState`.

proto/eraftpb.proto
```diff
@@ -80,6 +81,8 @@ message HardState {
...
+    ConfChange pending_membership_change = 4;
 }
 
 message ConfState {
```
hicqu: How about 
```protobuf
struct ConfState {
    repeated uint64 nodes = 1;
    repeated uint64 learners = 2;
    repeated uint64 conf_new_nodes = 3;
    repeated uint64 conf_new_learners = 4;
    int membership_change_start_at = 5;
}
```
And we don't need to store `ConfChange`. We can construct it from fields in `ConfState`.

proto/eraftpb.proto
hicqu: If you agree my comment about `ConfState`. Here I think we should add a new struct, for example:
```protobuf
struct MembershipChange {
    repeated uint64 nodes;
    repeated uint64 learner_nodes;
}
```

src/progress.rs
brson: I'm also curious. I would guess it's by-ref because it makes the implementation cleaner, but it's pretty unusual to take a concrete integer by reference, perhaps places some extra mental burden on the caller.

Hoverbear: It was mostly because `contains()` on a hashset takes a `&u64`, so I was trying to maintain that. I agree though, let's change it.

src/lib.rs
Hoverbear: No, the cluster would just be paused until it regains a majority.

src/raft.rs
```diff
@@ -573,6 +642,45 @@ impl<T: Storage> Raft<T> {
...
+            .pending_membership_change
+            .as_ref()
+            .map(|v| Some(v.get_start_index()))
+            .unwrap_or(None);
+
+        if let Some(index) = start_index {
+            // Invariant: We know that if we have commited past some index, we can also commit that index.
+            if applied >= index && self.state == StateRole::Leader {
+                // We must replicate the commit entry.
+                self.append_finalize_conf_change_entry();
```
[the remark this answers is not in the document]
Hoverbear: This is documented clearly in the `src/lib.rs`:

```
> **Note:** In order to maintain resiliency guarantees (progress while a majority of both peer sets is
active), it is very important to wait until the entire peer group has exited the transition phase
before taking old, removed peers offline.
```

Even if we chose to have the client call `propose_finalize_membership_change` they could not turn off those nodes until the entire peer group has **applied** the index. So in order to feel confident they can turn off a node, they must wait until the majority of nodes have applied this change.

proto/eraftpb.proto
Hoverbear: 😆 This is exactly what `ConfState` was originally.

src/progress.rs
```diff
@@ -167,60 +290,109 @@ impl ProgressSet {
...
     }
 
     /// Removes the peer from the set of voters or learners.
-    pub fn remove(&mut self, id: u64) -> Option<Progress> {
-        self.configuration.voters.remove(&id);
+    ///
+    /// # Errors
+    ///
+    /// * There is a pending membership change.
+    pub fn remove(&mut self, id: u64) -> Result<Option<Progress>> {
```
brson: Looks like this function should be named `remove_voter`. Compare with `insert_voter`, `insert/remove_learner`.

brson: Wait, this function does both learners and voters. Hm.

Hoverbear: This removes voters or learners, so we've need to provide `remove_learner` as well. :( So keeping it just `remove` is simpler.

brson: The docs should probably indicate this works for both voters and learners, since it's asymmetrical with the insert methods.

brson: Oh, wait again, the docs say that.

Hoverbear: Yup!

```
/// Removes the peer from the set of voters or learners.
```

src/progress.rs
```diff
@@ -167,60 +290,109 @@ impl ProgressSet {
...
         let removed = self.progress.remove(&id);
+
         self.assert_progress_and_configuration_consistent();
-        removed
+        Ok(removed)
     }
 
     /// Promote a learner to a peer.
-    pub fn promote_learner(&mut self, id: u64) -> Result<(), Error> {
+    pub fn promote_learner(&mut self, id: u64) -> Result<()> {
```
brson: Docs should probably contain errors cases like the others methods.

src/progress.rs
```diff
@@ -257,13 +430,28 @@ impl ProgressSet {
...
-        self.voters().for_each(|(_id, peer)| {
+        self.configuration.voters().iter().for_each(|id| {
+            let peer = &self.progress[id];
             matched.push(peer.matched);
         });
         // Reverse sort.
         matched.sort_by(|a, b| b.cmp(a));
-        // Smallest that the majority has commited.
-        matched[matched.len() / 2]
+        let mut mci = matched[matched.len() / 2];
```
brson: Why does the middle element give the 'max'?

proto/eraftpb.proto
Hoverbear: @hicqu and I spoke offline (In person, even!) about this and we think it might be more clear to think of Joint Consensus as moving **between** `ConfState`s, so instead we will try having a `pending_conf_state` stored alongside `conf_state` in the `RaftState` and the Snapshots.

src/progress.rs
```diff
@@ -273,61 +461,168 @@ impl ProgressSet {
...
+    #[inline]
+    pub fn is_in_membership_change(&self) -> bool {
+        self.next_configuration.is_some()
+    }
+
+    /// Enter a joint consensus state to transition to the specified configuration.
+    ///
+    /// The `next` provided should be derived from the `ConfChange` message. `progress` is used as
+    /// a basis for created peer `Progress` values. You are only expected to set `ins` from the
+    /// `raft.max_inflights` value.
```
brson: It's not clear to me what `ins` represents here. It's an an argument to the function.

src/progress.rs
```diff
@@ -273,61 +461,168 @@ impl ProgressSet {
...
+
+        // When a peer is first added/promoted, we should mark it as recently active.
+        // Otherwise, check_quorum may cause us to step down if it is invoked
+        // before the added peer has a chance to communicate with us.
+        progress.recent_active = true;
+        progress.paused = false;
+        for id in next.voters.iter().chain(&next.learners) {
+            self.progress.entry(*id).or_insert_with(|| progress.clone());
+        }
+        self.next_configuration = Some(next);
```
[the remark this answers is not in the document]
brson: At least in this file I don't see any code paths that deal with a failed membership change. Is that possible, or does the next configuration stay valid forever until both configurations reach quorum? Maybe overwriting a `Some` `next_configuration` just means we've given up on a previous attempt and started a new attempt?

It looks like there's some garbage collection in the progress set that needs to be done to clean up new nodes in the case of a failed membership change.

src/progress.rs
```diff
@@ -841,4 +1149,101 @@ mod test_progress_set {
...
+
+        let mut set = ProgressSet::default();
+        let default_progress = Progress::new(0, 10);
+
+        for starter in start_voters {
+            set.insert_voter(starter, default_progress.clone())?;
+        }
+        for starter in start_learners {
+            set.insert_learner(starter, default_progress.clone())?;
+        }
```
brson: Why is the initial `ProgressSet` created by calling `default` and then mutating it with accessors, when doing a membership change means passing in a complete `Configuration`? The latter seems more foolproof. The mutating accessors seem easy to misuse.

Thinking of creating the initial set itself as a membership change _sounds_ like a reasonable generalization.

src/progress.rs
```diff
@@ -841,4 +1149,101 @@ mod test_progress_set {
...
+    fn test_membership_change_configuration_empty_sets() {
+        assert!(check_membership_change_configuration((vec![], vec![]), (vec![], vec![])).is_err())
+    }
+
+    #[test]
+    fn test_membership_change_configuration_empty_voters() {
+        assert!(
+            check_membership_change_configuration((vec![1], vec![]), (vec![], vec![]),).is_err()
+        )
+    }
```
brson: The above two error cases are testing for errors in the starting configuration, but there are not similar tests for the ending configuration, and it looks to me the starting and ending cases have different code paths. If so, can you add more of these same tests but for the ending set?

src/progress.rs
```diff
@@ -841,4 +1149,101 @@ mod test_progress_set {
...
+        set.finalize_membership_change()?;
+        assert!(!set.is_in_membership_change());
+        assert_eq!(set.voter_ids(), end_voters, "End state voters inaccurate");
+        assert_eq!(
+            set.learner_ids(),
+            end_learners,
+            "End state learners inaccurate"
+        );
+        Ok(())
+    }
```
brson: Maybe it's impossible to happen in practice, but (similarly to a previous comment) it seems like the case where `begin_membership_change` is called twice before `finalize_membership_change` isn't represented in this file. If it's not possible, maybe an assertion in `begin_membership_change` is warrented; if it is possible, then tests.

brson: Other test ideas:

- finalize_membership_change called without a begin_membership_change
- finalize_membership_change called twice without intervening begin_membership_change

brson: Do `ProgressSet::voters_mut/learners_mut/iter_mut` need to be `pub`?  Seem like easy methods with which to break internal consistency. I don't see uses of voters_mut/learners_mut outside the mod, not sure about iter_mut.

With those mutable accessors, this module kinda wants a lot more tests of weird cases. Without those accessors there aren't as many weird cases.

Likewise with the `insert/remove` methods. If those can be called at any time, then it seems like there's a lot of mayhem one could cause, and more test cases to write. If those were eliminated and handled purely in the constructor or themselves via membership changes then there's fewer cases to think about.

Hoverbear:
We've discussed this in a meeting today and identified a possible issue.

Given the situation of going from peer group `ABC` to `DE` (or some other where the new set is disjoint):

During the intermediate phase between `Begin` and `Finalize` if the **entire** old (ABC) group disappears then the new group (DE) cannot elect a leader. 

This is problematic.

So, in order to avoid this, we will ban removing the leader from the group.

This helps us avoid the situation.

BusyJay:
Why not stick to the implementation described in the paper? So that we can avoid such problem and other potential problems completely.

src/progress.rs
```diff
@@ -167,60 +290,109 @@ impl ProgressSet {
...
+    pub fn insert_learner(&mut self, id: u64, pr: Progress) -> Result<()> {
+        debug!("Inserting learner with id {}.", id);
+
+        if self.learner_ids().contains(&id) {
+            return Err(Error::Exists(id, "learners"));
+        } else if self.voter_ids().contains(&id) {
             return Err(Error::Exists(id, "voters"));
+        } else if self.is_in_membership_change() {
+            return Err(Error::ViolatesContract(
+                "There is a pending membership change.".into(),
```
hicqu: No dots on errors please thanks

src/progress.rs
```diff
@@ -273,61 +461,168 @@ impl ProgressSet {
...
+    /// * Non-existing -> Voter
+    /// * Learner -> Voter
+    /// * Learner -> Non-existing
+    /// * Voter -> Non-existing
+    ///
+    /// Errors:
+    /// * Voter -> Learner
+    /// * Member as voter and learner.
+    /// * Empty voter set.
+    pub fn begin_membership_change(
```
hicqu:
```suggestion
    pub(crate) fn begin_membership_change(
```

src/raft.rs
```diff
@@ -692,6 +800,7 @@ impl<T: Storage> Raft<T> {
     ///
     /// Panics if a leader already exists.
     pub fn become_candidate(&mut self) {
+        trace!("ENTER become_candidate");
```
hicqu: Yank these

src/raft.rs
```diff
@@ -1652,6 +1898,7 @@ impl<T: Storage> Raft<T> {
     /// For a given message, append the entries to the log.
     pub fn handle_append_entries(&mut self, m: &Message) {
         if m.get_index() < self.raft_log.committed {
+            debug!("Got message with lower index than committed.");
```
hicqu: Tag here too

src/raft_log.rs
```diff
@@ -295,6 +297,7 @@ impl<T: Storage> RaftLog<T> {
 
     /// Appends a set of entries to the unstable list.
     pub fn append(&mut self, ents: &[Entry]) -> u64 {
+        trace!("Entries being appended to unstable list: {:?}", ents);
```
hicqu: Yank

BusyJay:
I don't think disable removing leader can solve the problem at all. Because the leader can be down at any time, in which case, all new peers may not be able to elect a leader successfully if they are all not part of the old configuration.

Hoverbear:
@busyjay I thought about this more last night too, and I agree it doesn't solve the problem.

src/progress.rs
```diff
@@ -257,13 +430,28 @@ impl ProgressSet {
     pub fn maximal_committed_index(&self) -> u64 {
```
hicqu: To make the API work for "membership change can work on received instead of applied", we need to add an argument `limit` (it's just Raft::pending_conf_index) to frobiden `BeginMembershipChange` to be committed by only old configuration.

hicqu: After think more about it, it's not necessary in fact, because when we fetch `Ready` from `RawNode` on leader, the `BeginMembershipChange` hasn't been broadcasted to other peers. So it will never be committed.
So please ignore this. :)

src/progress.rs
Hoverbear: Let's leave this for future work, it's not directly relevant to this PR and this PR has been held up long enough.

I agree though. :)

src/progress.rs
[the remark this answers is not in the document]
Hoverbear: This is guarded in `raft.rs`. While it's possible to misuse `progress.rs` code without going through `raft.rs` I think for now we have to accept that we can't change the whole world in one PR. 

At the moment we don't support a rollback, so if the first `begin_membership_change` call fails it's assumed that we never enter the state.

src/progress.rs
Hoverbear: I'm confused by this comment. :( Maybe you can clarify it with me.

src/progress.rs
Hoverbear: Oh, amazing. Great thoughts. You are so smart. :) I'm jealous.

tests/integration_cases/test_membership_changes.rs
```diff
@@ -0,0 +1,1689 @@
+// Copyright 2016 PingCAP, Inc.
```
nolouch: 2019?

tests/integration_cases/test_membership_changes.rs
```diff
@@ -0,0 +1,1689 @@
...
+            assert!(
+                peer.is_in_membership_change(),
+                "Peer {} should have been in a membership change.",
+                peer.id
+            );
+        }
+    }
+
+    /// Reads the pending entries to be applied to a raft peer, checks one is of the expected variant, and applies it. Then, it advances the node to that point in the configuration change.
+    fn expect_apply_membership_change_entry<'a>(
```
nolouch: So now we need to switch to on message?

tests/integration_cases/test_membership_changes.rs
```diff
@@ -0,0 +1,1689 @@
...
+    }
+
+    /// Send a message proposing a "one-by-one" style AddNode configuration.
+    /// If the peers are in the midst joint consensus style (Begin/FinalizeMembershipChange) change they should reject it.
+    fn propose_add_node_message(&mut self, id: u64) -> Result<()> {
+        info!("Proposing add_node message. Target: {:?}", id,);
+        let message = build_propose_add_node_message(
+            self.old_leader,
+            id,
+            self.peers[&1].raft_log.last_index() + 1,
```
nolouch: Why always 1?

src/progress.rs
Hoverbear: They don't need to be pub, but I think we should be cautious not to create too many API changes in this PR. It's more appropriate for a new PR.

Hoverbear:
PTAL again pre-merge. Please pay particular attention to documentation.

zhangjinpeng1987:
@BusyJay @hicqu PTAL again.

hicqu:
LGTM. PTAL @nolouch @Busyjay

PR #114, opened 2018-09-06
tests: use fail point

BusyJay:
Remove the hacky function pointer, and use failpoint to do the precise
control.


.travis.yml
```diff
@@ -25,3 +25,4 @@ matrix:
 
 script:
   - cargo test --all -- --nocapture
```
Hoverbear: Is there a reason not to have failpoints part of the normal `test`?

BusyJay: Because failpoints injest failure in code path, which will affect all concurrently running tests, Hence they need to be synchronized, which make tests slow.

Hoverbear: :+1: Great reasoning, can you please add that in a comment?

Hoverbear:
I pushed a `cargo fmt` run since it updated.

PR #118, opened 2018-09-07
Improve examples in documentation.

Hoverbear:
Adds another tested documentation example.

Part of #70


overvenus:
I'm curious about why CI reports ok while there are some panicking tests.

Hoverbear:
@overvenus Those are tests that `#[should_panic]`.

PR #119, opened 2018-09-07
Remove Progress.is_learner

Hoverbear:
Remove the `is_learner` value from `Progress`.

# Rationale

In the process of Joint Consensus (#101) we will have two different `Configuration` sets during some points in operation.

In #108 we made `Progress` a single set (instead of having separate sets for voters and learners). `Progress` still held an `is_learner` value, which meant we had two ways to determine if a node was a learner (via the `ProgressSet.has_learner()` and `Progress.is_learner`). This is not necessary.

We can use `ProgressSet.has_learner(id)` and `ProgressSet.has_voter(id)` exclusively.

# Benefits

Doing this helps minimize complexity, and reduces places we keep track of voter status.

# Future Work

In the future we may be in a joint consensus state, meaning that the `Progress.is_learner` value could be incorrect (If you promoted a learner, then the old configuration still has it has a non-voter).

This internalizes the concept of the `is_learner()` so we can minimize possible bug surface.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -153,7 +153,7 @@ impl ProgressSet {
     }
 
     /// Adds a voter node
-    pub fn insert_voter(&mut self, id: u64, mut pr: Progress) -> Result<(), Error> {
+    pub fn insert_voter(&mut self, id: u64, pr: Progress) -> Result<(), Error> {
         // If the progress exists already this is in error.
         if self.progress.contains_key(&id) {
             // Determine the correct error to return.
@@ -162,7 +162,6 @@ impl ProgressSet {
             }
             return Err(Error::Exists(id, "voters"));
         }
-        pr.is_learner = false;
         self.configuration.voters.insert(id);
         self.progress.insert(id, pr);
         self.assert_progress_and_configuration_consistent();
@@ -170,7 +169,7 @@ impl ProgressSet {
     }
 
     /// Adds a learner to the cluster
-    pub fn insert_learner(&mut self, id: u64, mut pr: Progress) -> Result<(), Error> {
+    pub fn insert_learner(&mut self, id: u64, pr: Progress) -> Result<(), Error> {
         // If the progress exists already this is in error.
         if self.progress.contains_key(&id) {
             // Determine the correct error to return.
@@ -179,7 +178,6 @@ impl ProgressSet {
             }
             return Err(Error::Exists(id, "voters"));
         }
-        pr.is_learner = true;
         self.configuration.learners.insert(id);
         self.progress.insert(id, pr);
         self.assert_progress_and_configuration_consistent();
@@ -197,15 +195,16 @@ impl ProgressSet {
 
     /// Promote a learner to a peer.
     pub fn promote_learner(&mut self, id: u64) -> Result<(), Error> {
-        match self.progress.get_mut(&id) {
-            Some(progress) => if !progress.is_learner {
-                Err(Error::Exists(id, "voters"))?;
-            } else {
-                progress.is_learner = false;
-                self.configuration.voters.insert(id);
-                self.configuration.learners.remove(&id);
-            },
-            None => Err(Error::NotExists(id, "learners"))?,
+        let is_learner = self.learner_ids().contains(&id);
+        let is_voter = self.voter_ids().contains(&id);
+
+        if is_voter {
+            return Err(Error::Exists(id, "voters"));
+        } else if is_learner {
+            self.configuration.voters.insert(id);
+            self.configuration.learners.remove(&id);
+        } else {
+            return Err(Error::NotExists(id, "learners"));
         }
         self.assert_progress_and_configuration_consistent();
         Ok(())
@@ -272,18 +271,14 @@ pub struct Progress {
     /// When a leader receives a reply, the previous inflights should
     /// be freed by calling inflights.freeTo.
     pub ins: Inflights,
-
-    /// Indicates the Progress is a learner or not.
-    pub is_learner: bool,
 }
 
 impl Progress {
     /// Creates a new progress with the given settings.
-    pub fn new(next_idx: u64, ins_size: usize, is_learner: bool) -> Self {
+    pub fn new(next_idx: u64, ins_size: usize) -> Self {
         Progress {
             next_idx,
             ins: Inflights::new(ins_size),
-            is_learner,
             ..Default::default()
         }
     }
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -717,20 +717,6 @@ mod test_progress_set {
     }
 
     #[test]
-    fn test_promote_learner_does_not_exist() -> Result<()> {
-        let mut set = ProgressSet::default();
-        assert!(
-            set.promote_learner(1).is_err(),
-            "Should return an error on invalid voter insert."
-        );
-        assert!(
-            set.get(1).is_none(),
-            "Should not have inserted a node on invalid promote_learner."
-        );
-        Ok(())
-    }
-
-    #[test]
     fn test_promote_learner_already_voter() -> Result<()> {
         let mut set = ProgressSet::default();
         let default_progress = Progress::default();
@@ -738,6 +724,10 @@ mod test_progress_set {
         let pre = set.get(1).expect("Should have been inserted").clone();
         assert!(
             set.promote_learner(1).is_err(),
+            "Should return an error on invalid promote_learner."
+        );
+        assert!(
+            set.promote_learner(2).is_err(),
             "Should return an error on invalid promote_learner."
         );
         assert_eq!(pre, *set.get(1).expect("Peer should not have been deleted"));
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -229,7 +229,7 @@ impl ProgressSet {
 }
 
 /// The progress of catching up from a restart.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default, Clone, PartialEq)]
 pub struct Progress {
     /// How much state is matched.
     pub matched: u64,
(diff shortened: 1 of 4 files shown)


src/raft.rs
```diff
@@ -1993,12 +1993,13 @@ impl<T: Storage> Raft<T> {
     fn check_quorum_active(&mut self) -> bool {
         let self_id = self.id;
         let mut act = 0;
+        let learners = self.prs().learner_ids().clone();
```
Hoverbear: The clone here seems the best way, as much as I hate it. Since we use `.mut_prs()` and iterate we have a mutable borrow out for `self`. 

We copy this `HashSet<u64>` which should only have learner _ids() in it, which should be not unreasonably large. An alternative is we can clone `voter_ids()` which should always be reasonable even if there are many learners.

src/progress.rs
```diff
@@ -197,15 +195,13 @@ impl ProgressSet {
...
-        match self.progress.get_mut(&id) {
-            Some(progress) => if !progress.is_learner {
-                Err(Error::Exists(id, "voters"))?;
-            } else {
-                progress.is_learner = false;
-                self.configuration.voters.insert(id);
-                self.configuration.learners.remove(&id);
-            },
-            None => Err(Error::NotExists(id, "learners"))?,
+        if self.learner_ids().contains(&id) {
```
BusyJay: Why not just `remove`?

src/raft.rs
BusyJay: You can use `take_prs` to get around the lifetime check just as other places already did.

src/raft.rs
```diff
@@ -1993,12 +1993,13 @@ impl<T: Storage> Raft<T> {
...
         for (&id, pr) in self.mut_prs().iter_mut() {
             if id == self_id {
                 act += 1;
                 continue;
             }
-            if !pr.is_learner && pr.recent_active {
+            if !learners.contains(&id) && pr.recent_active {
```
BusyJay: Seems we can split it into two loop for `voters` and `learners` to avoid the query.

src/raft.rs
```diff
@@ -1851,20 +1851,20 @@ impl<T: Storage> Raft<T> {
 
         // Ignore redundant inserts.
         // TODO: Remove these and have this function and related functions return errors.
-        if let Some(progress) = self.prs().get(id) {
-            // If progress.is_learner == learner, then it's already inserted as what it should be, return early to avoid error.
-            if progress.is_learner == learner {
+        if self.prs().get(id).is_some() {
```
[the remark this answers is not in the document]
BusyJay: Agree. `insert_learner` and similar functions do the check already. We can check the result to reflect errors.

src/raft.rs
```diff
@@ -1851,20 +1851,20 @@ impl<T: Storage> Raft<T> {
...
-            if !progress.is_learner && learner {
+            // If progress is a voter, and learner == true, then it's a demotion, return early to avoid an error.
+            if self.prs().voter_ids().contains(&id) && learner {
                 info!("{} Ignoring voter demotion of ID {}.", self.tag, id);
                 return;
             }
         };
 
-        let progress = Progress::new(self.raft_log.last_index() + 1, self.max_inflight, learner);
+        let progress = Progress::new(self.raft_log.last_index() + 1, self.max_inflight);
```
BusyJay: Better move it inside condition body. There will be an allocation when creating a progress.

src/progress.rs
Hoverbear: Yes, I could check if it is `Some(v)` and put it back, returning an error. Good idea. :)

src/raft.rs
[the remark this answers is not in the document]
Hoverbear: Ok. I will make this function (which is private) return a `Result<()>`. I will then make the callers (which are pub) panic on error in them instead to maintain behavior. I would like to make them return `Result<()>` in 0.5.0.

What do you think?

src/raft.rs
Hoverbear: I always feel so dirty doing this. :rofl: But yes you're right.

Hoverbear:
@BusyJay PTAL again when you have a chance. :)

src/progress.rs
```diff
@@ -195,13 +195,13 @@ impl ProgressSet {
...
     pub fn promote_learner(&mut self, id: u64) -> Result<(), Error> {
-        if self.learner_ids().contains(&id) {
-            self.configuration.voters.insert(id);
-            self.configuration.learners.remove(&id);
-        } else if self.voter_ids().contains(&id) {
+        if !self.configuration.learners.remove(&id) {
+            // Wasn't already a voter. We can't promote what doesn't exist.
+            return Err(Error::Exists(id, "learners"));
+        }
+        if !self.configuration.voters.insert(id) {
```
BusyJay: This should always be false, which is guaranteed by `assert_progress_and_configuration_consistent`.

src/raft.rs
```diff
@@ -1843,38 +1843,24 @@ impl<T: Storage> Raft<T> {
...
             self.mut_prs().promote_learner(id)
         } else {
+            let progress = Progress::new(self.raft_log.last_index() + 1, self.max_inflight);
             self.mut_prs().insert_voter(id, progress)
         };
 
-        if let Err(e) = result {
-            panic!("{}", e)
+        if let Err(ref e) = result {
+            error!("{}", e);
```
BusyJay: Should return here. And seems a test case is missing.

src/raft.rs
```diff
@@ -1993,17 +1982,20 @@ impl<T: Storage> Raft<T> {
...
-        let learners = self.prs().learner_ids().clone();
-        for (&id, pr) in self.mut_prs().iter_mut() {
-            if id == self_id {
-                act += 1;
-                continue;
+
+        self.prs = if let Some(mut prs) = self.prs.take() {
+            for (&id, pr) in prs.voters_mut() {
+                if id == self_id || pr.recent_active {
+                    act += 1;
```
BusyJay: Should continue when `id == self_id`.

src/progress.rs
Hoverbear: `assert_progress_and_configuration_consistent` validates conditions on the end of an operation.

This checks for a case such as:

```rust
prs.insert_voter(1); // Whoops! I meant learner.
prs.promote_learner(1); // -> Gives an Error::Exists(1, "voters")
```

src/raft.rs
Hoverbear: Ah, yes, you're very right.

src/progress.rs
BusyJay: The `assertion` appears everywhere, so the correctness should always be hold. The example you provide will return error at L200.

src/raft.rs
```diff
@@ -1993,16 +1983,23 @@ impl<T: Storage> Raft<T> {
...
         let self_id = self.id;
         let mut act = 0;
-        for (&id, pr) in self.mut_prs().iter_mut() {
-            if id == self_id {
-                act += 1;
-                continue;
+
+        self.prs = if let Some(mut prs) = self.prs.take() {
+            for (&id, pr) in prs.voters_mut() {
+                if id == self_id || pr.recent_active {
```
BusyJay: `recent_active` should be set to false before continue.

src/raft.rs
BusyJay: Any tests?

src/progress.rs
Hoverbear: The error at `L200` appears from:

```rust
// Do **not** add 1 as a learner.
prs.promote_learner(1); // Err(Error::Exists(id, "learners"))
```

The conditions of this function are:

* The `id` must currently be a learner. (Checked on L198),
* The `id` is not already a voter. (Checked here on L202),

src/raft.rs
Hoverbear: I guess you mean to test that the `ProgressSet` is not mutated if an `Error` is returned?

I think it is better to test the `prs.insert_voter()` and `prs.insert_learner()` functions if that's the case.

src/raft.rs
Hoverbear: Then we are back to the same code as before https://github.com/pingcap/raft-rs/pull/119/files/2fa4a6543e95d75ba929dc12bb3c8eaa274eec98..47e441f9f7c5db5e021fd13a831b99587c726b9b#r222180743

```rust
            for (&id, pr) in prs.voters_mut() {
                if id == self_id || pr.recent_active {
                    act += 1;
                    pr.recent_active = false;
                    continue;
                }
                pr.recent_active = false;
            }
```

Which is equivalent to:

```rust
            for (&id, pr) in prs.voters_mut() {
                if id == self_id || pr.recent_active {
                    act += 1;
                }
                pr.recent_active = false;
            }
```

src/raft.rs
BusyJay: It's not just `ProgressSet`. For example, L1867 mutate the internal state of raft state machine.

src/progress.rs
BusyJay: Can you write a test case that will make this function return `Err(Error::Exists(id, "voters"))`? From what I can see, this is impossible via the public APIs.

src/raft.rs
BusyJay: Apparently not. You can compare the code with original especially how `recent_active` is mutated when it's self_id or not.

src/progress.rs
Hoverbear: Using `Raft::add_node()` or `Raft::add_voter()` which call `Raft::add_voter_or_learner()` cannot result in this error, it's true.

Note that `raft.add_node(1);  raft.prs().promote_learner(1);` is completely possible for a user to do. So we must protect from it.

src/raft.rs
Hoverbear: Hm yes, I took a very close look and you're right. I have fixed this.

src/raft.rs
```diff
@@ -1993,16 +1983,24 @@ impl<T: Storage> Raft<T> {
...
+                if id == self_id {
+                    act += 1;
+                    continue;
+                }
+                if pr.recent_active { act += 1; }
+                pr.recent_active = false;
             }
-            if !pr.is_learner && pr.recent_active {
-                act += 1;
+            for (&_id, pr) in prs.learners_mut() {
```
hicqu: Is it not necessary? @BusyJay .

BusyJay: Not sure. But it's better keep the old behavior.

hicqu:
LGTM.

Hoverbear:
@BusyJay Can you take another look?

hicqu:
PTAL @BusyJay .

src/progress.rs
```diff
@@ -630,3 +622,115 @@ mod test {
...
+        assert_eq!(
+            *set.get(1).expect("Should be inserted."),
+            default_progress,
+            "The ProgressSet was mutated in a `insert_voter` that returned error."
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_insert_redundant_learner() -> Result<()> {
```
BusyJay: The test case doesn't match its name.

src/raft.rs
BusyJay: Where is the test case?

Hoverbear:
@BusyJay Please take a look again, the test name is fixed.

src/raft.rs
Hoverbear: I did not see this comment about L1867. Good catch. I'll add it.

tests/integration_cases/test_raft.rs
```diff
@@ -3755,6 +3755,22 @@ fn test_add_learner() {
     assert!(n1.prs().learner_ids().contains(&2));
 }
 
+// Ensure when add_voter is called on a peers own ID that it will be promoted.
+// When the action fails, ensure it doesn't mutate the raft state.
+#[test]
```
Hoverbear: Here is that mutation test, @BusyJay.

src/progress.rs
```diff
@@ -197,15 +195,13 @@ impl ProgressSet {
...
-                Err(Error::Exists(id, "voters"))?;
-            } else {
-                progress.is_learner = false;
-                self.configuration.voters.insert(id);
-                self.configuration.learners.remove(&id);
-            },
-            None => Err(Error::NotExists(id, "learners"))?,
+        if !self.configuration.learners.remove(&id) {
+            // Wasn't already a voter. We can't promote what doesn't exist.
+            return Err(Error::Exists(id, "learners"));
```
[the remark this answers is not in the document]
Hoverbear: Oh, yes, definitely. The comment is also wrong.

Hoverbear:
@siddontang Resolved, PTAL. I think it's ready for merge.

PR #122, opened 2018-09-12
Change return type of Raft::new() to Result

### LINKED ISSUE #94: Raft::new() returns a Result
Depends on:
* #88 

Currently calling `Raft::new()` will panic if it is provided an invalid configuration.

https://github.com/pingcap/raft-rs/blob/18f1cd8bb1a452dd9a2e3a8c0b507fc138c1862d/src/raft.rs#L390-L391

The consequence of this is that the user of the `Raft` is not provided an opportunity to handle certain types of error.

Fixing this issue would include changing `Raft::new()` to return a `Result<Raft<T>>` instead of a `Raft<T>`. In order to do this you may need to add new error variants:

https://github.com/pingcap/raft-rs/blob/18f1cd8bb1a452dd9a2e3a8c0b507fc138c1862d/src/errors.rs#L19-L22.

kleimkuhler:
This allows a `ConfigInvalid` error to propogate up and let user handle it.

Fixes #94

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -220,7 +220,7 @@ impl<T: Storage> Raft<T> {
     /// Creates a new raft for use on the node.
     pub fn new(c: &Config, store: T) -> Result<Raft<T>> {
         c.validate()?;
-        let rs = store.initial_state().expect("");
+        let rs = store.initial_state()?;
         let conf_state = &rs.conf_state;
         let raft_log = RaftLog::new(store, c.tag.clone());
         let mut peers: &[u64] = &c.peers;
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -218,8 +218,8 @@ pub fn quorum(total: usize) -> usize {
 
 impl<T: Storage> Raft<T> {
     /// Creates a new raft for use on the node.
-    pub fn new(c: &Config, store: T) -> Raft<T> {
-        c.validate().expect("configuration is invalid");
+    pub fn new(c: &Config, store: T) -> Result<Raft<T>> {
+        c.validate()?;
         let rs = store.initial_state().expect("");
         let conf_state = &rs.conf_state;
         let raft_log = RaftLog::new(store, c.tag.clone());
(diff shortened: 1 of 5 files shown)


DorianZheng:
@Hoverbear @BusyJay PTAL

src/raft.rs
```diff
@@ -223,8 +223,8 @@ pub fn quorum(total: usize) -> usize {
 
 impl<T: Storage> Raft<T> {
     /// Creates a new raft for use on the node.
-    pub fn new(c: &Config, store: T) -> Raft<T> {
-        c.validate().expect("configuration is invalid");
+    pub fn new(c: &Config, store: T) -> Result<Raft<T>> {
+        c.validate()?;
         let rs = store.initial_state().expect("");
```
Hoverbear: Maybe we can also push up this expect?

kleimkuhler: My intention here was to propagate the `Error:ConfigInvalid` returned by `validate()`. https://github.com/pingcap/raft-rs/blob/13ec5ac33a3800f89a6946cb8c54ceba4b84ab98/src/config.rs#L170

I can make the test case more explicit to show that? Changing my test case to the below would show that it propagates back up to the user.
```
fn test_new_raft_with_bad_config_errors() {
    let invalid_config = new_test_config(INVALID_ID, vec![1, 2], 1, 1);
    let raft = Raft::new(&invalid_config, new_storage());
    assert_eq!(
        raft.err(),
        Some(Error::ConfigInvalid("invalid node id".to_owned()))
    )
}
```

Hoverbear: Oh, yes, that's fine. I meant the `store.initial_state()` could probably use a `?` as well instead of an `expect()`.

kleimkuhler: Oh! Yes, sorry about that; I'll change that as well.

Hoverbear:
@kleimkuhler  Whoops! We both forgot the benchmarks! Could you please bump those too?

kleimkuhler:
@Hoverbear Totally missed the benchmarking; sorry about that! Everything should be good to go now.

hicqu:
LGTM.

Hoverbear:
Thanks a lot for your work on this @kleimkuhler !

PR #123, opened 2018-09-17
Release 0.4.0

Hoverbear:
Release 0.4.0.

Completes the [0.4.0 milestone](https://github.com/pingcap/raft-rs/milestone/4).

After merge @hoverbear will publish a new release and update TiKV's dependency.


Hoverbear:
@BusyJay I think that is also a good way. :) At the moment I'm just making a brief summary of each PR we merged so we never have a release with "Bug fixes and improvements" as the changelog.

I'd like to talk about release process / changelogs later with you and other members of the team, as I'd like to put in place an established release process for all our libaries (grpc, prometheus, fail, raft, etc) by the end of the year. But first I'd like to put together some document.

PR #124, opened 2018-09-18
Use the newly available stable clippy. Fix lints.

Hoverbear:
We can now use `clippy` on stable so we don't need it on nightly.

* We now recommend contributors to use stable instead of having nightly as well for clippy.
* We now run clippy on stable instead of nightly.
* Nightly is still tested for nightly users.


Hoverbear:
Blocked on #123

PR #139, opened 2018-11-08
[WIP] Prevent unbounded Raft log growth.

utsl42:
Issue #131

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -178,6 +178,11 @@ pub struct Raft<T: Storage> {
 
     /// Tag is only used for logging
     tag: String,
+
+    /// Limits the aggregate byte size of the uncommitted entries that may be appended to a leader's
+    /// log. Once this limit is exceeded, proposals will begin to return ErrProposalDropped errors.
+    /// Note: 0 for no limit.
+    max_uncommitted_entries_size: usize,
 }
 
 trait AssertSend: Send {}
@@ -258,6 +263,7 @@ impl<T: Storage> Raft<T> {
             max_election_timeout: c.max_election_tick(),
             skip_bcast_commit: c.skip_bcast_commit,
             tag: c.tag.to_owned(),
+            max_uncommitted_entries_size: c.max_uncommitted_entries_size,
         };
         for p in peers {
             let pr = Progress::new(1, r.max_inflight);
@@ -379,6 +385,14 @@ impl<T: Storage> Raft<T> {
     #[inline]
     pub fn skip_bcast_commit(&mut self, skip: bool) {
         self.skip_bcast_commit = skip;
+    }
+
+    #[inline]
+    fn get_max_uncommitted_size(&self) -> usize {
+        if self.state == StateRole::Leader {
+            return self.max_uncommitted_entries_size
+        }
+        return 0
     }
 
     // send persists state to stable storage and then sends to its mailbox.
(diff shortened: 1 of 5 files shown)


src/raft.rs
```diff
@@ -381,6 +387,14 @@ impl<T: Storage> Raft<T> {
         self.skip_bcast_commit = skip;
     }
 
+    #[inline]
+    fn max_uncommitted_size(&self) -> usize {
+        if self.state == StateRole::Leader {
```
Hoverbear: Can you please document why this conditional is needed?

src/raft.rs
```diff
@@ -625,20 +639,22 @@ impl<T: Storage> Raft<T> {
 
     /// Appends a slice of entries to the log. The entries are updated to match
     /// the current index and term.
-    pub fn append_entry(&mut self, es: &mut [Entry]) {
+    pub fn append_entry(&mut self, es: &mut [Entry]) -> Result<()> {
```
Hoverbear: Can you add a test that makes sure any future changes don't make it so the `Raft` is modified in an `append_entry` that results in an error?

src/raft_log.rs
```diff
@@ -481,6 +499,38 @@ impl<T: Storage> RaftLog<T> {
         self.committed = snapshot.get_metadata().get_index();
         self.unstable.restore(snapshot);
     }
+
+    /// increase_uncommitted_size computes the size of the proposed entries and
+    /// determines whether they would push leader over its max_uncommitted_entries_size limit.
+    /// If the new entries would exceed the limit, the method returns false. If not,
+    /// the increase in uncommitted entry size is recorded and the method returns
+    /// true.
```
Hoverbear: This needs a `# Panics` section which details that passing `0` as `max` will cause a panic.

src/raft.rs
utsl42: Main reason was that the Go change this is all based on decided they only need to apply this limit at the current leader.

I had thought this particular accessor would wind up getting used in more than one place. That turns out to not be the case, so it could be just hoisted into that function.

src/raft_log.rs
utsl42: I only added the panic to help me chase down an annoying bug where I broke all the tests. Can just pull it out.

src/raft.rs
utsl42: I can write a test that exercises this particular error condition, and makes sure it doesn't modify the raft log. There's no way to ensure that any hypothetical error conditions added in future changes also behave that way, since there's no way to ensure the test triggers those errors.

With this change, `append` and `maybe_append` both apply different checks and conditionally append to the log. It seems like it would be useful to more clearly document what their distinct purposes are.

src/raft_log.rs
```diff
@@ -252,6 +261,13 @@ impl<T: Storage> RaftLog<T> {
                 self.last_index()
             )
         }
+        let entries = self
+            .slice(
```
utsl42: Could this use `unstable_entries` instead of `slice`? If so, the unwrap would become unnecessary here.

src/raft.rs
Hoverbear: Improving the documentation would be good. :) Are you suggesting doing that in this PR or as future work?

src/raft_log.rs
Hoverbear: Ok. :)

src/raft.rs
utsl42: I don't know this code base well enough to do that. More of a suggestion for someone that does fully understand the intentions behind why those are separate.

src/raft_log.rs
Hoverbear: I worry in this situation that there would be entries written to stable storage, but not yet committed.

You can see here: https://docs.rs/raft/0.4.0/raft/raw_node/struct.Ready.html#structfield.committed_entries

> CommittedEntries specifies entries to be committed to a store/state-machine. These have previously been committed to stable store.

Meanwhile: https://docs.rs/raft/0.4.0/raft/raw_node/struct.Ready.html#structfield.entries

> Entries specifies entries to be saved to stable storage BEFORE Messages are sent.

So if we only rely on `unstable_entries` I think it means we might not see all the uncommitted entries.

src/raft.rs
Hoverbear: Ok! :ok_hand: I am still learning it too. Let's put this to the side for now.

Hoverbear:
Hi @utsl42! Are you still working on this?

utsl42:
Have had to pause for a bit, but I do plan to pick it back up later, likely
in January. At this point, I think it mostly just needs tests.

Hoverbear:
Hi @utsl42, still thinking of picking this back up?

652h:
@Hoverbear @utsl42 do you whant to finish it by youself or can i try to help you?

Hoverbear:
@652h If you wanted to open up a revival PR I'd be very happy to help shepherd it to merge! I was hoping to revisit this soon but haven't have a moment.

BusyJay:
This was implemented as #398, close it now, thanks!

PR #161, opened 2018-12-26
implement CommittedEntries pagination

Fullstop000:
Fixes https://github.com/pingcap/raft-rs/issues/105

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -348,6 +348,75 @@ fn test_progress_paused() {
 }
 
 #[test]
+fn test_progress_flow_control() {
+    let l = default_logger();
+    let mut cfg = new_test_config(1, 5, 1);
+    cfg.max_inflight_msgs = 3;
+    cfg.max_size_per_msg = 2048;
+    let s = MemStorage::new_with_conf_state((vec![1, 2], vec![]));
+    let mut r = new_test_raft_with_config(&cfg, s, &l);
+    r.become_candidate();
+    r.become_leader();
+
+    // Throw away all the messages relating to the initial election.
+    r.read_messages();
+
+    // While node 2 is in probe state, propose a bunch of entries.
+    r.mut_prs().get_mut(2).unwrap().become_probe();
+    let data: String = std::iter::repeat('a').take(1000).collect();
+    for _ in 0..10 {
+        let msg = new_message_with_entries(
+            1,
+            1,
+            MessageType::MsgPropose,
+            vec![new_entry(0, 0, Some(&data))],
+        );
+        r.step(msg).unwrap();
+    }
+
+    let mut ms = r.read_messages();
+    // First append has two entries: the empty entry to confirm the
+    // election, and the first proposal (only one proposal gets sent
+    // because we're in probe state).
+    assert_eq!(ms.len(), 1);
+    assert_eq!(ms[0].msg_type, MessageType::MsgAppend);
+    assert_eq!(ms[0].entries.len(), 2);
+    assert_eq!(ms[0].entries[0].data.len(), 0);
+    assert_eq!(ms[0].entries[1].data.len(), 1000);
+
+    // When this append is acked, we change to replicate state and can
+    // send multiple messages at once.
+    let mut msg = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+    msg.index = ms[0].entries[1].index;
+    r.step(msg).unwrap();
+    ms = r.read_messages();
+    assert_eq!(ms.len(), 3);
+    for (i, m) in ms.iter().enumerate() {
+        if m.msg_type != MessageType::MsgAppend {
+            panic!("{}: expected MsgAppend, got {:?}", i, m.msg_type);
+        }
+        if m.entries.len() != 2 {
+            panic!("{}: expected 2 entries, got {}", i, m.entries.len());
+        }
+    }
+
+    // Ack all three of those messages together and get the last two
+    // messages (containing three entries).
+    let mut msg = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+    msg.index = ms[2].entries[1].index;
+    r.step(msg).unwrap();
+    ms = r.read_messages();
+    assert_eq!(ms.len(), 2);
+    for (i, m) in ms.iter().enumerate() {
+        if m.msg_type != MessageType::MsgAppend {
+            panic!("{}: expected MsgAppend, got {:?}", i, m.msg_type);
+        }
+    }
+    assert_eq!(ms[0].entries.len(), 2);
+    assert_eq!(ms[1].entries.len(), 1);
+}
+
+#[test]
 fn test_leader_election() {
     let l = default_logger();
     test_leader_election_with_config(false, &l);
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -72,6 +72,16 @@ pub struct SoftState {
     pub raft_state: StateRole,
 }
 
+#[derive(Default)]
+pub struct HandleResponseContext {
+    maybe_commit: bool,
+    send_append: bool,
+    loop_append: bool,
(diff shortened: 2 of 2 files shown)


src/raft_log.rs
```diff
@@ -76,8 +77,13 @@ where
 }
 
 impl<T: Storage> RaftLog<T> {
-    /// Creates a new raft log with a given storage and tag.
+    /// Creates a new raft log with a given storage, tag and unlimited max_msg_size.
     pub fn new(storage: T, tag: String) -> RaftLog<T> {
+        RaftLog::new_with_size(storage, tag, NO_LIMIT)
+    }
+
+    /// Creates a new raft log with a given storage, tag and max_msg_size.
```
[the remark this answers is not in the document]
Fullstop000: Probably `set_max_msg_size` can be more easier to understand but I'm fine using `with_max_msg_size` .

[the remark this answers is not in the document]
Fullstop000: Maybe `set_max_msg_size` is easier to understand as we use it like :
```rust
let mut raft_log = RaftLog::new(store, c.tag.clone());
raft_log.set_max_msg_size(c.max_size_per_msg);
```
or just
```rust
raft_log.max_msg_size = c.max_size_per_msg;
```

Fullstop000:
@Hoverbear the second test `TestAppendPagination ` is to test whether `MaxSizePerMsg` is able to cap the total byte size of entries in each message. 
Do I need to add this as I don't clearly know whether we've tested this of not ( seems not )  ?

And I find that etcd-raft has added a new config parameter `MaxCommittedSizePerReady` to control this instead of `MsgSizePerMsg`. 
https://github.com/etcd-io/etcd/commit/e4af2be5bbcbea27976e8af543defb4a7b7e1245

src/raft_log.rs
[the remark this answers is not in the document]
Hoverbear: This looks great: 

```rust
let mut raft_log = RaftLog::new(store, c.tag.clone());
raft_log.set_max_msg_size(c.max_size_per_msg);
```

The reason this is preferred over setting an attribute is that if we change the struct layout it breaks API, if we have a function and change the struct later we can gracefully deprecate. :)

Hoverbear:
@Fullstop000 Can you run `cargo fmt` for me?

tests/integration_cases/test_raw_node.rs
```diff
@@ -111,6 +111,47 @@ fn test_raw_node_step() {
...
+    raw_node.campaign().expect("");
+    let rd = raw_node.ready();
+    storage.wl().append(rd.entries()).expect("");
+    raw_node.advance(rd);
+
+    let data = "*".repeat(400).into_bytes();
+    raw_node.propose(vec![], data.clone()).expect("");
+    raw_node.propose(vec![], data.clone()).expect("");
+    raw_node.propose(vec![], data.clone()).expect("");
+
```
Hoverbear: Please add a comment to explain the test.

For example:

```
// 3 proposals will commit in two batches
```

Hoverbear:
@Fullstop000 Great catch. I think this is a more clear functionality. @hicqu can you take a look?

Fullstop000:
I'll use the new parameter `max_committed_size_per_ready` instead. And the test for `max_size_per_msg` will be added.

siddontang:
Ping @BusyJay @hicqu

src/config.rs
```diff
@@ -75,6 +77,10 @@ pub struct Config {
     /// Note: math.MaxUusize64 for unlimited, 0 for at most one entry per message.
     pub max_size_per_msg: u64,
 
+    /// Limit the size of the committed entries which can be applied.
+    /// Note: Use math.MaxUusize64 for unlimited, 0 for at most one entry per message
```
BusyJay: `math.MaxUusize64` is invalid in rust. I think NO_LIMIT is a public constant in raft crate?

Fullstop000: Could I just use `raft::{ NO_LIMIT}` instead of `math.MaxUsize64` in the comments above? ( I'm not familiar with rust comment tricks )

src/raw_node.rs
```diff
@@ -145,11 +145,21 @@ impl Ready {
...
+        let mut hs = raft.hard_state();
         if &hs != prev_hs {
             if hs.get_vote() != prev_hs.get_vote() || hs.get_term() != prev_hs.get_term() {
                 rd.must_sync = true;
             }
+            // If we hit a size limit when loading CommittedEntries, clamp
+            // our HardState.Commit to what we're actually returning. This is
+            // also used as our cursor to resume for the next Ready.
+            let committed_entries = rd.committed_entries.as_ref().unwrap();
+            if !committed_entries.is_empty() {
```
Hoverbear: A useful trick in these circumstances when you want to conditionally use some value is to use `if let`. This helps avoid the visually alarming `unwrap()` call.:

```rust
if let Some(last) = committed_entries.last() {
    if last.index < hs.get_commit() {
        hs.set_commit(last.index);
    }
}
```

Fullstop000: Oh thanks. I'll improve that.

src/raw_node.rs
```diff
@@ -145,11 +145,23 @@ impl Ready {
...
-        let hs = raft.hard_state();
+        let mut hs = raft.hard_state();
         if &hs != prev_hs {
             if hs.get_vote() != prev_hs.get_vote() || hs.get_term() != prev_hs.get_term() {
                 rd.must_sync = true;
             }
+            // If we hit a size limit when loading CommittedEntries, clamp
+            // our HardState.Commit to what we're actually returning. This is
+            // also used as our cursor to resume for the next Ready.
+            if let Some(committed_entries) = rd.committed_entries.as_ref() {
```
BusyJay: What about `rd.committed_entries.as_ref().and_then(|c| c.last())`?

tests/integration_cases/test_raw_node.rs
```diff
@@ -111,6 +111,53 @@ fn test_raw_node_step() {
...
+// test_raw_node_committed_entries_pagination ensures that max_committed_size_per_ready
+// cap the size of committed entries in each ready
+#[test]
+fn test_raw_node_committed_entries_pagination() {
+    setup_for_test();
+    let mut cfg = new_test_config(1, vec![], 10, 1);
+    cfg.max_committed_size_per_ready = 1024;
+    let storage = new_storage();
+    let mut raw_node = RawNode::new(&cfg, storage.clone(), vec![new_peer(1)]).unwrap();
+    let rd = raw_node.ready();
```
BusyJay: This three lines seem unnecessary.

tests/integration_cases/test_raft.rs
```diff
@@ -2620,6 +2620,46 @@ fn test_leader_increase_next() {
     }
 }
 
+#[test]
+fn test_send_append_pagination_with_max_size_per_msg() {
```
BusyJay: What's the purpose of this case? Maybe you can add a hook to `Network` to implement a similar case with the original PR.

Hoverbear:
Hi @Fullstop000 have you had a chance to see this feedback? Can I help you at all with it?

Fullstop000:
@Hoverbear Sorry for the delay. Kind of busy here. Start to working on this.

src/raw_node.rs
Fullstop000: OMG, it's really beautiful here to use the functional pattern.

tests/integration_cases/test_raft.rs
Fullstop000: It ensures the  `hard_state.commit` matches with last committed entry index in `Ready` if the `committed_entries` overflow the `max_committed_size_per_ready`.

tests/integration_cases/test_raw_node.rs
Fullstop000: Do you mean the 3rd line or all the lines from 121 to 123? I couldn't figure it out why the `raw_node` should be removed.

PR #169, opened 2019-01-18
Fix ReadIndex request dropping issue

Fullstop000:
Fixes https://github.com/pingcap/raft-rs/issues/167


src/raft.rs
```diff
@@ -1469,8 +1469,11 @@ impl<T: Storage> Raft<T> {
             self.send_append(from, prs.get_mut(from).unwrap());
             self.set_prs(prs);
         }
-        if let Some(to_send) = more_to_send {
-            self.send(to_send)
+        if !more_to_send.is_empty() {
+            for to_send in more_to_send.iter() {
```
BusyJay: Why not use `drain`?

tests/integration_cases/test_raft.rs
```diff
@@ -2172,26 +2172,66 @@ fn test_read_only_option_safe() {
...
+            id,
+            id,
+            MessageType::MsgReadIndex,
+            vec![new_entry(0, 0, Some(wctx[1]))],
+        );
+
+        // `pending` indicates that a `ReadIndex` request will not get through quorum checking immediately
+        // so that it remains in the `read_index_queue`
+        if pending {
+            let mut send = |msgs: Vec<Message>| {
```
BusyJay: Why not just use `Network::ignore`?

tests/integration_cases/test_raft.rs
```diff
@@ -2203,16 +2243,18 @@ fn test_read_only_option_safe() {
...
-        if rs.index != wri {
-            panic!("#{}: read_index = {}, want {}", i, rs.index, wri)
-        }
-        let vec_wctx = wctx.as_bytes().to_vec();
-        if rs.request_ctx != vec_wctx {
-            panic!(
-                "#{}: request_ctx = {:?}, want {:?}",
-                i, rs.request_ctx, vec_wctx
-            )
+        assert_eq!(read_states.len(), 2);
```
BusyJay: Better use `wctx.len()` instead of 2.

tests/integration_cases/test_raft.rs
```diff
@@ -2203,16 +2243,18 @@ fn test_read_only_option_safe() {
...
+        for (j, rs) in read_states.iter().enumerate() {
```
BusyJay: You can use `read_states.iter().zip(wctx)`.

hicqu:
LGTM.

PR #173, opened 2019-01-29
fix a test case about power cycle.

hicqu:
Without the PR, test case `leader_power_cycles_no_compaction` fails because `conf state` is not persisted before power cycle. We need to do this explicitly.

The PR also contains some code format generated by `cargo fmt`. I have checked the cargo version, seems my toolchain is correct. Seems you forget to format code? If so, let's keep those changes.


hicqu:
PTAL @Hoverbear thank you!

PR #176, opened 2019-01-30
simplify snapshot restore.

hicqu:
I think when we restore snapshots, we don't need to call `begin_membership_change`. This PR changes this behavior, and simplify some code.


hicqu:
PTAL @Hoverbear .

PR #181, opened 2019-02-11
Ease protobuf dependency to match kvproto

Hoverbear:
This is a prepare-to-release patch for 0.5.0 which will ensure there is no required follow up version change in `kvproto` (https://github.com/pingcap/kvproto/blob/master/Cargo.toml#L13) leading to https://github.com/tikv/tikv/blob/master/Cargo.toml#L118 which is matches with TiKV's https://github.com/tikv/tikv/blob/master/Cargo.toml#L72.


src/eraftpb.rs
```diff
@@ -1,4 +1,4 @@
-// This file is generated by rust-protobuf 2.2.4. Do not edit
+// This file is generated by rust-protobuf 2.2.2. Do not edit
```
BusyJay: Can it work with protobuf <2.2? For example, will code generated by new compiler use a new API missing in old version?

Hoverbear: I haven't seen a build error or test error from this. I think we could even allow "~2.0-~2.2"

Hoverbear: PTAL

PR #184, opened 2019-02-19
Update to Rust 2018

nrc:
The first commit is the update, it is all pretty straightforward. The second commit removes the rust-toolchain file because I think it is a better developer experience not to have it checked in.

@Hoverbear @brson PTAL


src/eraftpb.rs
```diff
@@ -158,7 +158,7 @@ impl ::protobuf::Message for Entry {
         true
     }
 
-    fn merge_from(&mut self, is: &mut ::protobuf::CodedInputStream) -> ::protobuf::ProtobufResult<()> {
+    fn merge_from(&mut self, is: &mut ::protobuf::CodedInputStream<'_>) -> ::protobuf::ProtobufResult<()> {
```
Hoverbear: Will these still be reflected when we run `./generate-proto.sh`?

nrc: They won't, but they are only idiomatic changes rather than breaking. To fix this we'd need to upgrade rust-protoc. Longer term, Prost will emit 2018 code when I eventually land https://github.com/danburkert/prost/pull/147

Hoverbear: @nrc Ok, is there a way to mechanize these changes when we do run `generate-proto.sh`?

nrc:
I've updated generate-proto.sh to rerun cargo fix so we don't lose the changes.

Hoverbear:
@nrc Great, that solves that problem!

brson:
LGTM

nrc:
Rebased, reviewed, ready to merge

Hoverbear:
@nrc Builds are failing at the moment. :(

nrc:
@Hoverbear fixed!

Hoverbear:
Forcing merge with LGTMs from me and @brson (there was a rebase)

PR #185, opened 2019-02-20
Migrate to Slog

nrc:
Rebased version of #52, based on top of #184


nrc:
rebased

nrc:
Blocked on landing https://github.com/pingcap/raft-rs/pull/184

Hoverbear:
@nrc 2018 is merged now. :)

nrc:
rebased again

src/lib.rs
```diff
@@ -437,3 +438,73 @@ pub mod prelude {
 
     pub use crate::read_only::{ReadOnlyOption, ReadState};
 }
+
+/// Build a logger for tests.
```
Hoverbear: These were all moved to `harness` in #127 , maybe they got lost in your rebase.

harness/src/lib.rs
```diff
@@ -33,17 +33,7 @@ This module contains various testing harness utilities for Raft.
...
-extern crate rand;
-
 mod interface;
 mod network;
 
 pub use self::{interface::Interface, network::Network};
-
-/// Do any common test initialization. Eg set up logging.
-#[doc(hidden)]
-pub fn setup_for_test() {
```
Hoverbear: Previous place of logging function after #127

src/lib.rs
```diff
@@ -437,3 +438,73 @@ pub mod prelude {
...
+    use std::sync::Once;
+    static LOGGER_INITIALIZED: Once = Once::new();
+    static mut LOGGER: Option<Logger> = None;
+
+    unsafe {
+        LOGGER_INITIALIZED.call_once(|| {
+            let decorator = slog_term::TermDecorator::new().build();
+            let drain = slog_term::CompactFormat::new(decorator).build().fuse();
+            let drain = slog_envlogger::new(drain).fuse();
+            let drain = slog_async::Async::new(drain)
```
Hoverbear: For testing having a sync logger might be preferred, this way the stdout doesn't get mangled when interwoven with the testing output.


```
test integration_cases::test_raft::test_recv_msg_request_vote ...  ok
idtest integration_cases::test_raft::test_restore_with_learner ... ok:
test integration_cases::test_raft::test_send_append_for_progress_probe ...  ok
3test integration_cases::test_raft::test_send_append_for_progress_replicate ... ok

test integration_cases::test_raft::test_send_append_for_progress_snapshot ...  ok
test integration_cases::test_raft::test_single_node_candidate ... Febok
25test integration_cases::test_raft::test_single_node_commit ...  thread 'ok10integration_cases::test_raft::test_state_transition
:' panicked at 'test integration_cases::test_raft::test_single_node_pre_candidate ... 37assertion failed: `(left != right)`
  left: `Follower`,
 right: `Follower`: invalid transition [follower -> leader]ok:',
14src/raft.rstest integration_cases::test_raft::test_step_config ... .:ok166942
:test integration_cases::test_raft::test_step_ignore_config ...  9ok

ERROnote: Run with `RUST_BACKTRACE=1` for a backtrace.
test integration_cases::test_raft::test_transfer_non_member ... ok
test integration_cases::test_raft::test_vote_from_any_state ...  can't become learner when restores snapshot [index: 11, term: 11]okthread '
integration_cases::test_raft::test_state_transition, ' panicked at 'assertion failed: `(left != right)`
  left: `Leader`,
 right: `Leader`: invalid transition [leader -> pre-candidate]snapshot_term', src/raft.rs::913 :911
, snapshot_index: 11, tag:
thread 'integration_cases::test_raft::test_state_transition' panicked at 'assertion failed: `(left != right)`
  left: `Leader`,
 right: `Leader`: invalid transition [leader -> candidate]', src/raft.rs:889:9
```

src/lib.rs
```diff
@@ -437,3 +438,73 @@ pub mod prelude {
...
+fn default_logger() -> &'static Logger {
+    use std::sync::Once;
+    static LOGGER_INITIALIZED: Once = Once::new();
+    static mut LOGGER: Option<Logger> = None;
+
+    unsafe {
+        LOGGER_INITIALIZED.call_once(|| {
+            let drain = slog_stdlog::StdLog.fuse();
+            let drain = slog_envlogger::new(drain).fuse();
+            let drain = slog_async::Async::new(drain)
```
Hoverbear: Do you think it makes sense to make the default for raft an async logger? If folks don't pass us a logger they're probably using stdlog which would be synchronous. 🤔

src/lib.rs
```diff
@@ -437,3 +438,73 @@ pub mod prelude {
...
+                .fuse();
+            LOGGER = Some(slog::Logger::root(drain, o!()));
+        });
+        LOGGER.as_ref().unwrap()
+    }
+}
+
+/// The discard logger is used in `Default::default` implementations.
+///
+/// Currently, this is a `log` adaptor.
```
Hoverbear: I  think this comment is incorrect.

src/progress.rs
```diff
@@ -175,16 +177,22 @@ pub struct ProgressSet {
...
     // We use a cell to avoid taking a `&mut self`.
     sort_buffer: RefCell<Vec<u64>>,
+    logger: Logger,
 }
 
 impl ProgressSet {
     /// Creates a new ProgressSet.
-    pub fn new() -> Self {
-        Self::with_capacity(0, 0)
+    pub fn new<'a>(logger: impl Into<Option<&'a Logger>>) -> Self {
```
Hoverbear: I wonder if having a builder `.with_logger()` would be nicer from an API point of view. 🤔

src/progress.rs
```diff
@@ -421,7 +431,7 @@ impl ProgressSet {
 
     /// Promote a learner to a peer.
     pub fn promote_learner(&mut self, id: u64) -> Result<()> {
-        debug!("Promote learner with id {}.", id);
+        info!(self.logger, "Promoting peer with id {id}", id = id);
```
Hoverbear: Quite a few of these change from `info` to `debug`, that might have been stuff from #52

Hoverbear:
@nrc There is a lot of merge conflicts, do you think you could address them?

nrc:
I've rebased and addressed some of the easier review comments (finally, sorry it took so long). I'll address the other comments tomorrow.

src/lib.rs
nrc: I think it makes sense to use async. I think if users want the default logger, they will assume they get the best option for common cases (convention over configuration philosophy). If I understand the pros and cons of slog and Raft correctly, then async is usually the better option.

I don't think it is too important to match too closely with other loggers - if they really need that constraint then users should explicitly set the logger.

nrc:
All comments addressed. @Hoverbear PTAL

src/progress/progress_set.rs
```diff
@@ -159,16 +161,22 @@ pub struct ProgressSet {
...
     // We use a cell to avoid taking a `&mut self`.
     sort_buffer: RefCell<Vec<u64>>,
+    logger: Logger,
 }
 
 impl ProgressSet {
     /// Creates a new ProgressSet.
-    pub fn new() -> Self {
-        Self::with_capacity(0, 0)
+    pub fn new<'a>(logger: impl Into<Option<&'a Logger>>) -> Self {
```
Hoverbear: How about `fn logger<'a>(&mut self, logger: impl Into<Option<&'a Logger>>) -> Self {...}`? This can help preserve API and avoid the `None` stuff we now see in some tests.

src/lib.rs
Hoverbear: So I thought hard about this, and I'm not sure this is a good idea. If they're not using `slog` and these logging macros are triggering it means they *specifically* **opted in** to syncronous logging with the `log` crate. Using async logs by default could result in out of order logs compared to the relative time of other events in their system (which are acting syncronously).

For example, consider the user is using sync `log` to run their application and are checking when things are happening inside raft to verify their assumptions. Forcing async logging here means they might see raft logs happen *after* the return of their calls, which could create a log of confusion.

I think it's *convention* to opt-in to async logging through explicit configuration. You can see slog itself forces you to opt-in! :)

nrc:
@Hoverbear I changed the API as we discussed and the default logger to be sync. PTAL

Cargo.toml
```diff
@@ -30,14 +30,18 @@ lazy_static = "1.3.0"
 prost = "0.5"
 prost-derive = "0.5"
 bytes = "0.4.11"
+slog = "2.2"
 quick-error = "1.2.2"
 rand = "0.5.4"
 hashbrown = "0.1"
 fail = { version = "0.2", optional = true }
 getset = "0.0.6"
+slog-async = "2.3.0"
```
Hoverbear: I think this can be optional or dev only now?

nrc:
Comment addressed and rebased

Hoverbear:
@nrc I notice in some messages we get double logs? 

image

nrc:
CI is fixed by https://github.com/pingcap/raft-rs/pull/242

nrc:
> @nrc I notice in some messages we get double logs?

Is this from the tests? How can I repro?

nrc:
> @nrc I notice in some messages we get double logs?

This is a feature not a bug of slog - it is printing the key/value pairs as well as inlining them into the format string. Unfortunately there is not an option to turn this off (there is a PR - https://github.com/slog-rs/term/pull/19 - but it has sat with no comment since last October). Unfortunately it is not really possible to make our own formatter to reuse the existing ones (lots of stuff is private). I think our choices are either to fork slog_term or put up with the duplication.

Hoverbear:
@nrc So why not just say `DEBG Sending from 2 to 1, msg: {}` and letting slog output the message itself instead of formatting it?

nrc:
We could remove the message from the format string and just have it as a key/value. We would still duplicate the 2 and 1 (and we'd have to change this for every format string).

nrc:
I've improved the logging strings to eliminate some duplication. I doubt things are perfect, but since we need to fix the messages on an individual basis, I think I've hit a reasonable point of diminishing returns

Hoverbear:
Whoa check out how nice the output of `RUST_TEST_THREADS=1 RUST_LOG=raft=debug cargo test -- --nocapture` is!

image

src/raft.rs
```diff
@@ -1217,15 +1284,20 @@ impl<T: Storage> Raft<T> {
...
+                pending_changes = n;
+                "tag" => &self.tag,
             );
             return;
         }
         info!(
-            "{} is starting a new election at term {}",
-            self.tag, self.term
+            self.logger,
+            "Starting a new election";
```
```suggestion
            "starting a new election";
```

src/raft.rs
```diff
@@ -1336,33 +1408,35 @@ impl<T: Storage> Raft<T> {
...
+             at term {term}",
+            log_term = self.raft_log.last_term(),
+            log_index = self.raft_log.last_index(),
+            vote = self.vote,
+            from = m.from,
+            msg_term = m.log_term,
+            msg_index = m.index,
+            term = self.term;
+            "tag" => &self.tag,
+            "msg type" => ?m.msg_type(),
```
Hoverbear: @breeswish Does this naming style seem ok with you?

src/raft.rs
```diff
@@ -1485,39 +1570,51 @@ impl<T: Storage> Raft<T> {
     fn handle_transfer_leader(&mut self, m: &Message, prs: &mut ProgressSet) {
         let from = m.from;
         if prs.learner_ids().contains(&from) {
-            debug!("{} is learner. Ignored transferring leadership", from);
+            debug!(
+                self.logger,
+                "Ignored transferring leadership";
```
```suggestion
                "ignored transferring leadership";
```

src/raft.rs
```diff
@@ -1485,39 +1570,51 @@ impl<T: Storage> Raft<T> {
...
+                last_lead_transferee = last_lead_transferee.unwrap();
+                "tag" => &self.tag,
             );
         }
         if lead_transferee == self.id {
             debug!(
-                "{} is already leader. Ignored transferring leadership to self",
-                self.tag
+                self.logger,
+                "Already leader. Ignored transferring leadership to self";
```
```suggestion
                "already leader, ignored transferring leadership to self";
```

src/raft.rs
```diff
@@ -1933,7 +2068,11 @@ impl<T: Storage> Raft<T> {
     /// For a given message, append the entries to the log.
     pub fn handle_append_entries(&mut self, m: &Message) {
         if m.index < self.raft_log.committed {
-            debug!("{} Got message with lower index than committed.", self.tag);
+            debug!(
+                self.logger,
+                "Got message with lower index than committed.";
```
```suggestion
                "got message with lower index than committed.";
```

src/raft.rs
```diff
@@ -2140,9 +2295,10 @@ impl<T: Storage> Raft<T> {
         let config = config.into();
         config.valid()?;
         debug!(
-            "Replicating SetNodes with voters ({:?}), learners ({:?}).",
-            config.voters(),
-            config.learners()
+            self.logger,
+            "Replicating SetNodes";
```
```suggestion
            "replicating SetNodes";
```

src/raft.rs
```diff
@@ -2172,8 +2328,10 @@ impl<T: Storage> Raft<T> {
     /// * There is a pending membership change. (See `is_in_membership_change()`)
     fn add_voter_or_learner(&mut self, id: u64, learner: bool) -> Result<()> {
         debug!(
-            "Adding node (learner: {}) with ID {} to peers.",
-            learner, id
+            self.logger,
+            "Adding node (learner: {learner}) with ID {id} to peers.",
```
```suggestion
            "adding node (learner: {learner}) with ID {id} to peers.",
```

src/raft_log.rs
```diff
@@ -184,11 +214,12 @@ impl<T: Storage> RaftLog<T> {
             if !self.match_term(e.index, e.term) {
                 if e.index <= self.last_index() {
                     info!(
-                        "{} found conflict at index {}, [existing term:{}, conflicting term:{}]",
-                        self.tag,
-                        e.index,
-                        self.term(e.index).unwrap_or(0),
-                        e.term
+                        self.logger,
+                        "{tag} found conflict at index {index}, [existing term: {existing_term}, conflicting term: {conflicting_term}]",
```
Hoverbear: Seems `existing_term` and `conflicting_term` could be part of the structured data here.

src/raft_log.rs
```diff
@@ -425,7 +457,7 @@ impl<T: Storage> RaftLog<T> {
     /// Attempts to commit the index and term and returns whether it did.
     pub fn maybe_commit(&mut self, max_index: u64, term: u64) -> bool {
         if max_index > self.committed && self.term(max_index).unwrap_or(0) == term {
-            debug!("Committing index {}", max_index);
+            debug!(self.logger, "Committing index {index}", index = max_index);
```
Hoverbear:
```suggestion
            debug!(self.logger, "committing index {index}", index = max_index);
```

tests/integration_cases/test_membership_changes.rs
```diff
@@ -641,25 +645,33 @@ mod three_peers_replace_voter {
...
 
-        info!("Compacting leader's log");
+        info!(l, "Compacting leader's log");
         // This snapshot has a term 1.
         let snapshot = {
             let peer = scenario.peers.get_mut(&1).unwrap();
-            warn!("BLAH {:?}", peer.pending_membership_change().clone());
+            warn!(
+                l,
+                "BLAH";
```
Hoverbear: Oh this can probably be removed. 🤣

tests/integration_cases/test_membership_changes.rs
```diff
@@ -641,25 +645,33 @@ mod three_peers_replace_voter {
...
             peer.raft_log.store.wl().commit_to_and_set_conf_states(
                 3,
                 ConfState::from(peer.prs().configuration().clone()).into(),
                 peer.pending_membership_change().clone(),
             )?;
             let snapshot = peer.raft_log.snapshot()?;
-            warn!("BLAH {:?}", snapshot.get_metadata());
+            warn!(
+                l,
+                "BLAH";
```
Hoverbear: Oh this can probably be removed. 🤣

Hoverbear:
@breeswish PTAL

PR #189, opened 2019-02-22
added boilerplate to make doctests pass

KeeganMyers:
Closes https://github.com/pingcap/raft-rs/issues/70.


src/lib.rs
```diff
@@ -187,7 +187,22 @@ When your Raft node is ticked and running, Raft should enter a `Ready` state. Yo
...
+# };
+
+# let config = Config {
+#    id: 1,
+#    peers: vec![1],
+#     ..Default::default()
+# };
+
+# let storage = MemStorage::default();
+# config.validate().unwrap();
```
Hoverbear: It's ok to skip the `validate` call. It doesn't have an effect unless you're using the returned value.

src/lib.rs
```diff
@@ -187,7 +187,22 @@ When your Raft node is ticked and running, Raft should enter a `Ready` state. Yo
...
 state:
 
-```rust,ignore
+```
+# use raft::{
+#    Config,
+#    storage::MemStorage,
+#    raw_node::RawNode,
+# };
+
```
Hoverbear: These extra newlines do show up, please trim them out.

Hoverbear:
This PR makes me so happy! Thank you for doing this @KeeganMyers ! A couple small nits to fix, but great job. :)

src/lib.rs
Hoverbear: (@KeeganMyers I notice these are still here! Is it maybe a mistake?)

Hoverbear:
Closed in favor of #194.

Hoverbear:
Thanks for this contribution @KeeganMyers :) Seems you're quite busy so I'll get this sheparded to merge.

PR #193, opened 2019-03-05
Make conf change effective on received

hicqu:
This PR contains 3 major changes:
1. Make conf change effective on received;
2. Refactor MemStorage to make it more friendly to write tests;
3. Simplify `RawNode::new` and `Raft::new` to unify all instances in tests.

These makes the PR seems bigger than expected. I hope we can merge them in one PR, so let me explain why 3 parts are necessary, and which files they change:

### Part 1 make conf change effective on received
* Add `conf_states: Vec<ConfStateWithIndex>` in `Raft`
  Why we need to store a vector? Because snapshots are created based on *applied index*, not *last index*, so we need to store several configurations, and only the fitst one's index is less than *applied index*.
* Add method `recover_conf_states_ahead` in `Raft`
  It's for configuration changes on followers are overwritten by the leader with higher term.
* Add `handle_conf_changes_after_append` in `Raft`
  Available for both leader and follower. Can be called after entries are replicated.
* Extract `maybe_append` into `Raft::handle_append_entries`
  It's for we can call `handle_conf_changes_after_append` at right time.

### Part 2 refactor `MemStorage`
* Add `conf_states` in `RaftState`, just like in `Raft`
* Document `first_index` and `last_index` more clearly
  Internal implement of `RaftLog` depends some behaviors of them.
* Persist configuration changes when `append` is called
  Same as what we do in `Raft`.
* Make `snapshot` and `apply_snapshot` more clearly

### Part 3 simplify `RawNode::new` and `Raft::new`
Previously we can have 3 different `RawNode` in tests:
> 1 is created with empty `Config`, in which case configurations in both memory and storage are empty.
> 2 is created with no-empty `Config`, but empty peer list, in which case configuration in memory is not empty, but configuration in storage is empty.
> 3 is created with no-empty `Config` and no-empty peer list, in which case configurations in both memory and storage are not empty.

As you can see, the (2) is an illegal state exactly, because only things in memory is ok, and we can't sync them to stroage with `Ready`. When I update test cases, I'm very confused with those `RawNode`s created by (2). After struggle some days with it, I decide to remove the illegal thing from all test cases. I remove the 3rd parameter in `RawNode::new`, and the created `RawNode` will have consistent states in storage and memory. The main change about this is:
* A new function `new_mem_raw_node(&Config, MemStorage) -> RawNode`
  if `peers` and/or `learners` in `Config` is not empty, in the function the `MemStorage` will be initialized with the given configuration. After the initialization, the `first_index` of the `MemStorage` will be `2` instead of `1`.

However, *there are many test cases which do some assertions on indices need to be changed* because since the `first_index` changes. They are:
* Test cases in integration_tests/test_raft.rs
* Test cases in integration_tests/test_raft_paper.rs
* Test cases in integration_tests/test_raw_node.rs

But I think it's not hard to review these files, because all changed lines are just about *indices*. And I think it's worth because it can really help us to write more correct test cases.

### How to review this PR
I needs your help to merge all parts of this PR into master branch.
Could you help me to review changes in `src/raft.rs`, `src/progresses.rs` and `tests/integration_caes/test_membership_change.rs`? @Hoverbear 

Cold you help me to review changes in `src/storage.rs`, `src/raft_log.rs` and `tests/integration_cases/test_raft.rs`? @overvenus 

Could you help me to review changes in `tests/integration_cases/{test_raft_paper.rs, test_raw_node.rs`? @BusyJay


hicqu:
PTAL @Hoverbear @BusyJay @overvenus , thank you!

src/raw_node.rs
```diff
@@ -220,46 +216,15 @@ pub struct RawNode<T: Storage> {
 impl<T: Storage> RawNode<T> {
     #[allow(clippy::new_ret_no_self)]
     /// Create a new RawNode given some [`Config`](../struct.Config.html) and a list of [`Peer`](raw_node/struct.Peer.html)s.
-    pub fn new(config: &Config, store: T, mut peers: Vec<Peer>) -> Result<RawNode<T>> {
+    pub fn new(config: &Config, store: T) -> Result<RawNode<T>> {
```
Hoverbear: This is a break in public API (and a big one) so we'll need to make this a `0.6.0` release.

src/raft.rs
```diff
@@ -1222,102 +1256,20 @@ impl<T: Storage> Raft<T> {
...
-            return Err(Error::ViolatesContract(
-                "!ConfChange::has_start_index()".into(),
-            ));
-        };
-
-        self.set_pending_membership_change(conf_change.clone());
-        let max_inflights = self.max_inflight;
-        self.mut_prs()
-            .begin_membership_change(configuration, Progress::new(1, max_inflights))?;
+    fn begin_membership_change(&mut self, cs: &ConfState) -> Result<()> {
```
Hoverbear: It's OK to rewrite/correct the documentation on these functions. They're non-trivial and could use explanation for our contributors.

src/raft.rs
```diff
@@ -1222,102 +1256,20 @@ impl<T: Storage> Raft<T> {
...
-            } else {
-                // It's no longer safe to lookup the ID in the ProgressSet, remove it.
-                self.leader_id = INVALID_ID;
-            }
-        }
-
+    fn finalize_membership_change(&mut self) -> Result<()> {
+        assert!(self.is_in_membership_change());
+        // Here we can't call `become_follower` because we need to bcast the entry later.
+        // Call that function will cause wrong `next_idx`s.
```
```suggestion
        // Calling that function will cause wrong `next_idx`s.
```

Really great comment/explanation. 👍

src/raft.rs
```diff
@@ -2107,19 +2096,21 @@ impl<T: Storage> Raft<T> {
     ///
     /// ```rust
     /// use raft::{Raft, Config, storage::MemStorage, eraftpb::ConfState};
-    /// let config = Config {
+    /// use raft::raw_node::new_mem_raw_node;
```
Hoverbear: We shouldn't show users how to use our library with testing only functions.

Please revert this.

src/raft.rs
```diff
@@ -2107,19 +2096,21 @@ impl<T: Storage> Raft<T> {
...
+    /// let mut config = Config {
     ///     id: 1,
     ///     peers: vec![1],
     ///     ..Default::default()
     /// };
-    /// let mut raft = Raft::new(&config, MemStorage::default()).unwrap();
+    /// let mut node = new_mem_raw_node(&mut config, MemStorage::default()).unwrap();
```
Hoverbear: We shouldn't show users how to use our library with testing only functions.

Please revert this.

src/storage.rs
```diff
@@ -96,167 +151,230 @@ pub trait Storage {
...
+    }
+
+    fn inner_last_index(&self) -> u64 {
+        self.entries.last().unwrap().get_index()
+    }
 
-        let offset = self.entries[0].get_index();
-        if idx > self.inner_last_index() {
+    /// Overwrites the contents of this Storage object with those of the given snapshot.
+    pub fn apply_snapshot(&mut self, mut snapshot: Snapshot) {
```
Hoverbear: Oh, these are safe now?  What if a snapshot is out of date?

src/storage.rs
```diff
@@ -96,167 +151,230 @@ pub trait Storage {
...
     /// Discards all log entries prior to compact_index.
     /// It is the application's responsibility to not attempt to compact an index
     /// greater than RaftLog.applied.
-    pub fn compact(&mut self, compact_index: u64) -> Result<()> {
-        let offset = self.entries[0].get_index();
-        if compact_index <= offset {
-            return Err(Error::Store(StorageError::Compacted));
+    pub fn compact(&mut self, compact_index: u64) {
+        if compact_index <= self.inner_first_index() {
+            return;
```
Hoverbear: This used to be an error state. What changed?

examples/five_mem_node/main.rs
```diff
@@ -222,17 +218,7 @@ fn on_ready(
...
-                cc.merge_from_bytes(entry.get_data()).unwrap();
-                let node_id = cc.get_node_id();
-                match cc.get_change_type() {
-                    ConfChangeType::AddNode => raft_group.raft.add_node(node_id).unwrap(),
-                    ConfChangeType::RemoveNode => raft_group.raft.remove_node(node_id).unwrap(),
-                    ConfChangeType::AddLearnerNode => raft_group.raft.add_learner(node_id).unwrap(),
-                    ConfChangeType::BeginMembershipChange
-                    | ConfChangeType::FinalizeMembershipChange => unimplemented!(),
-                }
+                // For conf change messages, nothing need to do.
```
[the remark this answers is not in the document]
hicqu: They won't. Now all configuration changes become effective on received.

src/raw_node.rs
hicqu: I agree if you don't oppose to remove the parameter.

proto/eraftpb.proto
```diff
@@ -102,11 +111,13 @@ message ConfChange {
...
     // Used in `AddNode`, `RemoveNode`, and `AddLearnerNode`.
     uint64 node_id = 3;
+
+    // Some extra information used in configuration change.
     bytes context = 4;
+
     // Used in `BeginMembershipChange` and `FinalizeMembershipChange`.
     ConfState configuration = 5;
-    // Used in `BeginMembershipChange` and `FinalizeMembershipChange`.
-    // Because `RawNode::apply_conf_change` takes a `ConfChange` instead of an `Entry` we must
```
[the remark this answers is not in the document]
hicqu: I thought it's only used in `BeginMembershipChange`. And I planed to remove `RawNode::apply_conf_change` in this PR, but I found many test cases need it, so I keep it.
So I will keep the second line comment, and please help me to think about the field is needed or not for `FinalizeMembershipChange`, thank you! :)

src/storage.rs
hicqu: Panic. I will document it.

src/storage.rs
hicqu: Because I don't think it's an error, and no test cases depend on the behavior. So I change it to a more simple way. As you can see all functions of `MemStorageCore` don't return `Result` now, they just exit with success or panic. I think it can save more brain when we write test cases.

examples/five_mem_node/main.rs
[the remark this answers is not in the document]
Hoverbear: This changes the semantics of our Add/Remove/AddLearner then. Do we need to prove that it can still work correctly with TLA+? What do you think @BusyJay ?

src/raw_node.rs
Hoverbear: I support removing it if it's only meant for testing.

src/storage.rs
Hoverbear: Is it possible to keep it a `Result`? I would prefer to avoid any non-critical panics.

tests/integration_cases/test_membership_changes.rs
```diff
@@ -1034,50 +512,33 @@ mod three_peers_to_five_with_learner {
...
-            &[4, 5, 6],
-            2,
+            &[1, 2, 3, 4, 5, 6],
+            index,
             ConfChangeType::BeginMembershipChange,
         );
-        scenario.assert_in_membership_change(&[1, 2, 3, 4, 5, 6]);
+
+        assert!(!scenario.peers.get(&4).unwrap().is_learner);
+        assert!(!scenario.peers.get(&4).unwrap().is_learner);
```
Hoverbear: Duplicated line

tests/integration_cases/test_membership_changes.rs
```diff
@@ -1087,58 +548,36 @@ mod three_peers_to_five_with_learner {
...
+            &[1, 2, 4, 5, 6],
+            index,
             ConfChangeType::BeginMembershipChange,
         );
-        scenario.assert_in_membership_change(&[1, 2, 4, 5, 6]);
-        scenario.assert_not_in_membership_change(&[3]);
 
-        scenario.expect_read_and_dispatch_messages_from(&[4, 5, 6])?;
+        assert!(!scenario.peers.get(&4).unwrap().is_learner);
+        assert!(!scenario.peers.get(&4).unwrap().is_learner);
```
Hoverbear: Duplicated line

src/raft.rs
```diff
@@ -911,25 +942,22 @@ impl<T: Storage> Raft<T> {
...
             trace!(
                 "Checking if we need to finalize again..., began: {}, applied: {}, committed: {}",
                 index,
                 self.raft_log.applied,
                 self.raft_log.committed
             );
-            if index <= self.raft_log.committed {
+            if index <= self.raft_log.applied {
+                // Keep compatible with the another caller in `commit_apply`.
                 self.append_finalize_conf_change_entry();
```
hicqu: Please check it @BusyJay .

src/raft.rs
```diff
@@ -1013,6 +1041,12 @@ impl<T: Storage> Raft<T> {
 
         if m.get_term() == 0 {
             // local message
+            if m.get_msg_type() == MessageType::MsgRequestVote
```
hicqu: I think this part can be removed.

src/raft.rs
```diff
@@ -1936,35 +1898,41 @@ impl<T: Storage> Raft<T> {
...
+
+        if self.raft_log.match_term(m.get_index(), m.get_log_term()) {
+            let ents = m.get_entries();
+            let conflict_idx = self.raft_log.find_conflict(ents);
+            if conflict_idx != 0 {
+                let append_start = conflict_idx - (m.get_index() + 1);
+                let append_ents = &ents[append_start as usize..];
+                self.raft_log.append(append_ents);
+                self.recover_conf_states_ahead(append_ents[0].get_index());
+                self.handle_conf_changes_after_append(append_ents).unwrap();
```
hicqu: Remove the unwrap.

hicqu:
1. seperate it into 2 PRs.
2. Let MemStorage retruns Result instead of panic.
3. Remove new unwrap from raft.rs.

PR #195, opened 2019-03-11
refactor MemStorage and adjust some tests

hicqu:
Extracted from #193 .
This PR makes `MemStorage` more like a real `Storage`. Some behaviors are changed:
1. `create_snapshot` don't need an `index` as parameter. It uses `committed` index directly.
2. `compact` now can compact all entries. Previously the last one will be always reserved.

### DIFF
--- a/src/raft_log.rs
+++ b/src/raft_log.rs
@@ -1327,13 +1327,13 @@ mod test {
             (1000, vec![300, 299], vec![700, 700], false),
         ];
 
-        for (i, &(last_index, ref compact, ref wleft, wallow)) in tests.iter().enumerate() {
+        for (i, &(index, ref compact, ref wleft, wallow)) in tests.iter().enumerate() {
             let store = MemStorage::new();
-            for i in 1u64..last_index {
+            for i in 1u64..index {
                 store.wl().append(&[new_entry(i, 0)]).expect("");
             }
             let mut raft_log = new_raft_log(store);
-            raft_log.maybe_commit(last_index - 1, 0);
+            raft_log.maybe_commit(index - 1, 0);
             let committed = raft_log.committed;
             #[allow(deprecated)]
             raft_log.applied_to(committed);
--- a/src/raft_log.rs
+++ b/src/raft_log.rs
@@ -1341,7 +1341,9 @@ mod test {
             for (j, idx) in compact.into_iter().enumerate() {
                 let res =
                     panic::catch_unwind(AssertUnwindSafe(|| raft_log.store.wl().compact(*idx)));
-                assert!(should_panic ^ res.is_ok());
+                if !(should_panic ^ res.is_ok()) {
+                    panic!("#{}: should_panic: {}, but got: {:?}", i, should_panic, res);
+                }
                 if !should_panic {
                     let l = raft_log.all_entries().len();
                     if l != wleft[j] {
--- a/src/raft_log.rs
+++ b/src/raft_log.rs
@@ -685,7 +685,11 @@ mod test {
         #[allow(deprecated)]
         raft_log.applied_to(committed);
         let offset = 500u64;
-        raft_log.store.wl().compact(offset).expect("compact failed");
+        raft_log
+            .store
+            .wl()
+            .compact(offset - 1)
+            .expect("compact failed");
 
         assert_eq!(last_index, raft_log.last_index());
 
@@ -1321,10 +1325,10 @@ mod test {
                 1000,
                 vec![300, 500, 800, 900],
                 vec![700, 500, 200, 100],
-                true,
+                false,
             ),
             // out of lower bound
-            (1000, vec![300, 299], vec![700, 0], false),
+            (1000, vec![300, 299], vec![700, 700], false),
         ];
 
         for (i, &(last_index, ref compact, ref wleft, wallow)) in tests.iter().enumerate() {
--- a/src/raft_log.rs
+++ b/src/raft_log.rs
@@ -1316,7 +1316,7 @@ mod test {
         setup_for_test();
         let tests = vec![
             // out of upper bound
-            (1000, vec![1001u64], vec![0usize], false),
+            (1000, vec![1001u64], vec![0usize], true),
             (
                 1000,
                 vec![300, 500, 800, 900],
@@ -1327,7 +1327,7 @@ mod test {
             (1000, vec![300, 299], vec![700, 700], false),
         ];
 
-        for (i, &(index, ref compact, ref wleft, wallow)) in tests.iter().enumerate() {
+        for (i, &(index, ref compact, ref wleft, should_panic)) in tests.iter().enumerate() {
             let store = MemStorage::new();
             for i in 1u64..index {
                 store.wl().append(&[new_entry(i, 0)]).expect("");
@@ -1341,22 +1341,12 @@ mod test {
             for (j, idx) in compact.into_iter().enumerate() {
                 let res =
                     panic::catch_unwind(AssertUnwindSafe(|| raft_log.store.wl().compact(*idx)));
-                if res.is_err() {
-                    if wallow {
-                        panic!("#{}: has_panic = true, want false: {:?}", i, res);
+                assert!(should_panic ^ res.is_ok());
+                if !should_panic {
+                    let l = raft_log.all_entries().len();
+                    if l != wleft[j] {
+                        panic!("#{}.{} len = {}, want {}", i, j, l, wleft[j]);
                     }
-                    continue;
-                }
-                let compact_res = res.unwrap();
-                if let Err(e) = compact_res {
-                    if wallow {
-                        panic!("#{}.{} allow = false, want true, error: {}", i, j, e);
-                    }
-                    continue;
-                }
-                let l = raft_log.all_entries().len();
-                if l != wleft[j] {
-                    panic!("#{}.{} len = {}, want {}", i, j, l, wleft[j]);
                 }
             }
         }
--- a/src/raft_log.rs
+++ b/src/raft_log.rs
@@ -685,11 +685,7 @@ mod test {
         #[allow(deprecated)]
         raft_log.applied_to(committed);
         let offset = 500u64;
-        raft_log
-            .store
-            .wl()
-            .compact(offset - 1)
-            .expect("compact failed");
+        raft_log.store.wl().compact(offset).expect("compact failed");
 
         assert_eq!(last_index, raft_log.last_index());
 
@@ -1333,11 +1329,11 @@ mod test {
 
         for (i, &(last_index, ref compact, ref wleft, wallow)) in tests.iter().enumerate() {
             let store = MemStorage::new();
-            for i in 1u64..=last_index {
+            for i in 1u64..last_index {
                 store.wl().append(&[new_entry(i, 0)]).expect("");
             }
             let mut raft_log = new_raft_log(store);
-            raft_log.maybe_commit(last_index, 0);
+            raft_log.maybe_commit(last_index - 1, 0);
             let committed = raft_log.committed;
             #[allow(deprecated)]
             raft_log.applied_to(committed);
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -234,6 +234,10 @@ impl MemStorageCore {
     /// Discards all log entries prior to compact_index.
     /// It is the application's responsibility to not attempt to compact an index
     /// greater than RaftLog.applied.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `compact_index` is higher than `Storage::last_index(&self) + 1`.
     pub fn compact(&mut self, compact_index: u64) -> Result<()> {
         if compact_index <= self.inner_first_index() {
             // Don't need to treat this case as an error.
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -65,7 +65,7 @@ pub struct RaftState {
 }
 
 impl RaftState {
-    /// Indicates the `RaftState` is initialized or not.
+    /// Indicates if the `RaftState` is initialized or not.
     pub fn initialized(&self) -> bool {
         !self.conf_states.is_empty()
     }
@@ -127,7 +127,7 @@ impl Default for MemStorageCore {
             // The dummy entry is used to ensure `entries` never be empty, which can simplify
             // many logics.
             entries: vec![Entry::new()],
-            // Every time applys a snapshot to the storage, the metadata will be stored here,
+            // Every time a snapshot is applied to the storage, the metadata will be stored here,
             // and a dummy entry will be pushed to `entries`.
             snapshot_metadata: Default::default(),
         }
@@ -295,11 +295,7 @@ impl MemStorageCore {
             return Err(Error::Store(StorageError::Unavailable));
         }
 
-        let mut offset = compact_index - self.inner_first_index();
-        if self.entries[0].get_index() == self.snapshot_metadata.get_index() {
-            // For the dummy entry.
-            offset += 1;
-        }
+        let offset = compact_index - self.entries[0].get_index();
         self.entries = self.entries[offset as usize..].to_vec();
         if self.entries.is_empty() {
             // For the dummy entry.
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -289,15 +289,15 @@ impl MemStorageCore {
     /// It is the application's responsibility to not attempt to compact an index
     /// greater than RaftLog.applied.
     pub fn compact(&mut self, compact_index: u64) -> Result<()> {
-        if compact_index < self.inner_first_index() {
+        if compact_index <= self.inner_first_index() {
             return Ok(());
         }
 
-        if compact_index > self.inner_last_index() {
+        if compact_index > self.inner_last_index() + 1 {
             return Err(Error::Store(StorageError::Unavailable));
         }
 
-        let mut offset = compact_index - self.inner_first_index() + 1;
+        let mut offset = compact_index - self.inner_first_index();
         if self.entries[0].get_index() == self.snapshot_metadata.get_index() {
             // For the dummy entry.
             offset += 1;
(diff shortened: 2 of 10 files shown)


hicqu:
PTAL @Hoverbear @BusyJay @nolouch thank you!

src/raft_log.rs
```diff
@@ -1321,19 +1321,19 @@ mod test {
                 1000,
                 vec![300, 500, 800, 900],
                 vec![700, 500, 200, 100],
-                true,
+                false,
```
Hoverbear: Why this change?

hicqu: Because I think this case shouldn't panic.

hicqu:
I have no idea about why CI can't pass. Anyone can help me?

hicqu:
Thanks for @BusyJay all tests are pass!

src/storage.rs
```diff
@@ -275,6 +368,17 @@ impl MemStorage {
         }
     }
 
+    /// Create a `MemStorage` with a given `Config`.
+    pub fn initialize_with_config(&self, cfg: &Config) {
```
nolouch: This confused me.

src/storage.rs
```diff
@@ -96,166 +113,242 @@ pub trait Storage {
...
+            let cs = self.raft_state.conf_states[i].clone();
+            snapshot.mut_metadata().set_conf_state(cs.conf_state);
         }
-        self.snapshot.set_data(data);
-        Ok(&self.snapshot)
+        snapshot
     }
 
-    /// Discards all log entries prior to compact_index.
+    /// Discards all log entries prior to compact_index (included).
```
nolouch: what does the included mean? such as [0, 1, 2, 3],  it means `compact(3)` will remains [3] but allow to compact(4) or `compact(3)` will remains empty.

src/storage.rs
```diff
@@ -96,166 +113,242 @@ pub trait Storage {
...
+        Ok(())
+    }
+
+    /// Save raft logs and apply configuration chagnes in them.
+    pub fn append(&mut self, ents: &[Entry]) -> Result<()> {
+        if ents.is_empty() {
+            return Ok(());
+        }
+
+        // Gap is now allowed in raft logs.
```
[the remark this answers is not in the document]
hicqu: No, it should be `not allowed`. :(

src/storage.rs
hicqu: Not used in this PR. I can remove it.

src/storage.rs
hicqu: Stale comment. I will remove it.

src/storage.rs
```diff
@@ -31,32 +31,44 @@
...
+use crate::util::limit_size;
+
+/// Used to track the history about configuration changes.
+#[derive(Clone, Debug, Default)]
+pub struct ConfStateWithIndex {
+    /// Target configuration state.
+    pub conf_state: ConfState,
+    /// Index of the entry.
+    pub index: u64,
+    /// Indicates it's a membership change or not.
```
Hoverbear:
```suggestion
    /// Indicates if the node is undergoing a membership change, or not.
```

src/storage.rs
```diff
@@ -68,23 +80,28 @@ pub struct RaftState {
...
     fn term(&self, idx: u64) -> Result<u64>;
-    /// Returns the index of the first log entry that is
-    /// possible available via entries (older entries have been incorporated
-    /// into the latest snapshot; if storage only contains the dummy entry the
-    /// first log entry is not available).
+
+    /// Returns the index of the first log entry that is possible available via entries.
+    /// If the `Storage` is just initialized with a snapshot, `first_index` should return
+    /// the `snapshot index + 1`.
     fn first_index(&self) -> Result<u64>;
```
Hoverbear: Can we instruct what occurs when only the dummy entry is available?

src/storage.rs
```diff
@@ -96,166 +113,234 @@ pub trait Storage {
...
+            if e.get_index() == index {
+                self.raft_state.hard_state.set_term(e.get_term());
+                self.raft_state.hard_state.set_commit(index);
+                break;
+            }
         }
+        Ok(())
+    }
+
+    /// Save raft logs and apply configuration chagnes in them.
```
Hoverbear:
```suggestion
    /// Save raft logs and apply configuration changes in them.
```

src/storage.rs
```diff
@@ -96,166 +113,234 @@ pub trait Storage {
...
+    pub fn append(&mut self, ents: &[Entry]) -> Result<()> {
+        if ents.is_empty() {
+            return Ok(());
+        }
+
+        // Gap is not allowed in raft logs.
```
Hoverbear:
```suggestion
        // Gaps are not allowed in raft logs.
```

src/storage.rs
```diff
@@ -528,47 +579,41 @@ mod test {
...
+        storage.wl().compact(4).unwrap();
+        assert_eq!(storage.first_index(), Ok(4));
     }
 
     #[test]
     fn test_storage_compact() {
         setup_for_test();
         let ents = vec![new_entry(3, 3), new_entry(4, 4), new_entry(5, 5)];
         let mut tests = vec![
-            (2, Err(RaftError::Store(StorageError::Compacted)), 3, 3, 3),
```
Hoverbear: So this no longer returns an error? Why?

tests/integration_cases/test_raw_node.rs
```diff
@@ -57,22 +57,22 @@ fn conf_change(t: ConfChangeType, node_id: u64) -> ConfChange {
     cc
 }
 
-fn cmp_ready(
```
Hoverbear: I don't understand why this change. This is less flexible (you can't assert `!cmp_ready(...)` etc).

Is it required? I would hate to see a `must_cmp_ready_false(...)` or something be added in the future :(

hicqu: The reason about do this change is there are many assertions in the function, and they are chained with `&&`. So if one condition fails, we can't know which one it is. But if we do `assert` one by one in the function, we can know which condition fails easily.

src/storage.rs
hicqu: I think in this case an error is unnecessary. Please consider this case:
```rust
storage.compact(100);
storage.compact(80);
```
Firstly we compact to 100, and then compact to 80. Previously it will return an Error, but I think we can just do nothing internally, and return `Ok`. What do you think?

src/storage.rs
hicqu: Here I think we don't need to mention `dummy entry` because it's public document of `Storage`. Dummy entry is just a implementation detail of `MemStorage`.

Hoverbear: OK.

src/storage.rs
Hoverbear: I think just no-oping it is fine. :)

tests/integration_cases/test_raw_node.rs
Hoverbear: Ok, let's hope if we need such a functionality in the future we will be clever and add a boolean flag instead of re-implement :)

src/storage.rs
```diff
@@ -31,32 +31,44 @@
...
+    /// History configuration states and their indices. Not all history need to be reserved, the
+    /// `Storage` instance need only to ensure the first entry's index is less than applied index
+    /// so that when the peer sends snapshot to followers, it can choose a correct conf state.
+    ///
+    /// Entries in the field should be sorted by their indices.
+    pub conf_states: Vec<ConfStateWithIndex>,
+}
+
+impl RaftState {
+    /// Indicates the `RaftState` is initialized or not.
```
Hoverbear:
```suggestion
    /// Indicates if the `RaftState` is initialized or not.
```

src/storage.rs
```diff
@@ -96,166 +112,241 @@ pub trait Storage {
...
     fn default() -> MemStorageCore {
         MemStorageCore {
+            raft_state: Default::default(),
             // When starting from scratch populate the list with a dummy entry at term zero.
+            // The dummy entry is used to ensure `entries` never be empty, which can simplify
+            // many logics.
             entries: vec![Entry::new()],
-            hard_state: HardState::new(),
-            snapshot: Snapshot::new(),
+            // Every time applys a snapshot to the storage, the metadata will be stored here,
```
Hoverbear:
```suggestion
            // Every time a snapshot is applied to the storage, the metadata will be stored here,
```

src/storage.rs
```diff
@@ -96,166 +112,241 @@ pub trait Storage {
...
-            )
+
+        if compact_index > self.inner_last_index() + 1 {
+            return Err(Error::Store(StorageError::Unavailable));
         }
 
-        let i = (compact_index - offset) as usize;
-        let entries = self.entries.drain(i..).collect();
-        self.entries = entries;
+        let mut offset = compact_index - self.inner_first_index();
```
nolouch: maybe directly `offset =  compact_index - self.entries[0].get_index ` ?

hicqu:
PTAL @nolouch @Hoverbear thank you!

src/storage.rs
```diff
@@ -96,166 +112,237 @@ pub trait Storage {
...
+            // Every time a snapshot is applied to the storage, the metadata will be stored here,
+            // and a dummy entry will be pushed to `entries`.
+            snapshot_metadata: Default::default(),
         }
     }
 }
 
 impl MemStorageCore {
+    /// Append a configuration state. Only for tests.
+    pub fn append_conf_state(&mut self, cs: ConfStateWithIndex) {
```
BusyJay: Seems unrelated to this PR?

hicqu: It's used in `test_membership_change.rs`, for persisting configuration before power cycle.

src/storage.rs
```diff
@@ -291,64 +356,42 @@ impl MemStorage {
...
         let max_size = max_size.into();
         let core = self.rl();
-        let offset = core.entries[0].get_index();
-        if low <= offset {
+        if low < core.inner_first_index() {
             return Err(Error::Store(StorageError::Compacted));
         }
 
         if high > core.inner_last_index() + 1 {
-            panic!(
```
BusyJay: I prefer panicing here as it's unreachable. Reach this line means either storage or raft is not implemented correctly, return `Unavailable` is misleading.

src/storage.rs
```diff
@@ -126,136 +154,173 @@ impl MemStorageCore {
...
+            .position(|e| e.get_index() == first_idx)
+            .unwrap();
+        ents[i..].iter().for_each(|e| self.entries.push(e.clone()));
+        Ok(())
+    }
+
+    fn inner_first_index(&self) -> u64 {
+        let first_index = self.entries[0].get_index();
+        if self.snapshot_metadata.get_index() == first_index {
+            // The first entry's index equals to the snapshot index means the entry is "dummy"
```
BusyJay: How about removing the dummy entry completely?

src/storage.rs
```diff
@@ -126,136 +154,173 @@ impl MemStorageCore {
...
-            self.snapshot
-                .mut_metadata()
-                .set_pending_membership_change_index(idx);
+            self.raft_state.pending_conf_state = Some(cs);
+            self.raft_state.pending_conf_state_start_index = Some(idx);
+        }
+    }
+
+    /// Save raft logs and apply configuration changes in them.
+    pub fn append(&mut self, ents: &[Entry]) -> Result<()> {
```
BusyJay: Why reorder the function? It is not easy to review.

src/storage.rs
hicqu: I have tried. It proves that we need to set `snapshot_metadata` to a correct value in some cases. To avoid change those test cases, I prefer to keep it.

src/storage.rs
hicqu: Here I think return an Error is better because `entries` is a public API. Users may use the interface in a wrong way. For that maybe an Error is better than panic.

BusyJay: How about documenting it clearly in the trait method? Correctness is more important.

hicqu: Done. PTAL thanks.

src/storage.rs
```diff
@@ -68,23 +68,31 @@ pub struct RaftState {
...
     fn initial_state(&self) -> Result<RaftState>;
+
     /// Returns a slice of log entries in the range `[low, high)`.
     /// max_size limits the total size of the log entries returned if not `None`, however
     /// the slice of entries returned will always have length at least 1 if entries are
     /// found in the range.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `high` is higher than `Storage::last_index(self)`.
```
BusyJay: I think it should be higher than `storage::last_index(self) + 1`?

src/storage.rs
```diff
@@ -96,28 +104,52 @@ pub trait Storage {
...
+    pub fn hard_state(&self) -> &HardState {
+        &self.raft_state.hard_state
+    }
+
+    /// Apply to an index so that we can crate a latest snapshot from the storage.
+    pub fn apply_to(&mut self, index: u64) -> Result<()> {
+        if index < self.inner_first_index() || index > self.inner_last_index() {
+            return Err(Error::Store(StorageError::Unavailable));
+        }
+        for e in &self.entries {
```
BusyJay: I think you can jump to the index directly instead of iterating.

src/storage.rs
```diff
@@ -126,136 +158,173 @@ impl MemStorageCore {
...
+
+        if compact_index > self.inner_last_index() + 1 {
+            return Err(Error::Store(StorageError::Unavailable));
         }
 
-        let i = (compact_index - offset) as usize;
-        let entries = self.entries.drain(i..).collect();
-        self.entries = entries;
+        let offset = compact_index - self.entries[0].get_index();
+        self.entries = self.entries[offset as usize..].to_vec();
```
BusyJay: How about `self.entries.drain(..offset)`?

src/storage.rs
```diff
@@ -96,28 +108,46 @@ pub trait Storage {
...
+
+    /// Get the hard state.
+    pub fn hard_state(&self) -> &HardState {
+        &self.raft_state.hard_state
+    }
+
+    /// Apply to an index so that we can crate a latest snapshot from the storage.
+    pub fn apply_to(&mut self, index: u64) -> Result<()> {
+        if index < self.inner_first_index() || index > self.inner_last_index() {
+            return Err(Error::Store(StorageError::Unavailable));
```
BusyJay: I think it should panic here.

src/storage.rs
```diff
@@ -96,28 +108,46 @@ pub trait Storage {
...
+        }
+        let diff = (index - self.entries[0].get_index()) as usize;
```
BusyJay: What if it's called after applying a snapshot?

src/storage.rs
```diff
@@ -96,28 +108,46 @@ pub trait Storage {
...
+        self.raft_state.hard_state.set_commit(index);
```
BusyJay: Why change commit?

src/storage.rs
```diff
@@ -126,136 +156,163 @@ impl MemStorageCore {
...
-        let mut e = Entry::new();
-        e.set_term(snapshot.get_metadata().get_term());
-        e.set_index(snapshot.get_metadata().get_index());
-        self.entries = vec![e];
-        self.snapshot = snapshot;
+        self.snapshot_metadata = meta.clone();
+
+        // Update conf states.
+        self.raft_state.hard_state.set_term(term);
+        self.raft_state.hard_state.set_commit(index);
```
BusyJay: I think you need to clear `self.entries`.

src/storage.rs
hicqu: I think it's better to chagne the function name to `commit_to`. What do you think?

src/storage.rs
```diff
@@ -126,136 +164,164 @@ impl MemStorageCore {
...
-        } else if self.entries.len() as u64 == offset {
-            self.entries.extend_from_slice(te);
-        } else {
+        if self.inner_first_index() > ents[0].get_index() {
             panic!(
-                "missing log entry [last: {}, append at: {}]",
-                self.inner_last_index(),
-                te[0].get_index()
-            )
+                "overwrite compacted raft logs, compacted: {}, append: {}",
```
BusyJay: What if messages are reordered?

src/storage.rs
```diff
@@ -126,136 +164,164 @@ impl MemStorageCore {
...
+        pending_membership_change: Option<ConfChange>,
+    ) -> Result<()> {
+        if idx <= self.snapshot_metadata.get_index() {
+            return Err(Error::Store(StorageError::SnapshotOutOfDate));
+        }
+        if idx > self.inner_last_index() {
+            return Err(Error::Store(StorageError::Unavailable));
+        }
+        for e in &self.entries {
+            if e.get_index() != idx {
```
BusyJay: Loop is not necessary.

src/storage.rs
```diff
@@ -126,136 +164,164 @@ impl MemStorageCore {
...
+        ents.iter().for_each(|e| self.entries.push(e.clone()));
+        Ok(())
+    }
+
+    /// Initialize a snapshot in `MemStorageCore` with given index. Only used for tests.
+    pub fn create_snapshot(
+        &mut self,
+        idx: u64,
+        cs: Option<ConfState>,
+        pending_membership_change: Option<ConfChange>,
```
BusyJay: As a real storage, data is necessary.

src/storage.rs
hicqu: You mean `ents` is not ordered? A correct Raft ensures it's ordered.

hicqu: If you mean some raft messages are delayed in network, It's Raft's responsibility to truncate redundant entries.

src/storage.rs
hicqu: Emmm, key-value pairs (or other something after applied) are not stored in the storage. So I think it's weird to introduce `data` here. If users want to write a storage with real data stored, they can embed `MemStorage` into their own storage. What do you think?

src/storage.rs
hicqu: Fixed.

hicqu:
PTAL again @BusyJay thank you!

src/storage.rs
```diff
@@ -269,6 +269,10 @@ impl MemStorageCore {
     }
 
     /// Append the new entries to storage.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `ents` is not expected.
```
BusyJay: What does "expected" mean?

PR #196, opened 2019-03-11
unify RawNode::new for all test cases

hicqu:
Extracted from #193 . Please review #195 before this.


src/lib.rs
```diff
@@ -53,7 +53,8 @@ let storage = MemStorage::default();
 config.validate().unwrap();
 // We'll use the built-in `MemStorage`, but you will likely want your own.
 // Finally, create our Raft node!
-let mut node = RawNode::new(&config, storage, vec![]).unwrap();
+storage.initialize_with_config(&config);
```
Hoverbear: Will there be cased where a user wants to create a storage and then call `initialize_with_config` on it? Or will it always be just create and then run `initialize_with_config`? Maybe having `Storage::new_with_config` which constructs the new storage? :thinking:

src/raft_log.rs
```diff
@@ -1321,19 +1321,19 @@ mod test {
                 1000,
                 vec![300, 500, 800, 900],
                 vec![700, 500, 200, 100],
-                true,
+                false,
```
Hoverbear: Why this change?

src/raw_node.rs
```diff
@@ -220,46 +215,15 @@ pub struct RawNode<T: Storage> {
 impl<T: Storage> RawNode<T> {
     #[allow(clippy::new_ret_no_self)]
     /// Create a new RawNode given some [`Config`](../struct.Config.html) and a list of [`Peer`](raw_node/struct.Peer.html)s.
```
Hoverbear: Please update the documentation here.

src/storage.rs
```diff
@@ -126,136 +172,164 @@ impl MemStorageCore {
...
         }
+        if self.inner_last_index() + 1 < ents[0].get_index() {
+            // Gaps are not allowed in raft logs.
+            return Err(Error::Store(StorageError::LogGap));
+        }
+
+        // Remove all entries overwritten by `ents`, and truncate `ents` if need.
+        let diff = ents[0].get_index() - self.inner_first_index();
+        self.entries.drain(diff as usize..);
+        ents.iter().for_each(|e| self.entries.push(e.clone()));
```
Hoverbear: Maybe you can use `self.entries.append(ents)`?

tests/integration_cases/test_raft.rs
```diff
@@ -1076,85 +1062,37 @@ fn test_commit() {
...
+        (vec![2, 1, 1, 2], vec![empty_entry(1, 2)], 2, 1),
+        (vec![2, 1, 2, 2], vec![empty_entry(2, 2)], 2, 2),
+        (vec![2, 1, 2, 2], vec![empty_entry(1, 2)], 2, 1),
     ];
 
     for (i, (matches, logs, sm_term, w)) in tests.drain(..).enumerate() {
         let store = MemStorage::new();
-        store.wl().append(&logs).expect("");
+        let mut sm = new_test_raft(1, vec![1], 5, 1, store);
+        println!("iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii: {}", i);
```
Hoverbear: Maybe it's debugging?

tests/integration_cases/test_raft.rs
```diff
@@ -1222,29 +1160,35 @@ fn test_handle_msg_append() {
...
+        let mut sm = {
+            let mut sm = new_test_raft(1, vec![1], 10, 1, store);
+            let raft = sm.raft.take().unwrap();
+            raft.raft_log
+                .store
+                .wl()
+                .append(&[empty_entry(1, 2), empty_entry(2, 3)])
+                .unwrap();
+            let store = raft.raft_log.store;
+            new_test_raft(1, vec![1], 10, 1, store)
```
Hoverbear: Is it not possible to use the `sm` from the start of this scope?

src/raw_node.rs
```diff
@@ -32,18 +32,13 @@
...
 use protobuf::{self, RepeatedField};
 
-use super::config::Config;
-use super::errors::{Error, Result};
-use super::read_only::ReadState;
-use super::Status;
-use super::Storage;
-use super::{Raft, SoftState, INVALID_ID};
+use crate::config::Config;
+use crate::eraftpb::*;
```
nrc: nit: prefer list imports to glob imports

src/storage.rs
```diff
@@ -96,28 +116,54 @@ pub trait Storage {
...
+    pub fn hard_state(&self) -> &HardState {
+        &self.raft_state.hard_state
+    }
+
+    /// Commit to an index.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there is no such entry in raft logs.
+    pub fn commit_to(&mut self, index: u64) -> Result<()> {
```
nrc:
```suggestion
    pub fn commit_entry(&mut self, index: u64) -> Result<()> {
```

Unless this is public API elsewhere

src/storage.rs
```diff
@@ -96,28 +116,54 @@ pub trait Storage {
...
+        if index < self.inner_first_index()
+            || index > self.inner_last_index()
+            || self.entries.is_empty()
```
nrc: Could factor this into a method and turn the `if ... panic` into an assertion

src/storage.rs
```diff
@@ -126,136 +172,164 @@ impl MemStorageCore {
...
-                .set_pending_membership_change(cs);
-            self.snapshot
-                .mut_metadata()
-                .set_pending_membership_change_index(idx);
+            self.raft_state.pending_conf_state = Some(cs);
+            self.raft_state.pending_conf_state_start_index = Some(idx);
+        }
+    }
+
+    fn inner_first_index(&self) -> u64 {
```
nrc: For these `inner_` methods, what is 'inner' about them? If they are just meant to be for internal use, then I think you can drop the `inner_` prefix and rely on the fact that they are not `pub`.

src/storage.rs
```diff
@@ -126,136 +172,164 @@ impl MemStorageCore {
...
-            meta.set_pending_membership_change(pending_change.get_configuration().clone());
-            meta.set_pending_membership_change_index(pending_change.get_start_index());
+            .set_conf_state(self.raft_state.conf_state.clone());
+        if let Some(ref cs) = self.raft_state.pending_conf_state {
+            snapshot
+                .mut_metadata()
+                .set_pending_membership_change(cs.clone());
+            let i = self.raft_state.pending_conf_state_start_index.unwrap();
+            snapshot
+                .mut_metadata()
```
nrc: Can we call `mut_metadata` just once rather than every time?

src/storage.rs
nrc: I don't think so (`ents` isn't mutable) but I think you can use `self.entries.extend_from_slice(ents)`

src/storage.rs
```diff
@@ -126,136 +172,164 @@ impl MemStorageCore {
...
+        cs: Option<ConfState>,
+        pending_membership_change: Option<ConfChange>,
+    ) -> Result<()> {
+        if idx <= self.snapshot_metadata.get_index() {
+            return Err(Error::Store(StorageError::SnapshotOutOfDate));
+        }
+        if idx > self.inner_last_index() {
+            return Err(Error::Store(StorageError::Unavailable));
+        }
+        for e in &self.entries {
```
nrc: Could use `if let Some(e) = self.entries.iter().find(|e| e.get_index() == idx) { ... }`

src/storage.rs
```diff
@@ -275,6 +349,25 @@ impl MemStorage {
         }
     }
 
+    /// Create a new `MemStorage` with a given `Config`.
```
nrc: It would be good to state explicitly that the `MemStorage` has been initialised.

src/storage.rs
```diff
@@ -357,32 +432,31 @@ impl Storage for MemStorage {
...
-    use crate::storage::{MemStorage, Storage};
+    use std::panic::{self, AssertUnwindSafe};
+
     use harness::setup_for_test;
     use protobuf;
 
-    // TODO extract these duplicated utility functions for tests
+    use crate::eraftpb::{ConfState, Entry, Snapshot};
+    use crate::errors::{Error as RaftError, StorageError};
+    use crate::storage::{MemStorage, Storage};
```
nrc: Could use `use super::*;` (which is one of the few places I think using a glob import is OK :-) )

tests/integration_cases/test_membership_changes.rs
```diff
@@ -76,16 +78,16 @@ mod api {
...
+        let config = Config {
+            id: 1,
+            tag: "1".into(),
+            peers: vec![1, 2, 3],
+            learners: vec![4],
+            ..Default::default()
+        };
+        let store = MemStorage::new();
+        store.initialize_with_config(&config);
+        let mut raft = RawNode::new(&config, store)?.raft;
```
nrc: This seems like a pretty complicated way to get a `raft` object. Does this only affect tests or should we add some convenience API?

src/storage.rs
hicqu: We have `apply_to` elsewhere so that in this PR let's keep it `commit_to`? We can change all of them in an another PR later.

src/storage.rs
hicqu: We don't need to iterate it any more! :)

hicqu:
Thanks for your review! PTAL again if you are free @Hoverbear @nrc .

src/raw_node.rs
```diff
@@ -219,47 +217,16 @@ pub struct RawNode<T: Storage> {
...
-                rn.raft.add_node(peer.id)?;
-            }
-        }
-        rn.prev_ss = rn.raft.soft_state();
-        if last_index == 0 {
-            rn.prev_hs = Default::default();
-        } else {
-            rn.prev_hs = rn.raft.hard_state();
-        }
+        rn.prev_hs = rn.raft.hard_state();
```
nolouch: why no need set `prev_ss`?

Hoverbear:
(Fixed conflicts)

src/storage.rs
```diff
@@ -353,6 +358,26 @@ impl MemStorage {
...
+    /// Create a new `MemStorage` with a given `Config`. The given `Config` will be used to
+    /// initialize the storage.
+    pub fn new_with_config(cfg: &Config) -> MemStorage {
+        let store = MemStorage::new();
+        store.initialize_with_config(cfg);
+        store
+    }
+
+    /// Initialize a `MemStorage` with a given `Config`.
+    pub fn initialize_with_config(&self, cfg: &Config) {
```
BusyJay: Why not use `peers` directly? So that `peers` in `Config` can be removed completely.

Hoverbear:
@hicqu PTAL, I think Jay has a useful idea maybe.

tests/integration_cases/test_raw_node.rs
```diff
@@ -419,12 +378,15 @@ fn test_raw_node_start() {
...
     setup_for_test();
     let entries = vec![empty_entry(1, 1), new_entry(1, 2, Some("foo"))];
-    let st = hard_state(1, 1, 0);
 
-    let store = new_storage();
-    store.wl().set_hardstate(st);
-    store.wl().append(&entries).expect("");
-    let mut raw_node = new_raw_node(1, vec![], 10, 1, store, vec![]);
+    let mut raw_node = {
+        let raw_node = new_raw_node(1, vec![], 10, 1, new_storage());
```
BusyJay: Why not just create a storage directly?

src/raft.rs
```diff
@@ -1260,9 +1248,8 @@ impl<T: Storage> Raft<T> {
         };
 
         self.set_pending_membership_change(conf_change.clone());
-        let max_inflights = self.max_inflight;
-        self.mut_prs()
-            .begin_membership_change(configuration, Progress::new(1, max_inflights))?;
+        let pr = Progress::new(self.raft_log.last_index() + 1, self.max_inflight);
```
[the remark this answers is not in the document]
hicqu: Keep same with `add_voter` and `and_learner`. BTW some test cases will fail because on leader entry 1 is considered as "compacted" so that followers need to retrieve snapshots. But in test code we don't handle snapshots.

tests/integration_cases/test_raw_node.rs
```diff
@@ -81,22 +65,33 @@ fn new_raw_node(
...
 }
 
 // test_raw_node_step ensures that RawNode.Step ignore local message.
 #[test]
 fn test_raw_node_step() {
     setup_for_test();
     for msg_t in MessageType::values() {
-        let mut raw_node = new_raw_node(1, vec![], 10, 1, new_storage(), vec![new_peer(1)]);
+        if vec![
+            // Vote messages with term 0 will cause panics.
```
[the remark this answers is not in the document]
hicqu: In the past `self.term` is 1 but now it's `0`. So it panics at raft.rs line 463.

src/storage.rs
```diff
@@ -353,6 +359,33 @@ impl MemStorage {
...
+    pub fn initialize_with_conf_state<T>(&self, conf_state: T)
+    where
+        ConfState: From<T>,
+    {
+        assert!(!self.initial_state().unwrap().initialized());
+        trace!("create storage with given config");
+        let mut core = self.wl();
+        // Set index to 1 to make `first_index` greater than 1 so that there will be a gap between
+        // uninitialized followers and the leader. And then followers can catch up the initial
+        // configuration by snapshots.
```
BusyJay: Please explain why choose snapshot instead of logs.

tests/integration_cases/test_membership_changes.rs
```diff
@@ -143,16 +136,16 @@ mod three_peers_add_voter {
...
             &[2, 3],
-            2,
+            3,
             ConfChangeType::BeginMembershipChange,
         );
         scenario.assert_in_membership_change(&[1, 2, 3]);
 
         info!("Allowing new peers to catch up.");
-        scenario.expect_read_and_dispatch_messages_from(&[4, 1, 4])?;
+        scenario.expect_read_and_dispatch_messages_from(&[4, 1, 4, 1])?;
```
BusyJay: Why there is an additional 1?

hicqu: Previously in 2-phase membership changes new added peers' `next_idx` is 1 but now it's leaders' `last_idx`. So new added followers will reject the first `msg append` from their leader, so their leader needs to send more a `msg append`.

tests/integration_cases/test_raft_paper.rs
```diff
@@ -552,18 +553,22 @@ fn test_leader_commit_preceding_entries() {
...
+        vec![empty_entry(1, 2)],
     ];
 
     for (i, mut tt) in tests.drain(..).enumerate() {
-        let s = new_storage();
-        s.wl().append(&tt).expect("");
-        let mut r = new_test_raft(1, vec![1, 2, 3], 10, 1, s);
-        r.load_state(&hard_state(2, 0, 0));
+        let mut r = {
+            let mut r = new_test_raft(1, vec![1, 2, 3], 10, 1, new_storage());
```
BusyJay: Why not create the storage directly?

tests/integration_cases/test_raft.rs
```diff
@@ -848,11 +851,10 @@ fn test_dueling_candidates() {
         if nt.peers[&id].term != term {
             panic!("#{}: term = {}, want {}", i, nt.peers[&id].term, term);
         }
-        let base = ltoa(raft_log);
-        let l = ltoa(&nt.peers[&(1 + i as u64)].raft_log);
-        if base != l {
-            panic!("#{}: raft_log:\n {}, want:\n {}", i, l, base);
-        }
+
+        assert_eq!(nt.peers[&id].raft_log.committed, raft_logs[i].0);
```
BusyJay: Should keep the same style as L852.

tests/integration_cases/test_raft.rs
```diff
@@ -2267,15 +2206,19 @@ fn test_read_only_option_lease() {
...
+        (1, 10, 12, "ctx1"),
+        (2, 10, 22, "ctx2"),
+        (3, 10, 32, "ctx3"),
+        (1, 10, 42, "ctx4"),
+        (2, 10, 52, "ctx5"),
+        (3, 10, 62, "ctx6"),
     ];
 
     for (i, (id, proposals, wri, wctx)) in tests.drain(..).enumerate() {
+        assert_eq!(
```
BusyJay: Why add extra assertion?

tests/integration_cases/test_raft.rs
```diff
@@ -2494,20 +2437,21 @@ fn test_leader_append_response() {
...
+    sm.restore(s.clone());
+    sm.raft_log.store.wl().apply_snapshot(s).unwrap();
 
     sm.become_candidate();
     sm.become_leader();
     for i in 0..10 {
         sm.append_entry(&mut [empty_entry(0, i as u64 + 1)]);
     }
-    // slow follower
+
```
BusyJay: Why remove the comment?

tests/integration_cases/test_raft.rs
```diff
@@ -2646,6 +2591,8 @@ fn test_send_append_for_progress_probe() {
     r.become_leader();
     r.read_messages();
     r.mut_prs().get_mut(2).unwrap().become_probe();
+    // Because on index 1 there is a snapshot.
+    r.mut_prs().get_mut(2).unwrap().next_idx = 2;
```
BusyJay: Should not change `next_idx`.

PR #200, opened 2019-03-21
Add a new method for RaftState

nrc:
`RaftState` has private fields so this is required so that `RaftState` objects can be created by clients.

PTAL @Hoverbear @overvenus


nrc:
@Hoverbear can you force merge this please?

PR #201, opened 2019-03-30
Replace rust-protobuf with prost

ice1000:
This PR:

+ Introduce prost dependencies
+ Added `build.rs` to (re)generate prost structs and their corresponding protobuf wrappers (thanks to @nrc!)
 + The generated wrappers are manually edited a little bit, please don't run regenerate yourself ATM
+ **Passed all tests** by changing direct usages of protobuf APIs into prost APIs

I suppose test-passing means ready-for-review, please leave your comments!


ice1000:
Collecting some information from my local test output:

```
test integration_cases::test_raft::test_add_node ... ok
ERROR 2019-03-30T23:45:20Z: raft::raft: The node 1 already exists in the voters set.

test integration_cases::test_raft::test_step_config ... ok
ERROR 2019-03-30T23:45:20Z: raft::raft:  can't become learner when restores snapshot [index: 11, term: 11]

test integration_cases::test_raft_snap::test_snapshot_succeed ... ok
ERROR 2019-03-30T23:45:20Z: raft::raft: The node 1 already exists in the voters set.
ERROR 2019-03-30T23:45:20Z: raft::raft: The node 1 already exists in the voters set.

test integration_cases::test_raw_node::test_raw_node_start ... ok
ERROR 2019-03-30T23:45:20Z: raft::raft:  invalid format of MsgReadIndexResp from 0, entries count: 0
```

Am I supposed to see these errors?

Cargo.toml
```diff
@@ -11,16 +11,27 @@ documentation = "https://docs.rs/raft"
 description = "The rust language implementation of Raft algorithm."
 categories = ["algorithms", "database-implementations"]
 edition = "2018"
+build = "build.rs"
+
+[build-dependencies]
+protoc-rust = "~2.0-2.2"
+protoc-grpcio = "0.3.1"
```
siddontang: It is very strange to depend on a gRPC lib

nrc:
Could you squash (or even better organise) your commits please?

Cargo.toml
nrc: It shouldn't be necessary, using protobuf-build should be enough

build.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    for f in &file_names {
+        println!("cargo:rerun-if-changed={}", f);
+    }
+
+    // Delete the previously-generated wrapper, we're ok if the file does not exist
+    remove_file("rsprost/protobuf_compat.rs");
+
+    // Generate Prost files.
+    generate_prost_files(&file_names, "src/rsprost");
```
nrc: Could you change `rsprost` to just `prost`?

build.rs
```diff
@@ -0,0 +1,97 @@
...
+        text.push_str(".rs\");");
+        text.push_str("}\n\n");
+    }
+    text.push_str("pub mod protobuf_compat;\n");
+
+    let mut lib = File::create("src/rsprost.rs").expect("Could not create rsprost.rs");
+    lib.write_all(text.as_bytes())
+        .expect("Could not write rsprost.rs");
+
+    let protobuf_compat_text = "
```
nrc: I think it is nice not to use this, it should be easy enough to replace uses of `RepeatedField` with `Vec`

src/errors.rs
```diff
@@ -56,6 +57,20 @@ quick_error! {
             description(err.description())
             display("protobuf error {:?}", err)
         }
```
nrc: Can we remove this error?

src/lib.rs
```diff
@@ -365,22 +364,29 @@ before taking old, removed peers offline.
...
 #[cfg(feature = "failpoint")]
 #[macro_use]
 extern crate fail;
 
 #[macro_use]
 extern crate log;
+extern crate bytes;
+extern crate prost;
+extern crate prost_derive;
+extern crate protobuf;
```
nrc: We don't need these since we're using 2018 edition

src/log_unstable.rs
```diff
@@ -186,6 +186,7 @@ mod test {
     use crate::eraftpb::{Entry, Snapshot, SnapshotMetadata};
     use crate::log_unstable::Unstable;
     use harness::setup_for_test;
+    use protobuf::Message as Msg;
```
nrc: Where is this used?

src/rsprotobuf.rs
```diff
@@ -0,0 +1,3 @@
+pub use raft::eraftpb;
```
nrc: Can we get rid of all the protobuf stuff like this file?

src/util.rs
```diff
@@ -17,7 +17,7 @@
 use std::u64;
 
 use crate::eraftpb::{ConfChange, ConfChangeType, ConfState, Entry, Message};
-use protobuf;
+use protobuf::Message as Msg;
```
nrc: This `use` appears in a lot of places, I think it is not necesary in most places to alias it to `Msg`, and it is better not to do that if we can

build.rs
ice1000: I didn't do that because there's also an external crate called prost. Do you think it's a problem?

src/errors.rs
ice1000: Maybe later. There's a lot of `?;` codes raising protobuf errors.

nrc: Hmm, where do the protobuf errors come from if we're not using protobuf? Seems fine to leave it if that is the case though

ice1000:
> Could you squash (or even better organise) your commits please?

Can I do this after addressing all the comments?

src/util.rs
ice1000: This is necessary for using `protobuf`'s methods like `new()`.
Is it possible to drop this `use` (apart from completely migrating to prost)?

src/log_unstable.rs
ice1000: Like `new`s.

src/errors.rs
ice1000: > where do the protobuf errors come

Existing codes calling protobuf APIs (which are redirected to the wrapper). Once we've completely eliminate them we can remove `Codec`!

src/util.rs
nrc: I think we should not use `protobuf::Message::new`and instead use the inherent `new_` method from the wrapper. If we do need the `Message` trait for other things then it's fine to keep the import, but we should drop `as Msg`.

Cargo.toml
```diff
@@ -11,16 +11,26 @@ documentation = "https://docs.rs/raft"
 description = "The rust language implementation of Raft algorithm."
 categories = ["algorithms", "database-implementations"]
 edition = "2018"
+build = "build.rs"
+
+[build-dependencies]
+protoc-rust = "~2.0-2.2"
+regex = "1.1"
+protobuf-build = { git = "https://github.com/nrc/kvproto-build" }
```
nrc: I've published version 0.2.0 which should be used here. I don't think you should need the other build deps

Cargo.toml
```diff
@@ -11,16 +11,26 @@ documentation = "https://docs.rs/raft"
...
 
 [features]
 default = []
+regenerate = []
```
nrc: Could we use `gen` here? (And `prost-codec`/`protobuf-codec` for the codec libs). I'm trying to be consistent on these options across all the crates.

Cargo.toml
nrc: And now 0.3 (has some more implementations which are needed in TiKV)

ice1000:
The rest of the refactoring depends on kvproto-build#2

nrc:
> The rest of the refactoring depends on nrc/kvproto-build#2

For at least the proto which is exported via kvproto, the `Message` impl is used by TiKV, so I think we need to keep those Message impls around for now.

ice1000:
That sounds reasonable. I'll add the protobuf stuff back.

src/errors.rs
```diff
@@ -182,10 +189,6 @@ mod tests {
             Error::StepPeerNotFound,
             Error::Store(StorageError::Compacted)
         );
-        assert_ne!(
-            Error::Codec(ProtobufError::MessageNotInitialized { message: "" }),
-            Error::StepLocalMsg
-        );
```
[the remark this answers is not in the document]
ice1000: Due to the visibility of everything in `prost::*codeError`, I cannot construct such an error struct :(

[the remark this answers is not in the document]
nrc: OK

ice1000:
Test failing because

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

I have no idea about this.

nrc:
Travis is failing on Windows only, @Hoverbear can we ignore that?

Hoverbear:
@nrc We cannot!

PR #202, opened 2019-04-02
make configuration change effective on received

hicqu:
Make configuration changes effective on received.

Some test cases about membership change are removed:
* tests in test_membership_change::api.
  Because now these API are removed.
* pending_delete_fails_after_begin, pending_create_with_quorum_fails_after_begin, pending_create_and_destroy_both_fail.
  They are covered by minority_followers_halt.


Hoverbear:
@hicqu Can you make the CI pass?

hicqu:
PTAL @BusyJay @Hoverbear @nolouch thank you!

src/progress/progress_set.rs
```diff
@@ -77,17 +81,17 @@ where
     }
 }
 
-impl From<ConfState> for Configuration {
-    fn from(conf_state: ConfState) -> Self {
+impl From<&ConfState> for Configuration {
+    fn from(conf_state: &ConfState) -> Self {
```
Hoverbear: It's generally not idiomatic to implicitly clone, better to use the `ConfState` and let the caller clone.

src/progress/progress_set.rs
```diff
@@ -652,27 +663,62 @@ impl ProgressSet {
...
+            if self.get(id).is_none() {
+                self.progress.insert(id, progress.clone());
+            }
+        }
+        self.cut_progress();
+    }
+
+    #[inline]
+    fn set_recent_active(&mut self, id: u64) {
+        // When a node is first added/promoted, we should mark it as recently active.
```
Hoverbear: Maybe this is a doc (`///`) function comment?

src/raft.rs
```diff
@@ -152,21 +153,9 @@ pub struct Raft<T: Storage> {
...
-    /// one-by-one change method, in joint consensus based changes we track the state exactly. The
-    /// index here **must** only be set when a `BeginMembershipChange` is present at that index.
-    ///
-    /// # Caveats
-    ///
-    /// It is important that whenever this is set that `pending_conf_index` is also set to the
-    /// value if it is greater than the existing value.
-    ///
-    /// **Use `Raft::set_pending_membership_change()` to change this value.**
+    /// History configuration states. A memory copy of `RaftState::conf_states` in `Storage`.
```
```suggestion
    /// Configuration states history and future. A memory copy of `RaftState::conf_states` in `Storage`.
```

src/raft.rs
```diff
@@ -742,22 +713,116 @@ impl<T: Storage> Raft<T> {
         }
     }
 
+    // recover configuration states to the latest one before `index`.
```
```suggestion
    // recover configuration states until the latest one before `index`.
```

src/raft.rs
```diff
@@ -709,9 +670,19 @@ impl<T: Storage> Raft<T> {
...
+            panic!("{} append finalize entry shouldn't fail", self.tag);
+        }
+
         self.bcast_append();
+        if !self.prs().iter().any(|(id, _)| *id == self.id) {
+            self.become_follower(self.term, INVALID_ID);
+            debug!("{} append finalize entry and become follower", self.tag);
+            return;
+        }
+        debug!("{} append finalize entry", self.tag);
```
Hoverbear: Avoid the repeat log

```suggestion
        else { debug!("{} append finalize entry", self.tag); }
```

src/raft.rs
```diff
@@ -742,22 +713,116 @@ impl<T: Storage> Raft<T> {
         }
     }
 
+    // recover configuration states to the latest one before `index`.
+    fn recover_conf_states_ahead(&mut self, index: u64) {
```
Hoverbear: Perhaps we could add some `debug!()` here? (Eg "Revert needed, reverting to conf state {:?}")

src/raft.rs
```diff
@@ -900,25 +965,22 @@ impl<T: Storage> Raft<T> {
         // could be expensive.
         self.pending_conf_index = self.raft_log.last_index();
 
-        self.append_entry(&mut [Entry::new()]);
+        // Just unwrap is OK here.
```
Hoverbear: Why?

src/raft.rs
```diff
@@ -1211,101 +1273,21 @@ impl<T: Storage> Raft<T> {
...
-        if conf_change.get_start_index() == 0 {
-            return Err(Error::ViolatesContract(
-                "!ConfChange::has_start_index()".into(),
-            ));
-        };
-
-        self.set_pending_membership_change(conf_change.clone());
-        let pr = Progress::new(self.raft_log.last_index() + 1, self.max_inflight);
-        self.mut_prs().begin_membership_change(configuration, pr)?;
+    fn begin_membership_change(&mut self, cs: &ConfState) -> Result<()> {
```
Hoverbear: It could use some documentation for the reader!

src/raft.rs
```diff
@@ -1211,101 +1273,21 @@ impl<T: Storage> Raft<T> {
...
-            let last_term = self.raft_log.last_term();
-            if self.state == StateRole::Leader {
-                self.become_follower(last_term, INVALID_ID);
-            } else {
-                // It's no longer safe to lookup the ID in the ProgressSet, remove it.
-                self.leader_id = INVALID_ID;
-            }
-        }
-
+    fn finalize_membership_change(&mut self) -> Result<()> {
```
Hoverbear: Similar need of documentation for the reader!

src/raft.rs
```diff
@@ -1211,101 +1273,21 @@ impl<T: Storage> Raft<T> {
...
-        };
-
-        self.set_pending_membership_change(conf_change.clone());
-        let pr = Progress::new(self.raft_log.last_index() + 1, self.max_inflight);
-        self.mut_prs().begin_membership_change(configuration, pr)?;
+    fn begin_membership_change(&mut self, cs: &ConfState) -> Result<()> {
+        debug!("{} begin membership change to {:?}", self.tag, cs);
+        let progress = Progress::new(1, self.max_inflight);
+        self.mut_prs().begin_membership_change(cs, progress)?;
+        self.update_learner_flag();
```
Hoverbear: Eventually maybe we can get rid of `Raft.is_learner` all together. 🤔

src/progress/progress_set.rs
hicqu: I guess `cloned` won't cause real clone because `u64` is `Copy`. But I would like to use `.map(|n| *n)` to replace `.cloned()` to make it more clearly.

src/storage.rs
```diff
@@ -37,44 +37,36 @@ use crate::eraftpb::*;
...
 use crate::util::limit_size;
 
+/// Used to track the history about configuration changes.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct ConfStateWithIndex {
+    /// Target configuration state.
+    pub conf_state: ConfState,
+    /// Index of the entry.
+    pub index: u64,
+    /// Indicates it's a membership change or not.
```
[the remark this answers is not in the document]
hicqu: No. All configuration changes including AddNode, AddLearner and RemoveNode use the struct.

src/progress/progress_set.rs
hicqu: Clippy suggests `cloned` instead of `map(|n| *n)` so I revert it. However I have checked the implementation of `Iterator::cloned`, which makes me sure about there won't be any `clone` here. So could I keep it?

harness/src/network.rs
```diff
@@ -166,6 +170,12 @@ impl Network {
         }
     }
 
+    /// Filter msgs and then call `Network::send(&mut self, filtered)`.
+    pub fn filter_and_send(&mut self, mut msgs: Vec<Message>) {
```
BusyJay: How about calling `read_messages` inside the function too?

src/progress/progress_set.rs
```diff
@@ -77,17 +81,17 @@ where
...
         Self {
             voters: conf_state.get_nodes().iter().cloned().collect(),
             learners: conf_state.get_learners().iter().cloned().collect(),
         }
     }
 }
 
-impl From<Configuration> for ConfState {
-    fn from(conf: Configuration) -> Self {
+impl From<&Configuration> for ConfState {
```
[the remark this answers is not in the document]
Hoverbear: In this case `as_conf_state` would more correctly communicate that the structure is not being consumed.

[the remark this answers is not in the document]
BusyJay: I think `into_xxx` means self is consumed, but `to_xxx`.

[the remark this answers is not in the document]
hicqu: `as_conf_state` seems should return `&ConfState` instead of `ConfState`. So personally I prefer `to_conf_state`.

hicqu:
PTAL again, thank you. @Hoverbear @BusyJay

Hoverbear:
@hicqu I'm sorry about the merge conflicts, we merged https://github.com/pingcap/raft-rs/pull/201. :(

Could you fix the merge conflicts? I think in most cases it should be straightforward. You can ask @nrc or @ice1000 if you encounter any strange problems.

ice1000:
+ Replace `::new()` with `::new_()`
+ Replace `use protobuf::Message` with `use prost::Message as Msg`
+ Replace `x.compute_size()` with `Msg::encoded_len(&x)` (where this `Msg` is `prost::Message`)
+ Replace `RepeatedField::from_vec(x)` with `x`
+ Replace `let data = protobuf::Message::write_to_bytes(&cc)?;` with `let mut data = Vec::with_capacity(ProstMsg::encoded_len(&cc)); cc.encode(&mut data)?;`

hicqu:
@Hoverbear it doesn't matter. I was waiting for #214 exactly. Could you take a look again? Thanks!

Hoverbear:
@hicqu You have a clippy lint fail :(

```
The command "if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo fmt -- --check; fi" exited with 0.
9.10s$ if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo clippy -- -D clippy::all; fi
    Updating crates.io index
   Compiling raft v0.5.0 (/home/travis/build/pingcap/raft-rs)
error: redundant closure found
   --> src/progress/progress_set.rs:673:49
    |
673 |         self.next_configuration = next_conf.map(|c| c.into());
    |                                                 ^^^^^^^^^^^^ help: remove closure as shown: `std::convert::Into::into`
    |
note: lint level defined here
   --> src/lib.rs:360:9
    |
360 | #![deny(clippy::all)]
    |         ^^^^^^^^^^^
    = note: #[deny(clippy::redundant_closure)] implied by #[deny(clippy::all)]
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
error: aborting due to previous error
error: Could not compile `raft`.
To learn more, run the command again with --verbose.
The command "if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo clippy -- -D clippy::all; fi" exited with 101.
```

harness/src/network.rs
```diff
@@ -172,6 +178,12 @@ impl Network {
         }
     }
 
+    /// Call `self.read_messages()` and then send them.
```
Hoverbear: Seems this doesn't filter either?

src/prost/wrapper_eraftpb.rs
```diff
@@ -163,40 +163,49 @@ impl SnapshotMetadata {
...
+        self.conf_state_index = v;
     }
-    pub fn set_pending_membership_change(&mut self, v: ConfState) {
-        self.pending_membership_change = ::std::option::Option::Some(v);
+    pub fn get_conf_state_index(&self) -> u64 {
+        self.conf_state_index
     }
-    pub fn get_pending_membership_change(&self) -> &ConfState {
-        match self.pending_membership_change.as_ref() {
+    pub fn has_next_conf_state(&self) -> bool {
```
Hoverbear: If `get_next_conf_state`  returned an option then this would be unnecessary? 🤔 It seems a bit awkward. It's not a problem though, but I think it's confusing.

src/raft.rs
```diff
@@ -116,12 +117,6 @@ pub struct Raft<T: Storage> {
     /// The current role of this node.
     pub state: StateRole,
 
-    /// Whether this is a learner node.
-    ///
-    /// Learners are not permitted to vote in elections, and are not counted for commit quorums.
-    /// They do replicate data from the leader.
-    pub is_learner: bool,
```
Hoverbear: ❤️

tests/integration_cases/test_membership_changes.rs
```diff
@@ -471,8 +297,8 @@ mod remove_leader {
         scenario.send(messages);
 
         let peer_leader = scenario.peer_leaders();
-        assert_ne!(peer_leader[&2], 1);
-        assert_ne!(peer_leader[&3], 1);
+        assert!(peer_leader[&2] != 1);
```
Hoverbear: Why this change? It gives a better error message the other way.

harness/src/network.rs
hicqu: I also changed `read_message`, now it does filter.

src/prost/wrapper_eraftpb.rs
hicqu: It's generated by prost. I guess it follows CPP's style. I also think it's weird and it's easy to make mistakes. Could we change the codegen to return an `Option`? @ice1000

ice1000: Correction: it's not generated by prost but by our wrapper generator. Prost structs' fields are public, there's no need of getters/setters/hases. These methods are generated to ease the migration from rust-protobuf to prost.

ice1000: There must be a similar method that rust-protobuf will generate, too.
This can be an issue to github.com/nrc/kvproto-build

Hoverbear: Let's leave it for now.

Hoverbear:
@qupeng I remember you said you wanted to add another test here that @busyjay suggested?

Hoverbear:
@hicqu Please address the merge conflict :)

src/progress/progress_set.rs
```diff
@@ -79,24 +91,15 @@ where
     }
 }
 
-impl From<ConfState> for Configuration {
-    fn from(conf_state: ConfState) -> Self {
+impl From<&ConfState> for Configuration {
```
BusyJay: This looks really strange. How about add a `from_conf_state` method to `Configuration`?

src/progress/progress_set.rs
```diff
@@ -185,13 +190,34 @@ impl ProgressSet {
...
             ),
             sort_buffer: RefCell::from(Vec::with_capacity(voters)),
-            configuration_capacity: (voters, learners),
             configuration: Configuration::with_capacity(voters, learners),
             next_configuration: Option::default(),
             logger: default_logger().new(o!()),
         }
     }
 
+    pub(crate) fn restore_conf_states(
```
BusyJay: Strange method name. I would expect it mutate self instead of returning a new instance.

src/progress/progress_set.rs
```diff
@@ -203,36 +229,29 @@ impl ProgressSet {
         meta: &SnapshotMetadata,
         next_idx: u64,
         max_inflight: usize,
-        logger: &Logger,
```
BusyJay: Why remove logger?

hicqu:
PTAL @BusyJay thanks!

src/progress/progress_set.rs
```diff
@@ -64,6 +68,22 @@ impl Configuration {
             learners: learners.into_iter().collect(),
         }
     }
+
+    /// Create a new `ConfState` from the configuration itself.
+    pub fn to_conf_state(&self) -> ConfState {
```
Hoverbear: I take it there were some problems with the `From` implementations?

src/progress/progress_set.rs
```diff
@@ -195,51 +199,72 @@ impl ProgressSet {
...
+        next_idx: u64,
+        max_inflight: usize,
+    ) {
+        if conf_states.is_empty() {
+            return;
+        }
+        let mut meta = SnapshotMetadata::default();
+        let len = conf_states.len();
+        if conf_states[len - 1].in_membership_change {
+            meta.set_conf_state(conf_states[len - 2].conf_state.clone());
```
Hoverbear: I worry about slices of length `1`, do you think it's possible?

hicqu: By design, if the peer is in membeship change, the `conf_states` must have 2 elements at least. Otherwise it must be a bug.

src/progress/progress_set.rs
hicqu: Generally `From` needs to move the `Configuration`. Previously I write `impl From<&Configuration> for ConfState`, but Jay prefers `to_conf_state`.

hicqu:
PTAL again @BusyJay thanks!

src/progress/progress_set.rs
Hoverbear: Yes I do too, that would create an implicit copy.

src/raft_log.rs
```diff
@@ -483,26 +490,23 @@ impl<T: Storage> RaftLog<T> {
...
                     Error::Store(StorageError::Compacted) => return Err(e),
                     Error::Store(StorageError::Unavailable) => panic!(
                         "{} entries[{}:{}] is unavailable from storage",
-                        self.tag,
-                        low,
-                        cmp::min(high, self.unstable.offset)
+                        self.tag, low, unstable_high,
                     ),
                     _ => panic!("{} unexpected error: {:?}", self.tag, e),
+                },
```
BusyJay: Can you send a separate PR with such identical changes? I think they can be merged more quickly and make this PR more focus.

src/raft.rs
```diff
@@ -2125,31 +2133,35 @@ impl<T: Storage> Raft<T> {
...
-                    from = m.from;
-                    "index" => m.index,
-                    "tag" => &self.tag,
-                    "logterm" => ?self.raft_log.term(m.index),
-                );
-                to_send.index = m.index;
-                to_send.reject = true;
-                to_send.reject_hint = self.raft_log.last_index();
-                self.send(to_send);
+            if !append_ents.is_empty() {
```
BusyJay: Why additional recovery is necessary now while unnecessary in the past?

proto/proto/eraftpb.proto
```diff
@@ -29,10 +29,19 @@ message Entry {
 }
 
 message SnapshotMetadata {
+    // The current `ConfState`.
     ConfState conf_state = 1;
-    ConfState pending_membership_change = 4;
-    uint64 pending_membership_change_index = 5;
+    // The index of the current `ConfState`.
+    uint64 conf_state_index = 5;
```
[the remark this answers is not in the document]
hicqu: Only the field name is changed, the number is keeped. It's compatible with old version.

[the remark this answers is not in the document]
BusyJay: I think the missing is changed too. It was an index related to pending membership change, but now it refers to the current one.

src/raft.rs
hicqu: In this PR configration changes become active on received, so some uncommitted configuration changes could be overwritten. So here we need to do such a recover. However in the past configuration changes become active on committed, so it's unnecessary.

proto/proto/eraftpb.proto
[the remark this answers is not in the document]
hicqu: Yes, it's a break change. However I think we can't do better.

src/raft_log.rs
```diff
@@ -188,10 +193,19 @@ impl<T: Storage> RaftLog<T> {
                         "conflicting term" => e.term,
                     );
                 }
-                return e.index;
+                conflict_idx = e.index;
+                break;
             }
         }
-        0
+        if conflict_idx > 0 && conflict_idx <= self.committed {
```
BusyJay: It's redundant.

src/progress/progress_set.rs
```diff
@@ -63,6 +67,22 @@ impl Configuration {
             learners: learners.into_iter().collect(),
         }
     }
+
+    /// Create a new `ConfState` from the configuration itself.
+    pub fn to_conf_state(&self) -> ConfState {
+        let mut state = ConfState::default();
+        state.set_nodes(self.voters.iter().cloned().collect());
```
BusyJay: I suggest you to use `vec` instead of `HashSet`. Benches shows that iterating vec is way faster than querying HashSet when the size of vec is very small.

```
running 2 tests
test bench_set_contains ... bench:          73 ns/iter (+/- 0)
test bench_vec_contains ... bench:          11 ns/iter (+/- 0)
```

hicqu: Let's do it after the PR.

src/progress/progress_set.rs
```diff
@@ -650,6 +674,29 @@ impl ProgressSet {
...
+        next_conf: Option<Configuration>,
+        progress: Progress,
+    ) {
+        self.progress.retain(|id, _| {
+            conf.contains(*id) || next_conf.as_ref().map_or(false, |c| c.contains(*id))
+        });
+
+        self.configuration = conf;
+        self.next_configuration = next_conf;
+        for id in self.voter_ids().into_iter().chain(self.learner_ids()) {
```
BusyJay: Is it necessary?

src/raft.rs
```diff
@@ -1716,24 +1720,35 @@ impl<T: Storage> Raft<T> {
...
+                            self.logger,
+                            "propose conf entry ignored since the previous one is not appended";
+                            "entry" => ?e, "previous" => first_conf_index,
+                        );
+                    }
+                    *e = Entry::default();
+                    e.set_entry_type(EntryType::EntryNormal);
                 }
-                self.append_entry(&mut m.mut_entries());
+                self.append_entry(&mut m.mut_entries())?;
```
BusyJay: Why not just panic?

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -234,48 +234,42 @@ fn test_raw_node_propose_add_duplicate_node() {
...
     let mut propose_conf_change_and_apply = |cc| {
-        raw_node.propose_conf_change(vec![], cc).expect("");
-        let rd = raw_node.ready();
-        s.wl().append(rd.entries()).expect("");
-        for e in rd.committed_entries.as_ref().unwrap() {
-            if e.get_entry_type() == EntryType::EntryConfChange {
-                let mut conf_change = ConfChange::default();
-                conf_change.merge_from_bytes(&e.data).unwrap();
-                raw_node.apply_conf_change(&conf_change).ok();
+        if raw_node.propose_conf_change(vec![], cc).is_ok() {
```
BusyJay: Why not unwrap directly?

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -300,13 +294,9 @@ fn test_raw_node_propose_add_learner_node() -> Result<()> {
...
     );
 
-    let e = &rd.committed_entries.as_ref().unwrap()[0];
-    let mut conf_change = ConfChange::default();
-    conf_change.merge_from_bytes(&e.data).unwrap();
-    let conf_state = raw_node.apply_conf_change(&conf_change)?;
-    assert_eq!(conf_state.nodes, vec![1]);
-    assert_eq!(conf_state.learners, vec![2]);
-
+    assert!(!rd.conf_states().is_empty());
```
BusyJay: Data should be checked too.

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -491,5 +481,5 @@ fn test_skip_bcast_commit() {
 
     assert_eq!(nt.peers[&1].raft_log.committed, 7);
     assert_eq!(nt.peers[&2].raft_log.committed, 7);
-    assert_eq!(nt.peers[&3].raft_log.committed, 7);
+    assert_eq!(nt.peers[&3].raft_log.committed, 5);
```
BusyJay: This shows a problem that a node will not get notified that it's removed.

BusyJay:
Bugs described in https://groups.google.com/forum/#!msg/raft-dev/t4xj6dJTP6E/d2D9LrWRza8J is not addressed in this PR.

src/progress/progress_set.rs
```diff
@@ -193,6 +194,29 @@ impl ProgressSet {
...
+        conf_states: &[ConfStateWithIndex],
+        next_idx: u64,
+        max_inflight: usize,
+    ) {
+        if conf_states.is_empty() {
+            return;
+        }
+        let mut meta = SnapshotMetadata::default();
+        let len = conf_states.len();
+        if conf_states[len - 1].in_membership_change {
```
BusyJay: This is not verbose. why not just `meta.set_conf_state(conf_states[0])`. And set next if there is additional values.

src/progress/progress_set.rs
```diff
@@ -439,23 +466,15 @@ impl ProgressSet {
...
-                    .map_or(false, |c| c.voters.contains(v))));
-        assert_eq!(
-            self.voter_ids().len() + self.learner_ids().len(),
-            self.progress.len()
-        );
+        debug_assert!(self.next_configuration.as_ref().map_or(true, |c| c
+            .voters
+            .union(&c.learners)
+            .all(|v| self.progress.contains_key(v))));
+        debug_assert!(self.next_configuration.as_ref().map_or(true, |c| c
```
BusyJay: Should it also be checked for the current configuration?

src/progress/progress_set.rs
```diff
@@ -631,24 +650,37 @@ impl ProgressSet {
...
+                "Finalizing membership change";
+                "config" => ?self.configuration,
+            );
+            return Ok(());
+        }
+        Err(Error::NoPendingMembershipChange)
+    }
+
+    pub(crate) fn revert(&mut self, conf: Configuration, next_conf: Option<Configuration>) {
+        self.progress.retain(|id, _| {
```
BusyJay: Why not leave it to gc?

src/progress/progress_set.rs
```diff
@@ -401,12 +425,15 @@ impl ProgressSet {
...
-        Ok(removed)
+        if self.configuration.voters.remove(&id) {
+            self.assert_progress_and_configuration_consistent();
+            return Ok(());
+        }
+        if self.configuration.learners.remove(&id) {
+            self.assert_progress_and_configuration_consistent();
+            return Ok(());
+        }
+        Err(Error::NotExists(id, "configuration"))
```
BusyJay: Is it safe to return Error? `Ok` is returned in the pass.

hicqu: It depends on how to deal with the `Error` outside. It's ignored in the caller, so it's safe to return `Error` here. Handle `Error` outside is just for unifying logs.

src/progress/progress_set.rs
```diff
@@ -776,14 +809,12 @@ mod test_progress_set {
...
-        assert!(check_membership_change_configuration((vec![], vec![]), (vec![], vec![])).is_err())
+        assert!(check_membership_change_configuration((vec![], vec![]), (vec![], vec![])).is_ok())
     }
 
     #[test]
     fn test_membership_change_configuration_empty_voters() {
-        assert!(
-            check_membership_change_configuration((vec![1], vec![]), (vec![], vec![]),).is_err()
-        )
+        assert!(check_membership_change_configuration((vec![1], vec![]), (vec![], vec![]),).is_ok())
```
BusyJay: /cc @Hoverbear PTAL

hicqu: It's because currently an empty configuration is allowed as target. Do you think we need to ban it? @Hoverbear

src/storage.rs
```diff
@@ -492,7 +479,9 @@ impl Storage for MemStorage {
         } else {
             let mut snap = core.snapshot();
             if snap.get_metadata().index < request_index {
+                // NOTE: the logic is only for tests.
                 snap.mut_metadata().index = request_index;
+                snap.mut_metadata().conf_state_index = request_index;
```
BusyJay: This is confusing. What if there are conf change between `conf_state_index` and `request_index`?

src/raw_node.rs
```diff
@@ -130,6 +132,18 @@ impl Ready {
             entries: raft.raft_log.unstable_entries().unwrap_or(&[]).to_vec(),
             ..Default::default()
         };
+        if let Some(e) = rd.entries.first() {
+            for i in (0..raft.conf_states().len()).rev() {
```
BusyJay: Why not use `Iterator::find`? It seems more understandable.

src/raw_node.rs
```diff
@@ -130,6 +132,18 @@ impl Ready {
...
+                if raft.conf_states()[i].index >= e.index {
+                    continue;
+                }
+                if i < raft.conf_states().len() {
```
BusyJay: I think it's always true.

src/progress/progress_set.rs
Hoverbear: I'm not really sure it's a valid state.

hicqu: I guess it's valid means there is a machanism to remove all peers from a Raft group, although TiKV doesn't need it.

src/raw_node.rs
hicqu: Seems it's better to find it from tail to head because there could be many unapplied entries in `conf_states`.

BusyJay: You mean `entries.iter().rev().position()`?

hicqu:
Close. Need to find other solutions for https://github.com/etcd-io/etcd/issues/11284.

PR #207, opened 2019-04-06
problem: `README.md` is outdated

652h:
solution: `README.md` was updated according to the introduced by #203 Rust 2018 Edition

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,20 +31,20 @@ A complete Raft model contains 4 essential parts:
 
 ## Developing the Raft crate
 
-`raft` is intended to track the latest `stable`, though you'll need to use `nightly` to simulate a full CI build with `clippy`.
+`raft` is intended to track [Rust 2018 Edition](https://doc.rust-lang.org/edition-guide/rust-2018/index.html).
 
 Using `rustup` you can get started this way:
 
 ```bash
-rustup component add clippy-preview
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,20 +31,20 @@ A complete Raft model contains 4 essential parts:
 
 ## Developing the Raft crate
 
-`raft` is intended to track the latest `stable`, though you'll need to use `nightly` to simulate a full CI build with `clippy`.
+`raft` is intended to track [Rust 2018 Edition](https://doc.rust-lang.org/edition-guide/rust-2018/index.html).
```
nrc: This phrasing could be a little clearer - the 2018 is not a channel which can be tracked but a kind of mega-feature gate. Perhaps: "Raft is built using the latest version of stable Rust, using the 2018 edition."

652h: Tnks, and if to be perfectly precise, code is be checked for compatibility with lower Rust 2018 compiler version `1.31.0` (https://github.com/pingcap/raft-rs/blob/master/.travis.yml#L11).

@nrc that about "`raft` is intended to track the latest `stable`, version `1.31.0` or higher ([Rust 2018 Edition](https://doc.rust-lang.org/edition-guide/rust-2018/index.html))."?

652h: But anyway, I found your variant is probably better, thank you.

PR #213, opened 2019-04-10
fix a bug about snapshot and adjust some test cases

hicqu:
The bug causes an exmaple panics in #205 .

In our implemetation, `RaftLog::term(un_received_index)` returns `0`. So if a peer receives a snapshot or normal entry with log-term `0`, `RaftLog::match_term(log_term, un_received_index)` will return `true` but it should be `false` exactly.

This PR fixes it by making `RaftLog::match_term` cover this case.

And, add a case `test_snapshot_with_term_0` to cover it, and do some adjusts for some other test cases. At last, remove `Interface::initial` because we don't need it any more.

### DIFF
--- a/examples/five_mem_node/main.rs
+++ b/examples/five_mem_node/main.rs
@@ -103,6 +103,7 @@ fn main() {
     add_all_followers(proposals.as_ref());
 
     // Put 100 key-value pairs.
+    println!("We get a 5 nodes Raft cluster now, now propose 100 proposals");
     (0..100u16)
         .filter(|i| {
             let (proposal, rx) = Proposal::normal(*i, "hello, world".to_owned());
@@ -113,6 +114,9 @@ fn main() {
         })
         .count();
 
+    println!("Propose 100 proposals success!");
+
+    // FIXME: the program will be blocked here forever. Need to exit gracefully.
     for th in handles {
         th.join().unwrap();
     }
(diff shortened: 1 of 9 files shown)


hicqu:
PTAL @Hoverbear @BusyJay thank you!

examples/five_mem_node/main.rs
```diff
@@ -207,6 +211,15 @@ fn on_ready(
         return;
     }
 
+    // Apply the snashot. It's also necessary with same reason as above.
```
Hoverbear:
```suggestion
    // Apply the snapshot. It's also necessary with same reason as above.
```

Also, which reason? That the program will be blocked forever?

harness/src/network.rs
```diff
@@ -61,21 +61,39 @@ pub struct Network {
 }
 
 impl Network {
-    /// Initializes a network from peers.
+    /// Get a base config. Calling `Network::new` will initialize peers with this config.
+    pub fn base_config() -> Config {
```
Hoverbear:
```suggestion
    pub fn default_test_config() -> Config {
```

Hoverbear:
@hicqu Seems maybe `RaftLog::term(un_received_index)` should return `None` instead? Would that be a good way to fix this problem?

hicqu:
Thanks for your review @Hoverbear ! I will address them in #214 . Now I will close this PR because I prefer the another solution. :)

bytewatch:
I think cause the root cause of the panic of five_mem_node is MemStorage::initialize_with_conf_state() :

```
        // To fix this panic,  add a more line: `core.raft_state.hard_state.set_commit(1)` , 
        // and handle the case that `prs.get_mut(self.id)` is None in restore_snapshot.
        // Or , just commet these two lines: 
        core.snapshot_metadata.set_index(1);  
        core.raft_state.hard_state.set_commit(1);  
        core.raft_state.conf_state = ConfState::from(conf_state);

```

src/raft_log.rs
```diff
@@ -199,7 +199,12 @@ impl<T: Storage> RaftLog<T> {
 
     /// Answers the question: Does this index belong to this term?
     pub fn match_term(&self, idx: u64, term: u64) -> bool {
-        self.term(idx).map(|t| t == term).unwrap_or(false)
+        match self.term(idx) {
+            // For uninitialized storage, should return false.
+            Ok(0) => term == 0 && self.last_index() > 0,
```
bytewatch: I don't think it need to add some logics that exceeds etcd-raft.

PR #214, opened 2019-04-10
fix a bug about initialized storage's term

hicqu:
Try to resolve same problem with #213 , but with a different way.

This PR changes initialized storage's term from `0` to `1`. So all place we need to call `match_term` won't get `0` as parameter, so the problem is resolved.

However more test cases need to be updated.
Please compare this PR with #213 , we only need to merge 1.


Hoverbear:
@hicqu Which solution do you think is most elegant and simple to understand?

harness/src/network.rs
```diff
@@ -61,21 +61,39 @@ pub struct Network {
 }
 
 impl Network {
-    /// Initializes a network from peers.
+    /// Get a base config. Calling `Network::new` will initialize peers with this config.
+    pub fn base_config() -> Config {
```
Hoverbear:
```suggestion
    pub fn default_test_config() -> Config {
```

src/raft.rs
```diff
@@ -2040,7 +2044,10 @@ impl<T: Storage> Raft<T> {
 
         let next_idx = self.raft_log.last_index() + 1;
         let mut prs = ProgressSet::restore_snapmeta(meta, next_idx, self.max_inflight);
-        prs.get_mut(self.id).unwrap().matched = next_idx - 1;
+        if let Some(pr) = prs.get_mut(self.id) {
+            // The snapshot could be too old so that it doesn't contain the peer.
```
BusyJay: Why? How can a leader send a snapshot to a node that it doesn't know?

harness/src/network.rs
```diff
@@ -61,21 +61,39 @@ pub struct Network {
...
-    /// Explicitly set the pre_vote option on newly created rafts.
-    ///
-    /// **TODO:** Make this accept any config.
-    pub fn new_with_config(mut peers: Vec<Option<Interface>>, pre_vote: bool) -> Network {
-        let size = peers.len();
-        let peer_addrs: Vec<u64> = (1..=size as u64).collect();
+    /// Initialize a network from `peers` with explicitly specified `config`.
+    pub fn new_with_config(mut peers: Vec<Option<Interface>>, config: &Config) -> Network {
+        let peer_addrs: Vec<u64> = (1..=peers.len() as u64).collect();
+        for (i, id) in peer_addrs.iter().enumerate() {
```
BusyJay: Why not use `zip`?

harness/src/network.rs
```diff
@@ -84,23 +102,13 @@ impl Network {
...
-                        ..Default::default()
-                    };
+                    let mut config = config.clone();
+                    config.id = id;
+                    config.tag = format!("{}", id);
                     let r = Raft::new(&config, store).unwrap().into();
                     npeers.insert(id, r);
                 }
-                Some(mut p) => {
-                    p.initial(id, &peer_addrs);
```
BusyJay: Why initial is unnecessary?

tests/integration_cases/test_raft_snap.rs
```diff
@@ -134,3 +134,22 @@ fn test_snapshot_abort() {
...
+// Initialized storage should be at term 1 instead of 0. Otherwise the case will fail.
+#[test]
+fn test_snapshot_with_min_term() {
+    setup_for_test();
+    let do_test = |pre_vote: bool| {
+        let n1 = new_test_raft_with_prevote(1, vec![1, 2], 10, 1, new_storage(), pre_vote);
+        let n2 = new_test_raft_with_prevote(2, vec![], 10, 1, new_storage(), pre_vote);
+        let mut nt = Network::new(vec![Some(n1), Some(n2)]);
+        nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
+        let messages = nt.read_messages();
```
BusyJay: I don't think there will be any messages.

tests/integration_cases/test_raft.rs
```diff
@@ -1418,35 +1414,36 @@ fn test_recv_msg_request_vote() {
...
         (StateRole::Follower, 1, 2, INVALID_ID, true),
         (StateRole::Follower, 1, 3, INVALID_ID, false),
-        (StateRole::Follower, 2, 0, INVALID_ID, true),
         (StateRole::Follower, 2, 1, INVALID_ID, true),
-        (StateRole::Follower, 2, 2, INVALID_ID, false),
+        (StateRole::Follower, 2, 1, INVALID_ID, true),
+        (StateRole::Follower, 2, 2, INVALID_ID, true),
         (StateRole::Follower, 2, 3, INVALID_ID, false),
-        (StateRole::Follower, 3, 0, INVALID_ID, true),
+        (StateRole::Follower, 3, 1, INVALID_ID, true),
```
BusyJay: What's the difference between L1425 and L1426?

tests/integration_cases/test_raft.rs
```diff
@@ -1418,35 +1414,36 @@ fn test_recv_msg_request_vote() {
...
         (StateRole::Follower, 3, 1, INVALID_ID, true),
         (StateRole::Follower, 3, 2, INVALID_ID, false),
         (StateRole::Follower, 3, 3, INVALID_ID, false),
-        (StateRole::Follower, 3, 2, 2, false),
-        (StateRole::Follower, 3, 2, 1, true),
-        (StateRole::Leader, 3, 3, 1, true),
-        (StateRole::PreCandidate, 3, 3, 1, true),
-        (StateRole::Candidate, 3, 3, 1, true),
+        (StateRole::Follower, 4, 1, INVALID_ID, true),
```
BusyJay: L1429 and L1430 seem the same.

harness/src/network.rs
hicqu: Because all need tests which need `initial` is fixed. I think `initial` is a not a good logic because it does additional changes to the peers, which should be already set correctly when the peer is created by `Raft::new`.

hicqu:
@Hoverbear I think this PR is better because we won't meet `term ==0` anymore. Messages with term 0 is treated as local messages.

Hoverbear:
@hicqu Ok, do you want to close the other then?

PR #219, opened 2019-04-18
Address some warnings

nrc:
@BusyJay PTAL (causing CI failure on master).


hicqu:
LGTM.

nrc:
Added another fix, PTAL @hicqu and @BusyJay

PR #227, opened 2019-04-29
Add more convenient lite-weight interfaces

BusyJay:
This PR introduces two simple and lite weight interfaces:

- ping to trigger heartbeats without ticking,
- status_ref to borrow the progress set instead of cloning.


hicqu:
LGTM.

PR #230, opened 2019-05-07
*: tiny refactor

BusyJay:
Make status borrow data instead of cloning. Also introduce a new API to
force leader broadcast heartbeats.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -651,7 +651,9 @@ impl<T: Storage> Raft<T> {
         self.set_prs(prs);
     }
 
-    /// Broadcasts heartbeats to all the followers if it's leader.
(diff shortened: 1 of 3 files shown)


hicqu:
LGTM

src/raft.rs
```diff
@@ -651,6 +651,13 @@ impl<T: Storage> Raft<T> {
         self.set_prs(prs);
     }
 
+    /// Broadcasts heartbeats to all the followers if it's leader.
```
Hoverbear: Seems this and the other ping in raw_node could share the same comment.

Hoverbear:
#232 should fix this failed CI, merging.

PR #233, opened 2019-05-08
Ease protobuf dependency

Hoverbear:
I realized when trying to use Raft 0.5.0 in kvproto/tikv that the protobuf dependecy in Raft is wayyy too strict.


hicqu:
LGTM

Issue #136, opened 2018-10-31
RFC: Support Follower Replication


Hoverbear:
While discussing #135 we determined that supporting replication between two followers, for instance in the same datacenter, to reduce WAN overhead, may be a desirable feature.

This RFC proposes the feature to be added after #135, assuming there will be some structural changes to Raft to support follower to follower communication cleanly.

PR #238, opened 2019-05-12
[WIP] Introduce Follower replication

Fullstop000:
## Implementation
For a rough view, you can see **Detail Design** part in [this doc](https://docs.google.com/document/d/1Sp9Tnc_nk_i0feTOgLGPT3jZYVFjfP7C6pwA-wuVHko/edit?usp=sharing)

The basic implementation points: 
* Two new functions are added to `RawNode`: `remove_imitators` and `add_imitators`.
* Every node (except Leader) maintains an extra part in `ProgressSet` for `imitators` which is used to manage the progress of its imitators. 
* A new field `imitators` is added to `ConfState` for just recovery. And the `imitators` in `Snapshot` is ignored when applying.

## WIP TODOs
* Solid test cases
* `ReadIndex` in an imitator
* Add API to support that an imitator is able to request logs directly from its connector

## Related issues & PRs
#135 #136


Hoverbear:
This is so, so exciting!

PR #243, opened 2019-06-10
Add request snapshot

overvenus:
This adds a feature that allows follower/learner request snapshot actively.

Impls https://github.com/tikv/rfcs/pull/26


proto/eraftpb.proto
```diff
@@ -74,6 +74,7 @@ message Message {
     bool reject = 10;
     uint64 reject_hint = 11;
     bytes context = 12;
+    bool request_snapshot = 13;
```
BusyJay: Can it be moved to L74?

src/errors.rs
```diff
@@ -63,6 +63,10 @@ quick_error! {
         NotExists(id: u64, set: &'static str) {
             display("The node {} is not in the {} set.", id, set)
         }
+        /// The request snapshot if dropped.
```
BusyJay: The sentence doesn't make sense.

src/progress.rs
```diff
@@ -248,8 +248,8 @@ impl Progress {
         self.pending_snapshot = 0;
     }
 
-    /// Unsets pendingSnapshot if Match is equal or higher than
-    /// the pendingSnapshot
+    /// Unsets pending_snapshot if match is equal or higher than
```
BusyJay: s/match/matched/

src/raft.rs
```diff
@@ -107,6 +107,10 @@ pub struct Raft<T: Storage> {
     /// The maximum length (in bytes) of all the entries.
     pub max_msg_size: u64,
 
+    /// The peer is requesting snapshot, it can not handle any appedn messages
```
BusyJay: s/appedn/append/

src/raft.rs
```diff
@@ -1225,7 +1230,10 @@ impl<T: Storage> Raft<T> {
         if pr.state == ProgressState::Replicate && pr.ins.full() {
             pr.ins.free_first_one();
         }
-        if pr.matched < self.raft_log.last_index() {
+        if pr.matched < self.raft_log.last_index()
+            || self.raft_log.term(pr.next_idx - 1).is_err() // Does it request snapshot?
```
BusyJay: How about `pr.next_idx < pr.matched`? Querying term and entries is expensive operations.

src/progress.rs
```diff
@@ -278,19 +283,29 @@ impl Progress {
...
+            }
             return true;
         }
 
-        // the rejection must be stale if "rejected" does not match next - 1
-        if self.next_idx == 0 || self.next_idx - 1 != rejected {
+        // The rejection must be stale if "rejected" does not match next - 1.
+        // Alway consider decreased if the peer is requesting snapshot.
+        if (self.next_idx == 0 || self.next_idx - 1 != rejected)
+            && request_snapshot == INVALID_INDEX
```
BusyJay: The check seems redundant.

src/raft.rs
```diff
@@ -525,8 +530,7 @@ impl<T: Storage> Raft<T> {
         let ents = self.raft_log.entries(pr.next_idx, self.max_msg_size);
         let mut m = Message::new();
         m.set_to(to);
-        if term.is_err() || ents.is_err() {
-            // send snapshot if we failed to get term or entries
+        if term.is_err() || ents.is_err() || pr.pending_request_snapshot != INVALID_INDEX {
```
BusyJay: Better move `pending_request_snapshot` check forward to avoid unnecessary loading entries.

src/raft.rs
```diff
@@ -1710,9 +1717,48 @@ impl<T: Storage> Raft<T> {
         Ok(())
     }
 
+    /// Request a snapshot from a leader.
+    pub fn request_snapshot(&mut self, request_index: u64) -> Result<()> {
+        if !self.is_learner && self.prs().voters().len() == 1 {
```
BusyJay: Seems redundant as it's checked by L1729 and L1736.

src/raft.rs
```diff
@@ -1757,6 +1803,10 @@ impl<T: Storage> Raft<T> {
     /// For a message, commit and send out heartbeat.
     pub fn handle_heartbeat(&mut self, mut m: Message) {
         self.raft_log.commit_to(m.get_commit());
+        if self.pending_request_snapshot != INVALID_INDEX {
```
BusyJay: You may need to check whether `AppendResponse` set `recent_active` to true.

src/raft.rs
```diff
@@ -1710,9 +1717,48 @@ impl<T: Storage> Raft<T> {
...
+            return Err(Error::RequestSnapshotDropped);
+        }
+        if self.get_snap().is_some() {
+            info!(
+                "{} there is a pending snapshot; dropping request snapshot",
+                self.tag
+            );
+            return Err(Error::RequestSnapshotDropped);
+        }
+        self.pending_request_snapshot = request_index;
```
BusyJay: What if there is already a `pending_request_snapshot`?

src/raft.rs
overvenus: `AppendResponse` does set `recent_active` to true

src/raft.rs
```diff
@@ -521,17 +526,23 @@ impl<T: Storage> Raft<T> {
...
+        if pr.pending_request_snapshot != INVALID_INDEX {
+            // Check pending request snapshot first to avoid unnecessary loading entries.
             if !self.prepare_send_snapshot(&mut m, pr, to) {
                 return;
             }
         } else {
-            self.prepare_send_entries(&mut m, pr, term.unwrap(), ents.unwrap());
+            let term = self.raft_log.term(pr.next_idx - 1);
+            let ents = self.raft_log.entries(pr.next_idx, self.max_msg_size);
+            if pr.pending_request_snapshot != INVALID_INDEX || term.is_err() || ents.is_err() {
```
BusyJay: Duplicated with L531.

src/raft.rs
```diff
@@ -1837,6 +1891,7 @@ impl<T: Storage> Raft<T> {
             meta.get_term()
         );
 
+        self.pending_request_snapshot = INVALID_INDEX;
```
BusyJay: Should it also be reset in `reset` method?

src/storage.rs
```diff
@@ -89,6 +90,7 @@ pub struct MemStorageCore {
     // TODO: maybe vec_deque
     // entries[i] has raft log position i+snapshot.get_metadata().get_index()
     entries: Vec<Entry>,
+    trigger_snap_unavailable: bool,
```
BusyJay: Can you add some comments?

tests/integration_cases/test_raft.rs
```diff
@@ -4184,3 +4185,229 @@ fn test_conf_change_check_before_campaign() {
...
+    nt.send(vec![msg.clone()]);
+    assert_eq!(nt.peers[&1].raft_log.committed, 16);
+    assert_eq!(
+        nt.peers[&1].prs().voters()[&2].state,
+        ProgressState::Snapshot
+    );
+    assert_eq!(nt.peers[&2].raft_log.committed, 15);
+
+    // Util snapshot success or fail.
+    let report_ok = new_message(2, 1, MessageType::MsgSnapStatus, 0);
```
BusyJay: I thought it's reported automatically.

src/raft.rs
```diff
@@ -717,9 +730,11 @@ impl<T: Storage> Raft<T> {
 
     /// Converts this node to a follower.
     pub fn become_follower(&mut self, term: u64, leader_id: u64) {
+        let pending_request_snapshot = self.pending_request_snapshot;
```
hicqu: These 2 line seems useless?

src/progress.rs
```diff
@@ -278,25 +283,41 @@ impl Progress {
...
-        self.next_idx = cmp::min(rejected, last + 1);
-        if self.next_idx < 1 {
-            self.next_idx = 1;
+        // Do not decrease next index if it's requesting snapshot.
+        if request_snapshot == INVALID_INDEX {
+            self.next_idx = cmp::min(rejected, last + 1);
+            if self.next_idx < 1 {
+                self.next_idx = 1;
+            }
+        } else if self.pending_request_snapshot == INVALID_INDEX {
```
hicqu: Why `else if` instead of `else`?

src/raft.rs
```diff
@@ -1710,9 +1727,43 @@ impl<T: Storage> Raft<T> {
...
+            self.send_request_snapshot();
+            return Ok(());
+        }
+        Err(Error::RequestSnapshotDropped)
+    }
+
     // TODO: revoke pub when there is a better way to test.
     /// For a given message, append the entries to the log.
     pub fn handle_append_entries(&mut self, m: &Message) {
+        if self.pending_request_snapshot != INVALID_INDEX {
```
hicqu: what if the entries is more new than pending_request_snapshot?

src/raft.rs
```diff
@@ -1794,7 +1849,10 @@ impl<T: Storage> Raft<T> {
 
     fn restore_raft(&mut self, snap: &Snapshot) -> Option<bool> {
         let meta = snap.get_metadata();
-        if self.raft_log.match_term(meta.get_index(), meta.get_term()) {
+        // Do not fast-forward commit if we are requesting snapshot.
```
hicqu: Why?

overvenus: Because the snapshot may be dropped if there is no further proposal after requesting a snapshot.

src/raft.rs
overvenus: It should be dropped if a follower is requesting a snapshot. After applying a snapshot, the follower will receive raft log entries again.

src/progress.rs
overvenus: We don't want to update pending request snapshot if there is one already, let's keep it simple.

src/progress.rs
```diff
@@ -278,25 +283,41 @@ impl Progress {
     /// Returns false if the given index comes from an out of order message.
     /// Otherwise it decreases the progress next index to min(rejected, last)
     /// and returns true.
-    pub fn maybe_decr_to(&mut self, rejected: u64, last: u64) -> bool {
+    pub fn maybe_decr_to(&mut self, rejected: u64, last: u64, request_snapshot: u64) -> bool {
         if self.state == ProgressState::Replicate {
             // the rejection must be stale if the progress has matched and "rejected"
             // is smaller than "match".
-            if rejected <= self.matched {
+            // Or rejected equals to matched and request_snapshot is the INVALID_INDEX.
```
zhangjinpeng1987: Can you give more explanation for this condition?

zhangjinpeng1987: Got it.

src/raft.rs
zhangjinpeng1987: Because reset will reset the pending_request_snapshot, so we should save it first.

src/raft.rs
```diff
@@ -1333,6 +1349,7 @@ impl<T: Storage> Raft<T> {
         // out the next msgAppend.
         // If snapshot failure, wait for a heartbeat interval before next try
         pr.pause();
+        pr.pending_request_snapshot = 0;
```
zhangjinpeng1987: Maybe should use INVALID_INDEX.

src/raft.rs
```diff
@@ -2058,4 +2117,15 @@ impl<T: Storage> Raft<T> {
     pub fn abort_leader_transfer(&mut self) {
         self.lead_transferee = None;
     }
+
+    fn send_request_snapshot(&mut self) {
+        let mut m = Message::new();
+        m.set_msg_type(MessageType::MsgAppendResponse);
+        m.set_index(self.raft_log.committed);
+        m.set_reject(true);
```
zhangjinpeng1987: Why set reject?

src/raft.rs
```diff
@@ -1710,9 +1727,43 @@ impl<T: Storage> Raft<T> {
...
+            return Ok(());
+        }
+        Err(Error::RequestSnapshotDropped)
+    }
+
     // TODO: revoke pub when there is a better way to test.
     /// For a given message, append the entries to the log.
     pub fn handle_append_entries(&mut self, m: &Message) {
+        if self.pending_request_snapshot != INVALID_INDEX {
+            self.send_request_snapshot();
```
zhangjinpeng1987: Does the leader can tolerant follower `send_request_snapshot` multiple times with the same `pending_request_snapshot`?

src/storage.rs
```diff
@@ -89,6 +90,9 @@ pub struct MemStorageCore {
     // TODO: maybe vec_deque
     // entries[i] has raft log position i+snapshot.get_metadata().get_index()
     entries: Vec<Entry>,
+    // If the it is true, the next snapshot will return a
```
zhangjinpeng1987: If it is true.

src/raft.rs
overvenus: It needs to be rejected, so it can enter requesting snapshot state in leader's view. See more at `handle_append_response`.

src/raft.rs
overvenus: Yes, because request snapshot message may be dropped on the network, it needs to be able to send the same message multiple times.

overvenus:
PTAL @hicqu @BusyJay Thanks!

PR #244, opened 2019-06-11
[five_mem_example] Add terminate signal to stop the threads after pro…

Ishiihara:
This PR adds terminate signals to stop the threads after all proposals has been accepted.


hicqu:
PTAL @Hoverbear

PR #247, opened 2019-06-18
Move protos to a separate crate

nrc:
Makes Raft a workspace project

Motivated by https://github.com/pingcap/kvproto/issues/280

PTAL @Hoverbear @BusyJay


proto/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[features]
+default = []
+gen = []
+
+[build-dependencies]
+protobuf-build = "0.6"
+
+[dependencies]
+bytes = "0.4.11"
+lazy_static = "1.3.0"
```
BusyJay: Why does it depend on lazy_static?

nrc: Some of the rust-protobuf API we are emulating requires an initialised default with static lifetime, lazy_static is the best way to support that.

BusyJay: Interesting. It may also be the reason why it doesn't perform as expected.

Hoverbear: Could `const fn` help solve it one day?

nrc: Maybe, I'm not sure. I hope we can just eliminate uses of that wrapper function and then remove lazy static

PR #249, opened 2019-07-01
[WIP] Introduce follower replication

Fullstop000:
## New features
* Add a new message type `MsgBroadcast` for asking a node to send raft logs to others
* Add a new protocol `Commission` representing the `Append` or `Snapshot` operation to a node
* A message now could be sent to a **proxy** and the proxy redirects the message to the destination

## TODOs
- [ ] Follower's handing when receiving a `MsgBroadcast`
- [ ] Functional test cases
- [ ] E2E tests


Fullstop000:
I've finished the main part of this PR. But I'm not sure the best way to implement the `ConfChange` for altering raft groups dynamically so I want a discussion with your guys.

PTAL @siddontang @hicqu @BusyJay @Hoverbear

Hoverbear:
@Fullstop000 Can you please include links to any design docs you worked on?

Fullstop000:
@Hoverbear Can I make a WIP PR in tikv/rfc ？

Hoverbear:
@Fullstop000 I think it's OK. :) I'd still really appreciate seeing an RFC (like https://docs.google.com/document/d/1Sp9Tnc_nk_i0feTOgLGPT3jZYVFjfP7C6pwA-wuVHko/edit?ts=5c8c5fe3 which you worked on)

Hoverbear:
@hicqu PTAL

src/raft_log.rs
```diff
@@ -416,7 +416,7 @@ impl<T: Storage> RaftLog<T> {
         }
 
         let length = self.last_index() + 1 - first_index;
-        if low < first_index || high > first_index + length {
+        if high > first_index + length {
```
hicqu: Why remove the first check?

src/raft.rs
```diff
@@ -606,12 +649,74 @@ impl<T: Storage> Raft<T> {
     pub fn bcast_append(&mut self) {
         let self_id = self.id;
         let mut prs = self.take_prs();
-        prs.iter_mut()
-            .filter(|&(id, _)| *id != self_id)
-            .for_each(|(id, pr)| self.send_append(*id, pr));
+        if self.follower_delegate {
```
hicqu: Is it possible to not change `bcast_append`, but do all logics in `send_append`?

src/raft_log.rs
Fullstop000: I think it has already been checked in here https://github.com/tikv/raft-rs/blob/7123d06fb557d26d5e453e32328ab4f0955a87cd/src/raft_log.rs#L414-L420

src/raft.rs
Fullstop000: I think both of these two approaches need to pick the delegate of each group first?

hicqu: I think `send_append` can be reused for this (maybe also need to do a little change for `send_append`). For many calls to `send_append` with a same delegate, we can just append a new `Commission` for it.

Fullstop000: I'm worried about too much state will be maintained in such an implementation.

src/raft.rs
```diff
@@ -2225,4 +2535,364 @@ impl<T: Storage> Raft<T> {
...
+                    return delegate;
+                } else {
+                    let term = self.raft_log.term(target_pr.next_idx - 1);
+                    let ents = self.raft_log.entries(target_pr.next_idx, self.max_msg_size);
+                    if target_pr.pending_request_snapshot != INVALID_INDEX
+                        || term.is_err()
+                        || ents.is_err()
+                    {
+                        // Let the leader do its job if the target follower needs snapshot
+                        return delegate;
```
hicqu: In the current implementation leader sends snapshots to followers directly, the delegate is bypassed.
However I think we can let the delegate sends snapshots to followers. Here the leader needs only call `pr.become_snapshot`.
And, if the delegate meets `SnapshotTemporarilyUnavailable` when sending the snapshot, it will report it to the leader in `BroadcastResponse`.
After that, leader can call `pr.snapshot_failure()` and `pr.become_probe()` and then probe the follower.

src/raft.rs
```diff
@@ -1465,6 +1628,7 @@ impl<T: Storage> Raft<T> {
                 if pr.state == ProgressState::Replicate {
                     pr.become_probe();
                 }
+                self.groups.remove_delegate(m.from);
```
hicqu: What if it's delegate? We need to tell the leader the target becomes probe status.

hicqu: If delegates also handle peers' progress, it's ok.

src/group/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+/// The option used for feature [Follower Replication](https://github.com/tikv/rfcs/pull/33/files)
+#[derive(Clone, Debug)]
+pub struct FollowerReplicationOption {
+    /// The leader choose a follower in a raft group as a delegate and ask it to send entries to
+    /// the rest group members.
+    /// If there is no group configured, no delegate will be picked
+    pub follower_delegate: bool,
+
+    /// The raft group definition. See [`GroupsConfig`](group/struct.GroupsConfig.html) for detail.
+    pub groups: GroupsConfig,
```
hicqu: I think it's not necessary if the group is built dynamically with `group_id` in `Message`s.

src/group/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    /// The max number of members of a group of which contains leader and the leader never pick a delegate.
+    /// If the group size is larger than this, a delegate will be picked even the leader belongs to this group.
+    pub max_leader_group_no_delegate: usize,
```
hicqu: Personally I prefer to remove this, because without explicit information about `group_id`, follower replication may be not helpful. For example 7 peers in one datacenter, follower replication can't save any network traffics.

src/group/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+        Self {
+            meta,
+            indexes,
+            ..Default::default()
+        }
+    }
+
+    /// Return the <group id>-<node id> pairs iterator
+    #[inline]
+    pub fn meta_iter(&self) -> Iter<'_, u64, Vec<u64>> {
```
hicqu: Seems it's unnecessary. So is field `meta`.

src/raft.rs
```diff
@@ -2215,14 +2355,103 @@ impl<T: Storage> Raft<T> {
...
+    // Our target is to reduce the workload in the leader which means the group
+    // delegate should be able to send entries as much as possible. Therefore,
+    // the delegate must not be requiring a snapshot (either or requesting
+    // a snapshot) from the perspective of leader. And the more entries the delegate
+    // is able to send to the other group members, the lower pressure the leader has
+    // syncing raft logs to peers in typical log replication.
+    // Also, A peer with smaller 'match' is able to receive more un-compacted entries from leader
+    // and then send them to others in same group.
+    //
+    fn pick_delegate(&self, group: &[u64], prs: &ProgressSet) -> u64 {
```
hicqu: Seems in the new implementation, choose the peer with most raft logs is better?

Fullstop000: It seems more reasonable. Actually maybe we can choose any unpaused peer?

src/raft.rs
```diff
@@ -130,6 +130,9 @@ pub struct Raft<T: Storage> {
     /// The list of messages.
     pub msgs: Vec<Message>,
 
+    /// The especial map for messages to be sent to group delegates.
+    pub delegated_msgs: HashMap<u64, Message>,
```
hicqu: I prefer to put it with `groups`

src/raft.rs
```diff
@@ -545,41 +546,159 @@ impl<T: Storage> Raft<T> {
         is_batched
     }
 
+    // Whether the given peer could use Follower Replication
+    fn use_delegate(&self, to: u64) -> bool {
```
hicqu: I prefer
```rust
fn use_delegate(&self, to: u64) -> bool {
    self.is_leader() &&
        self.groups.get_members(to).map_or(0, |members| members.len()) > 1 &&
        !self.groups.in_same_group(self.id, to)
}
```
And we need to make sure that peers with invalid group_id shouldn't exist in `groups`.

src/raft.rs
```diff
@@ -545,41 +546,159 @@ impl<T: Storage> Raft<T> {
...
+        self.group_id == INVALID_ID
+    }
+
     /// Sends RPC, with entries to the given peer.
-    pub fn send_append(&mut self, to: u64, pr: &mut Progress) {
+    /// This can be called by the leader or a delegate. If the delegate call this, the field `to`
+    /// of produced message should be set to the leader id.
+    pub fn send_append(&mut self, to: u64, prs: &mut ProgressSet) {
+        if self.use_delegate(to) {
+            if let Some(gid) = self.groups.get_group_id(to) {
```
hicqu: Too many indents. Maybe we can define `use_delegate(to) -> Option<group_id>`.

src/raft.rs
```diff
@@ -545,41 +546,159 @@ impl<T: Storage> Raft<T> {
...
+                        return;
+                    }
+                    None => {
+                        let members = self.groups.get_members(to).cloned().unwrap(); // this is safe because of the checking in `self.use_delegate`
+                        let delegate = match self.groups.get_delegate(gid) {
+                            Some(cached) => cached,
+                            None => self.pick_delegate(&members, prs),
+                        };
+                        if delegate != INVALID_ID {
+                            if let Some(pr) = prs.get_mut(delegate) {
```
hicqu: I think calling `unwrap` is better, because the progress must exist.

src/raft.rs
```diff
@@ -545,41 +546,159 @@ impl<T: Storage> Raft<T> {
...
+                        if let Some(pr) = prs.get_mut(to) {
+                            // This just clean all the Inflights though there could be some MsgAppendResp on the way. And from
+                            // now the delegate will take over the flow control of the peer `to`.
+                            pr.become_delegated();
+                        }
+                        return;
+                    }
+                    None => {
+                        let members = self.groups.get_members(to).cloned().unwrap(); // this is safe because of the checking in `self.use_delegate`
+                        let delegate = match self.groups.get_delegate(gid) {
```
hicqu: I prefer `let delegate = self.pick_delegate(prs)`, which already contains `get_delegate` logic. And, its valid only if the `pr` is not paused.

src/raft.rs
```diff
@@ -1281,10 +1397,80 @@ impl<T: Storage> Raft<T> {
                 pr.become_probe();
             }
             ProgressState::Replicate => pr.ins.free_to(m.index),
+            // In the leader's ProgressSet, the flow control of a `Delegated` progress is turned off
+            ProgressState::Delegated => {}
         }
         *maybe_commit = true;
     }
 
+    fn handle_append_response_in_delegate(
```
hicqu: Seems only difference between this and `handle_append_response` is the latter needs to handle leader transferring. That logic is compatible for delegates. So I think this function is unnecessary.

src/raft.rs
```diff
@@ -1617,20 +1805,24 @@ impl<T: Storage> Raft<T> {
...
-            let mut prs = self.take_prs();
-            self.send_append(from, prs.get_mut(from).unwrap());
-            self.set_prs(prs);
+            if m.from_delegate == INVALID_ID {
+                let mut prs = self.take_prs();
+                self.send_append(from, &mut prs);
+                self.set_prs(prs);
+            } else {
+                let members = self.groups.get_members(m.from_delegate).cloned();
+                self.bcast_append(members.as_ref());
```
hicqu: `bcast_append` calls `send_append`, which checks members' delegate again. I believe it can be improved, maybe we can take a try?

src/raft.rs
```diff
@@ -1811,14 +1993,63 @@ impl<T: Storage> Raft<T> {
                 };
                 self.read_states.push(rs);
             }
+            MessageType::MsgAppendResponse => {
+                if self.prs().get(m.from).is_none() {
```
hicqu: It's a common case for receiving responses from removed peers. So I think the log is unnecessary. Without it the code could be cleaner.

src/raft.rs
```diff
@@ -2215,14 +2450,93 @@ impl<T: Storage> Raft<T> {
...
+            id;
+            "progress" => ?progress,
+        );
+    }
+
+    // Pick a delegate of the given group
+    //
+    // The delegate must satisfy conditions below:
+    // 1. Must be 'recent_active'
+    // 2. The progress state should be 'Replicate' but not 'paused'
```
hicqu: Seems `Probe` needs also to be allowed?

Fullstop000: `Probe` might not be suitable for a delegate in the leader's view because the `Probe` member can probably just recover from a crash or is suffering the network partition.

src/raft.rs
```diff
@@ -534,41 +531,186 @@ impl<T: Storage> Raft<T> {
...
+            && !self.groups.in_same_group(self.id, to)
+        {
+            self.groups.get_group_id(to)
+        } else {
+            None
+        }
+    }
+
+    #[inline]
+    fn is_follower_replication_enabled(&self) -> bool {
```
hicqu: I think it's not accurate. Every peer can't know `follower replication` is enabled or not. It just choose a strategy after receives followers' messages.

Fullstop000: We can add some comments to make this clear.

harness/src/interface.rs
```diff
@@ -48,7 +48,11 @@ impl Interface {
     /// Read messages out of the raft.
     pub fn read_messages(&mut self) -> Vec<Message> {
         match self.raft {
-            Some(_) => self.msgs.drain(..).collect(),
+            Some(_) => {
+                let mut msgs = self.msgs.drain(..).collect::<Vec<_>>();
```
Hoverbear: You might be able to avoid reallocation here by combining `drain` and `chain`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=47f3604fa7d8fa5cbe9b3105d47b4d68

harness/src/network.rs
```diff
@@ -150,13 +150,21 @@ impl Network {
...
     /// Note: messages are not filtered by any configured filters.
+    #[inline]
     pub fn read_messages(&mut self) -> Vec<Message> {
         self.peers
             .iter_mut()
             .flat_map(|(_peer, progress)| progress.read_messages())
             .collect()
     }
 
+    #[inline]
```
Hoverbear: @nrc advises us in the PingCAP style guide that this is generally unnecessary https://pingcap.github.io/style-guide/rust/functions.html#functions

harness/tests/integration_cases/test_raft_follower_replication.rs
```diff
@@ -0,0 +1,546 @@
+// Copyright 2019 PingCAP, Inc.
```
Hoverbear: Please use our new copyright, sorry! :)

harness/src/network.rs
Fullstop000: Good advice! I'll remove the `#[inline]` involved in this pr. And maybe later I can post a PR to remove them all.

Fullstop000: @Hoverbear I heard about that removing `#[inline]` might hurt the performance in practice 😂.

src/raft.rs
```diff
@@ -380,6 +386,12 @@ impl<T: Storage> Raft<T> {
         self.skip_bcast_commit = skip;
     }
 
+    /// Returns the inner raft group configuration
+    #[inline]
+    pub fn groups(&self) -> &Groups {
```
hicqu: I think it's better to make `Groups` be private for external view. So I prefer `pub fn groups(&self) -> HashMap<group-id, Vec<peer-id>>`.

src/raft.rs
```diff
@@ -175,6 +175,10 @@ pub struct Raft<T: Storage> {
     min_election_timeout: usize,
     max_election_timeout: usize,
 
+    /// Raft groups inner state
+    groups: Groups,
+    group_id: u64,
```
hicqu: If `group_id` is not pub, I think we need to add an API to change it.

src/raft.rs
```diff
@@ -472,38 +486,16 @@ impl<T: Storage> Raft<T> {
...
-            "progress" => ?pr,
-        );
-        pr.become_snapshot(sindex);
-        debug!(
-            self.logger,
-            "paused sending replication messages to {}",
-            to;
-            "progress" => ?pr,
-        );
+        self.pr_become_snapshot_and_log(to, pr, sindex, sterm);
```
hicqu: I guess `pr_become_snapshot` is better.

src/raft.rs
```diff
@@ -546,7 +538,15 @@ impl<T: Storage> Raft<T> {
             return;
         }
         let mut m = Message::default();
+        if self.leader_id != self.id {
+            // Self is a delegate, bring the leader id to the other member
+            // TODO: This should be safe as the delegate just calls `handle_append_message` which
+            // means it calls `become_follower`. Is that true?
```
hicqu: It's true because after a candidate peer receives MsgAppend from leader, it calls `become_follower`.

src/raft.rs
```diff
@@ -565,12 +565,24 @@ impl<T: Storage> Raft<T> {
...
     }
 
+    fn send_append_with_bcast_targets(&mut self, to: u64, pr: &mut Progress, targets: Vec<u64>) {
+        let old_msg_len = self.msgs.len();
+        self.send_append(to, pr);
+        let new_msg_len = self.msgs.len();
+        if new_msg_len > old_msg_len {
+            self.msgs[old_msg_len].set_bcast_targets(targets);
+        } else {
+            // TODO(qupeng): is it possible? Remove it later.
```
hicqu: I still can't make sure it's possible or not...

src/raft.rs
```diff
@@ -590,20 +602,62 @@ impl<T: Storage> Raft<T> {
         self.send(m);
     }
 
-    /// Sends RPC, with entries to all peers that are not up-to-date
+    /// Sends RPC, with entries to all or given peers that are not up-to-date
```
hicqu: s/or given //g

src/raft.rs
```diff
@@ -1615,12 +1675,22 @@ impl<T: Storage> Raft<T> {
             }
         }
 
+        // The leader receives the reject message from a delegate
         if send_append {
             let from = m.from;
             let mut prs = self.take_prs();
-            self.send_append(from, prs.get_mut(from).unwrap());
+            if m.delegate != INVALID_ID {
```
hicqu: Seems it's wrong. I think it should be
```rust
if m.delegate != INVALID_ID {
    // The follower receives entries from a delegate, skip send append to it.
} else if !m.bcast_targets().is_empty() {
    // The delegate rejects the last append message.
    self.send_append_with_bcast_targets(m.from);
} else {
    send.send_append(from, ...);
}
```

src/raft.rs
```diff
@@ -2124,6 +2257,20 @@ impl<T: Storage> Raft<T> {
         }
     }
 
+    /// Takes the `Groups`.
```
hicqu: These APIs conflict with other parts. I guess what we only need is `fn set_group_id(&mut self, id: u64)`?

src/progress/progress_set.rs
```diff
@@ -421,20 +427,27 @@ impl ProgressSet {
...
     /// This should only be called by the leader.
-    pub fn quorum_recently_active(&mut self, perspective_of: u64) -> bool {
+    pub fn quorum_recently_active<F>(&mut self, perspective_of: u64, mut f: F) -> bool
+    where
+        F: FnMut(u64, bool),
+    {
         let mut active = HashSet::default();
-        for (&id, pr) in self.voters_mut() {
+        let learners = self.configuration.take_learners();
+        for (&id, pr) in self.iter_mut() {
```
hicqu: So if the peer is a learner, it will be inserted into `active`? I think it's wrong.

TennyZhuang:
Any updates?

Xuanwo:
This PR is discontinued for so long, any updates here? Do we have a plan to get it merged?

PR #250, opened 2019-07-01
Bump three dependencies

ice1000:
Signed-off-by: ice1000 <ice1000kotlin@foxmail.com>

As title.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,8 @@ bytes = "0.4.11"
 slog = "2.2"
 quick-error = "1.2.2"
 raft-proto = { path = "proto" }
-rand = "0.6.4"
-hashbrown = "0.3"
+rand = "0.7.0"
+hashbrown = "0.5"
 fail = { version = "0.2", optional = true }
 getset = "0.0.7"
 slog-stdlog = "3.0.2"
--- a/harness/Cargo.toml
+++ b/harness/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 # Make sure to synchronize updates with Raft.
 [dependencies]
 raft = { path = ".." }
-rand = "0.6.4"
+rand = "0.7.0"
 slog = "2.2"
 slog-term = "2.4.0"
 slog-envlogger = "2.1.0"


ice1000:
```
error[E0658]: imports can only refer to extern crate names passed with `--extern` on stable channel (see issue #53130)
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.6/src/error_impls.rs:13:5
   |
8  | extern crate std;
   | ----------------- not an extern crate passed with `--extern`
...
13 | use std::{error, io};
   |     ^^^
   |
note: this import refers to the extern crate imported here
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.6/src/error_impls.rs:8:1
   |
8  | extern crate std;
   | ^^^^^^^^^^^^^^^^^
                                                                                
error[E0658]: imports can only refer to extern crate names passed with `--extern` on stable channel (see issue #53130)
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.6/src/use_file.rs:15:5
   |
10 | extern crate std;
   | ----------------- not an extern crate passed with `--extern`
...
15 | use std::{fs::File, io::Read};
   |     ^^^
   |
note: this import refers to the extern crate imported here
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.6/src/use_file.rs:10:1
   |
10 | extern crate std;
   | ^^^^^^^^^^^^^^^^^
                                                                                
error[E0658]: imports can only refer to extern crate names passed with `--extern` on stable channel (see issue #53130)
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.6/src/linux_android.rs:15:5
   |
10 | extern crate std;
   | ----------------- not an extern crate passed with `--extern`
...
15 | use std::io;
   |     ^^^
   |
note: this import refers to the extern crate imported here
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.6/src/linux_android.rs:10:1
   |
10 | extern crate std;
   | ^^^^^^^^^^^^^^^^^
                                                                                
error: aborting due to 3 previous errors
```

Why? Does this mean we have to drop rust 1.31 support?

nrc:
The crate getrandom only supports Rust 1.32 and later (see https://github.com/rust-random/getrandom and brief discussion at https://github.com/rust-random/getrandom/pull/15#pullrequestreview-217912149), so if we want to keep that dep, then I think we to drop 1.31 support (which isn't such a bad thing, IMO we should only support current stable).

ice1000:
> The crate getrandom only supports Rust 1.32 and later (see https://github.com/rust-random/getrandom and brief discussion at [rust-random/getrandom#15 (review)](https://github.com/rust-random/getrandom/pull/15#pullrequestreview-217912149)), so if we want to keep that dep, then I think we to drop 1.31 support (which isn't such a bad thing, IMO we should only support current stable).

Ok (and it sounds like it's not an issue of my own dependency bump)

ice1000:
PTAL @Hoverbear

PR #256, opened 2019-07-05
Eliminate usages of setters generated by the wrapper

ice1000:
Similar to #218, but for setters.

FWIW I'm using this regex (this does not always work): `\.set_(index|from|term|to|learners|nodes|data|entries|commit|vote|log_term|reject)\(([^\(\)]+)\)`


ice1000:
PTAL @nrc @Hoverbear

ice1000:
PTAL @overvenus

PR #259, opened 2019-07-11
Port #243 to the Prost branch

nrc:
This is a straight port of #243, but not a cherry-pick because a few changes needed to be made to use Prost, the diff didn't apply cleanly because of earlier changes to the branch, and this branch is on the 2018 edition

PTAL @overvenus @ice1000

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -26,6 +26,7 @@
 // limitations under the License.
 
 use crate::errors::Error;
+use crate::raft::INVALID_INDEX;
 use fxhash::FxHashMap;
 use std::cmp;
 use std::collections::hash_map::HashMap;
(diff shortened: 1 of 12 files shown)


src/progress.rs
```diff
@@ -278,25 +283,42 @@ impl Progress {
...
-            if rejected <= self.matched {
+            // Or rejected equals to matched and request_snapshot is the INVALID_INDEX.
+            if rejected < self.matched
+                || (rejected == self.matched && request_snapshot == INVALID_INDEX)
+            {
                 return false;
             }
             self.next_idx = self.matched + 1;
+            if request_snapshot == INVALID_INDEX {
+                self.next_idx = self.matched + 1;
```
ice1000:
```rust
self.next_idx = self.matched + 1;
if request_snapshot == INVALID_INDEX { 
    self.next_idx = self.matched + 1; 
```

Are you sure? I think there's a better way of writing this.

nrc: I just applied the diff and fixed it up. This is from https://github.com/pingcap/raft-rs/pull/243/files#diff-2731f87d144f80edb31dac3d81fd9107R296

ice1000: This line was deleted https://github.com/pingcap/raft-rs/pull/243/files#diff-2731f87d144f80edb31dac3d81fd9107L288

So in your case you should probably delete line 288 too

nrc: Oh right, d'oh. Thanks

Issue #223, opened 2019-04-24
Getters of public API are not conformant to RFC 344


fin-ger:
**Is your feature request related to a problem? Please describe.**
In [RFC 344](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis) the idiomatic naming for getters and setters in rust is described. This library contains many methods in its public API that are not conformant to this RFC:

```
$ grep -RPn "pub fn.*get_" src
src/raft.rs:326:    pub fn get_store(&self) -> &T {
src/raft.rs:338:    pub fn get_snap(&self) -> Option<&Snapshot> {
src/raft.rs:384:    pub fn get_election_timeout(&self) -> usize {
src/raft.rs:389:    pub fn get_heartbeat_timeout(&self) -> usize {
src/raft.rs:394:    pub fn get_heartbeat_elapsed(&self) -> usize {
src/raft.rs:399:    pub fn get_randomized_election_timeout(&self) -> usize {
src/util.rs:86:    pub fn get_voters(&self) -> &[u64] {
src/progress/progress_set.rs:308:    pub fn get_mut(&mut self, id: u64) -> Option<&mut Progress> {
src/raw_node.rs:399:    pub fn get_snap(&self) -> Option<&Snapshot> {
src/raw_node.rs:483:    pub fn get_store(&self) -> &T {
src/prost/wrapper_eraftpb.rs:16:    pub fn get_entry_type(&self) -> EntryType {
src/prost/wrapper_eraftpb.rs:28:    pub fn get_term(&self) -> u64 {
src/prost/wrapper_eraftpb.rs:40:    pub fn get_index(&self) -> u64 {
src/prost/wrapper_eraftpb.rs:52:    pub fn get_data(&self) -> &[u8] {
src/prost/wrapper_eraftpb.rs:72:    pub fn get_context(&self) -> &[u8] {
src/prost/wrapper_eraftpb.rs:92:    pub fn get_sync_log(&self) -> bool {
src/prost/wrapper_eraftpb.rs:179:    pub fn get_conf_state(&self) -> &ConfState {
src/prost/wrapper_eraftpb.rs:209:    pub fn get_pending_membership_change(&self) -> &ConfState {
src/prost/wrapper_eraftpb.rs:237:    pub fn get_pending_membership_change_index(&self) -> u64 {
src/prost/wrapper_eraftpb.rs:249:    pub fn get_index(&self) -> u64 {
[...]

PR #260, opened 2019-07-12
*: support rust-protobuf and prost at the same time

BusyJay:
After this PR, people can choose between rust-protobuf and prost without changing any code.

This PR keeps compatible with both rust-protobuf and prost, users can choose either via features. Under the hook, proc-macros is used to generate a compatible layer for prost instead of writing ugly wrappers. 

This solution is also portable. To make a project support both rust-protobuf and prost at the same time, you just need to add two features *repr-protobuf* and *repr-prost*. And then use jinkela-build to generate codes and include the generated sources in module files.

One more thing: protoc is not required anymore.

*repr-protobuf* is enabled by default as it's more production-proven.


README.md
```diff
@@ -61,6 +61,13 @@ cargo watch -s "cargo check"
 
 See [instructions](proto/README.md) in the proto subdirectory.
 
+Both rust-protobuf and prost are supported. rust-protobuf is used by default. If you wish to use prost, you need to enable the prost features.
+
+```
+[dependencies]
+raft = { git = "https://github.com/pingcap/raft-rs.git", default-features = false, features = ["repr-prost"] }
+```
```
Hoverbear: > To make a project support both rust-protobuf and prost at the same time, you just need to add two features repr-protobuf and repr-prost. And then use jinkela-build to generate codes and include the generated sources in module files.

What about this part?

BusyJay: It's instructions for using the jinkela library itself, not related to raft.

src/raft.rs
```diff
@@ -467,10 +467,10 @@ impl<T: Storage> Raft<T> {
             "msg" => ?m,
         );
         m.from = self.id;
-        if m.msg_type() == MessageType::MsgRequestVote
-            || m.msg_type() == MessageType::MsgRequestPreVote
-            || m.msg_type() == MessageType::MsgRequestVoteResponse
-            || m.msg_type() == MessageType::MsgRequestPreVoteResponse
+        if m.get_msg_type() == MessageType::MsgRequestVote
```
Hoverbear: This (and many other getters) violates https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis, see https://github.com/pingcap/raft-rs/issues/223

README.md
Hoverbear: Can you add it to https://github.com/BusyJay/jinkela/blob/master/README.md ?

src/raft.rs
BusyJay: I'm afraid the original method is not safe to be used. prost will drop unknown enums silently with `value()`. `get_value` makes sure unknown enums will lead to panic, which is not the best way but ensures data consistency.

README.md
BusyJay: Of course. May do it when I wake up.

BusyJay:
The library jinkela provides a compatible layer which ensures that we can switch between rust-protobuf and prost seamlessly. It has two benifits:
1. If prost doesn't work as expected like what I mention in https://github.com/pingcap/raft-rs/pull/260#discussion_r303163667, we still have plan B.
2. If we want to test more implements of rust-protobuf, we can just make jinkela to be compatible with them.
3. We don't need to maintain a branch for prost anymore, which can be conflict with master easily.
4. Works on features development and migrating on prost can be do concurrently now.

Hoverbear:
@BusyJay Can you rustfmt this PR so it can pass CI?

Cargo.toml
```diff
@@ -16,33 +16,32 @@ edition = "2018"
 members = ["proto"]
 
 [features]
-default = []
+default = ["repr-protobuf"]
 # Enable failpoints
 failpoint = ["fail"]
+repr-prost = ["raft-proto/repr-prost", "harness/repr-prost"]
+repr-protobuf = ["raft-proto/repr-protobuf", "harness/repr-protobuf"]
```
nrc: Could we use `protobuf-codec` and `prost-codec`? It's a convention across a few crates and using the same feature names everywhere keeps things less confusing.

proto/src/lib.rs
```diff
@@ -1,8 +1,10 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-pub use crate::prost::eraftpb;
+mod protos {
+    include!(concat!(env!("OUT_DIR"), "/protos/mod.rs"));
+}
 
-mod prost;
+pub use protos::eraftpb;
```
nrc:
```suggestion
pub use crate::protos::eraftpb;
```

proto/src/lib.rs
```diff
@@ -1,8 +1,10 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-pub use crate::prost::eraftpb;
+mod protos {
+    include!(concat!(env!("OUT_DIR"), "/protos/mod.rs"));
```
nrc: Could we get this path from Jinkela? Otherwise it is a bit magic and keeping it up to date might be annoying

src/errors.rs
```diff
@@ -49,20 +49,13 @@ quick_error! {
...
             description(desc)
         }
-        /// A Prost message encode failed in some manner.
-        ProstEncode(err: EncodeError) {
+        /// Error when handling ser/de.
+        Codec(err: ProtobufError) {
             from()
             cause(err)
             description(err.description())
             display("prost encode error {:?}", err)
```
nrc:
```suggestion
             display("protobuf encode/decode error {:?}", err)
```

nrc:
Thanks for getting involved and working on the branch issue. The code changes look fine, I left a few minor comments inline. There is an alternative to this using the current approach and Cargo features. This works in TiPB and grpc-rs, we only took the Prost-only approach here because we didn't think it would be necessary. I think using Jinkela has some advantages, but it is a change of direction, so I want to confirm the requirements and consider what is the right approach.

This PR seems to be addressing a lot of separable requirements and I'm unclear about which are high priority:
  * allow using both Prost and rust-protobuf on one branch (this is the only one which blocks using kvproto master and I think the highest priority one)
  * generate code in target dir rather than check it in to source
  * simplify build.rs
  * address the unknown enum variant issue
  * replace code generation with procedural macros
  * don't require protoc
  * support other protobuf libraries

AIUI, all of these could be met with adjustments to the current approach and don't require the new approach. On the otherhand, Jinkela does present a stronger compatibility layer. In general, Jinkela seems to have different trade-offs compared to the current approach, rather than being objectively better. One concern is that the current approach has evolved quite a lot over the past few months and we've fixed quite a lot of bugs, I worry that switching approaches now to something which has not been proven out risks introducing bugs or other unforeseen issues. In particular, raft-rs is one of the simpler repos in terms of how it uses protobuf, we might need more changes for kvproto, and will definitely need a lot more for TiKV.

So, the requirements

### allow using both Prost and rust-protobuf on one branch

This can be accomplished with the existing approach, see TiDB or grpc-rs. A PoC for raft is at https://github.com/nrc/raft-rs/tree/prost-both.

### generate code in target dir rather than check it in to source

This can be done with protobuf-build (we do this in grpc-rs). I've tried to change projects as little as possible which is why I have not done this here. I personally think this is a good idea, but it does have downsides: it makes it more difficult to know what API exists. I think we should get some buy-in from the rest of the team before making this change.

### simplify build.rs

My goal with protobuf-build was to support maximum flexibility so that I could preserve the differing project structures of each project. If we're happy to unify structures (e.g., by always generating code into the target directory), then we can make build.rs in the current approach much simpler.

### address the unknown enum variant issue

Addressed in protobuf-build 0.7 (https://github.com/tikv/protobuf-build/pull/20). We'd still need to make some changes in raft itself (as in this PR, use `get_foo` for enums). This is a separable issue.

### replace code generation with procedural macros

I don't see much value in this. Could you explain why you think the current wrappers are ugly? It seems very similar to what Jinkela is doing.

### don't require protoc

Prost doesn't require protoc in any case. We currently never require protoc unless you are rebuilding the protos. If we want to address this, I don't think it is difficult.

### support other protobuf libraries

I don't think this is much harder with the current approach. Jinkela does encapsulate things a bit better, if we think this is desirable, then we should adopt that. Personally I don't think we should support other protobuf libraries and I look forward to just having a single one rather than two.

Cargo.toml
BusyJay: Sure.

proto/src/lib.rs
BusyJay: Actually the path is given by build.rs. So that it does not have to be put inside build script or even the same directory. But I think your concern is correct, maybe we can output an environment variable in build script instead.

BusyJay:
> I worry that switching approaches now to something which has not been proven out risks introducing bug

Bugs fixed lately are either due to codec library itself or the build tools. Problems like recursion limit or unknown enum variables has been fixed in rust-protobuf long time ago. The fixes in prost will be kept using in jinkela library. I very appreciated your work on prost in the past few months, without your work, prost may still not meet our goal.

>  we might need more changes for kvproto, and will definitely need a lot more for TiKV

It's true that we still need some changes to jinkela for kvproto, but overall the code in kvproto should become more clean. And TiKV should do less than before.

>This can be accomplished with the existing approach, 

Jinkela ensures that only switching feature can switch different codecs without any code modifications.

>  it makes it more difficult to know what API exists.

I had the same feelings when I saw the code generated by prost, which generates API by proc-macros! But I think users can and probably should see the documentations for the API of a library rather than digging into the implementations.

> I don't see much value in this. Could you explain why you think the current wrappers are ugly? It seems very similar to what Jinkela is doing.

It doesn't require extra file to work, and it's more rusty. But I don't think it's a required goal to meet, it's just a nice to have.

> never require protoc unless you are rebuilding the protos

Indeed. This PR just remove the requirement totally.

> If we want to address this, I don't think it is difficult.

We have been always suffered from this. Issues are even opened for TiKV.

> Personally I don't think we should support other protobuf libraries and I look forward to just having a single one rather than two.

Both rust-protobuf and prost are great libraries. It's true that rust-protobuf has been proven by production for quite a long time. It's more features complete and has been approved and actively maintained. prost on the other hand is still yound, and has been found several bugs in the past few months. I think it's too aggressively to switch to prost all in once. We should have a plan B.

We are very open minded on switching codecs implement. That's why we accept switching from rust-protobuf to prost. If there is another codecs library with better performance we can definitely switch to it. With the help of jinkela, the switch will not be as hard as today. (And maybe it's time to test Cap’n Proto 😸)

@siddontang what do you think?

nrc:
Sorry, to clarify, I am not suggesting that we use only Prost, I am suggesting that we permit using Prost and rust-protobuf using protobuf-build rather than Jinkela.

I think if we want to support multiple libraries long-term we should think about designing that abstraction layer - I'm not sure how we can make cap'n proto work with grpc-rs without some modification (even more so with Tower).

> Bugs fixed lately are either due to codec library itself or the build tools

I mean primarily, bugs in the build compatibility layer, I found lots of bugs when using protobuf-build in different repos etc., and I expect we will with Jinkela too.

> Jinkela ensures that only switching feature can switch different codecs without any code modifications.

This is a very nice feature! I think that if we plan to support multiple libraries long-term, then we should aim for this everywhere. Note that using protobuf-build, this is also true for raft-rs, with https://github.com/nrc/raft-rs/tree/prost-both you can build with either `--features prost-codec` or `--features protobuf-codec` and not change anything else.

It is only in TiKV where this is not possible, and I think with some effort it will be possible there too, it is only because it has not been a goal that I haven't tried to do it.

> But I think users can and probably should see the documentations for the API of a library rather than digging into the implementations.

I agree :-) But I think we should at least check with others.

> It doesn't require extra file to work

I think we must still have some extra files because the grpc codegen outputs code for protobuf services. But any generated files can always be done in `target`, so it is transparent to the user.

BusyJay:
> make cap'n proto work with grpc-rs without some modification

gRPC is design for bytes instead of structs, so codec is not required to be protobuf.

> using protobuf-build rather than Jinkela

The differences I can see from protobuf-build and jinkela are that:

1. protobuf-build doesn't provide a clean abstraction, that even after switching to prost, it still requires rust-protobuf to work, this is the major difference between the two libraries.
2. protobuf-build requires protoc to work ATM.
3. no tests in protobuf-build, and that is part of reason why it breaks things occasionally.
4. proc-macro is used in jinkela to reduce files, but I don't think it's a must.

If protobuf-build needs to take as much as effort to support the same abstraction, I would rather choose jinkela, as it has done the work already. And in a optimistic view, we may fix the broken master today.

nrc:
> protobuf-build doesn't provide a clean abstraction, that even after switching to prost, it still requires rust-protobuf to work, this is the major difference between the two libraries.

It only requires the `Message` trait - protobuf-build uses `protobuf::Message` as its API, whereas Jinkela uses `GenericMessage`. I think we will need to add more functions to `GenericMessage` to support all the use cases in kvproto and TiKV, and then I do not see much benefit in `GenericMessage` over `protobuf::Message`. Jinkela approach is probably better in the long-term if we want to support very different codecs (e.g., cap'n proto), but I think that will require some more changes and is not urgent.

> protobuf-build requires protoc to work ATM.

true, but not urgent, I think

> no tests in protobuf-build, and that is part of reason why it breaks things occasionally.

I don't think that is the case, breakage has been because of subtle ways that Prost and rust-protobuf diverge (e.g., [behaviour of non-message, optional fields](https://github.com/tikv/protobuf-build/commit/a78fccaa75bdd3ce8455d5007af00593b224fbaa). Having said that, it would be much better to have some tests for protobuf-build.

As I see it, both approaches work now. Jinkela solves some more issues (e.g., protoc), whereas protobuf-build has had more to be tested against with more projects. In particular, protobuf-build has been used with kvproto and TiKV (although not in the case of supporting both Prost and rust-protobuf in TiKV).

IMO, we should stick with protobuf-build for now, but integrate some of the good ideas from Jinkela into it, that way we can get the design benefits of Jinkela without the risk of refixing issues that have already been resolved.

OTOH, if we think that the advantages of Jinkela outweigh the risks of switching now in one large move, then I am OK to do that.

> gRPC is design for bytes instead of structs, so codec is not required to be protobuf.

I know, but I think there are assumptions about protobuf in our implementation which must be adapted.

BusyJay:
> I think we will need to add more functions to GenericMessage to support all the use cases in kvproto and TiKV, and then I do not see much benefit in GenericMessage over protobuf::Message

I will show you today that `GenericMessage` is really generic enough. :smile:

nrc:
KvProto with Prost and rust-protobuf support: https://github.com/nrc/kvproto/commit/b10b5bd320b64b23cd8b8a716bc2ab7f2d2c99ec

nrc:
It occurs to me that we should be targeting the 0.4.3 branch, since that is what TiKV is using, not master

Hoverbear:
@BusyJay Please close this if it is stale

BusyJay:
This is replaced by #264 .

PR #261, opened 2019-07-15
cargo: update to fail 0.3

lucab:
This updates to latest fail 0.3. It also aligns feature naming to use `failpoints` everywhere, like the
underlying library.


Cargo.toml
```diff
@@ -18,7 +18,7 @@ members = ["proto"]
 [features]
 default = []
 # Enable failpoints
-failpoint = ["fail"]
+failpoints = ["fail", "fail/failpoints"]
```
BusyJay: I think `["fail/failpoints"]` is enough.

lucab: Indeed, I tried and that works. I just learned something new here, thanks for the suggestion!

lucab:
Amended with @BusyJay suggestion, PTAL.

PR #263, opened 2019-07-17
Cherry pick all 0.4.x

hicqu:
task list:
- [x] cherry pick all commits from 0.4.x to master
- [x] fix all test cases


hicqu:
@nrc some commits are empty because the work has been done in both master and 0.4.x branch. Let's keep the original commit log so that we can trace it one by one.
Now all test cases are fixed. PTAL thank you!

BusyJay:
@overvenus Would you like to take a look?

src/raft.rs
```diff
@@ -2062,16 +2066,55 @@ impl<T: Storage> Raft<T> {
...
+            return;
+        }
         if m.index < self.raft_log.committed {
             debug!(
                 self.logger,
                 "got message with lower index than committed.";
                 "tag" => &self.tag,
             );
             let mut to_send = Message::default();
+            to_send.to = m.from;
```
overvenus: Dup with line 2119.

src/raft.rs
```diff
@@ -2115,7 +2158,12 @@ impl<T: Storage> Raft<T> {
     /// For a message, commit and send out heartbeat.
     pub fn handle_heartbeat(&mut self, mut m: Message) {
         self.raft_log.commit_to(m.commit);
+        if self.pending_request_snapshot != INVALID_INDEX {
+            self.send_request_snapshot();
+            return;
+        }
         let mut to_send = Message::default();
+        to_send.to = m.from;
```
overvenus: Same as above.

src/storage.rs
```diff
@@ -669,18 +689,23 @@ mod test {
...
-        for (i, (idx, wresult)) in tests.drain(..).enumerate() {
+        for (i, (idx, wresult, windex)) in tests.drain(..).enumerate() {
             let storage = MemStorage::new();
             storage.wl().entries = ents.clone();
             storage.wl().raft_state.hard_state.commit = idx;
             storage.wl().raft_state.hard_state.term = idx;
             storage.wl().raft_state.conf_state = conf_state.clone();
 
-            let result = storage.snapshot();
+            let result = storage.snapshot(windex);
```
overvenus: We should trigger unavilable if `wresult` is an error.

tests/integration_cases/test_raft.rs
```diff
@@ -4364,3 +4291,284 @@ fn test_conf_change_check_before_campaign() {
     }
     assert_eq!(nt.peers[&1].state, StateRole::Candidate);
 }
+
+fn prepare_request_snapshot() -> (Network, Snapshot) {
+    /************************
```
overvenus: Please uncomment or clean up.

src/storage.rs
hicqu: Could you explain it with more details? I thinks it's already what we want.

tests/integration_cases/test_raft.rs
hicqu: I guess you review the code before the `fix test` commit.

overvenus: Oh, I missed it.

BusyJay:
This is blocked by #264.

src/raft.rs
```diff
@@ -594,14 +595,14 @@ impl<T: Storage> Raft<T> {
...
         let mut is_batched = false;
         for msg in &mut self.msgs {
-            if msg.msg_type() == MessageType::MsgAppend && msg.to == to {
+            if msg.get_msg_type() == MessageType::MsgAppend && msg.to == to {
                 if !ents.is_empty() {
                     if !util::is_continuous_ents(msg, ents) {
                         return is_batched;
                     }
-                    let mut batched_entries = msg.take_entries();
+                    let mut batched_entries: Vec<_> = msg.take_entries().into();
```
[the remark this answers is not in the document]
BusyJay: `take_entries()` return a `RepeatedField`.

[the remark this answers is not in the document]
overvenus: `RepeatedField` has `push` too, so I think the conversion is unnecessary.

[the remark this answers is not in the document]
BusyJay: The conversion is to support switching prost and rust-protobuf without modification of code.

PR #262, opened 2019-07-16
Support both Prost and rust-protobuf

nrc:
This is an alternative to https://github.com/pingcap/raft-rs/pull/260. It does less, hopefully the minimum required to get the 0.4.x branch to support both prost and rust-protobuf. It effectively merges the 0.4.x and prost branches. I don't merge to master since TiKV (and thus kvproto) are using the 0.4.x branch, so this is what is needed to permit development on master of kvproto.

PR #264, opened 2019-07-18
support protobuf build by default

BusyJay:
And support prost optionally

This is a derive work of #262 and #260.


Hoverbear:
@BusyJay Seems there are some clippy errors. :(

nrc:
Could you summarize how this different from #262 and #260 please? Why does this block #263 rather than #263 blocking this?

BusyJay:
This PR uses protobuf-build to generate codes just as #262 does, but the generated codes are put inside target directory instead which is the same as #260.

BusyJay:
Because CI in #260 is broken due to the changes to grpcio-compiler is out of sync to protobuf-build. This PR updates protobuf-build to a compatible version.

hicqu:
@BusyJay are you fixing the CI?

BusyJay:
Yes.

.travis.yml
```diff
@@ -31,7 +31,10 @@ install:
 script:
   - if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo fmt -- --check; fi
   - if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo clippy -- -D clippy::all; fi
+  - if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo clippy --no-default-features --features prost-codec -- -D clippy::all; fi
   - cargo test --all -- --nocapture
+  # There is a bug in protobuf-build that cached size is not supported yet.
```
overvenus: Is the bug still here?

BusyJay: Yes. It can be fixed later. Making rust-protobuf back is with highest priority.

PR #265, opened 2019-07-22
*: break circle dependency

BusyJay:
Cargo is not good at dealing with features with circle dependency, see
rust-lang/cargo#4866. In the past, raft depends on harness due to tests,
and harness depends on raft as it's only a wrapper of raft.  This PR
breaks the circle dependencies by moving all the tests from raft to
harness, so only harness needs to depend on raft.


hicqu:
So unit cases use `default_logger` but integration cases use `testing_logger` instead? To avoid block more works I think it's ok for now, but I think it's better to unify them in future.

overvenus:
(Off-topic, seem tikv has the same issue, tikv(`--test`) depends on test_raftstore, and test_raftstore depends on tikv.)

PR #268, opened 2019-08-05
Use protobuf-build builder API

nrc:
This also updates protobuf build. Other than bug fixes, there is no difference to the generated code.

PTAL @BusyJay @Hoverbear


nrc:
I'm a bit perplexed by the test failures - the tests pass locally, I've cleared the travis cache, and these look like legit errors and I don't see an obvious way these could be due to changes to protobuf generation.

BusyJay:
@hicqu PTAL

nrc:
CI is fixed after a rebase.

PTAL @Hoverbear @BusyJay

PR #269, opened 2019-08-06
uninitialized peers should become learners when restore some snapshots

hicqu:


src/raft.rs
```diff
@@ -2257,14 +2257,18 @@ impl<T: Storage> Raft<T> {
             "tag" => &self.tag,
         );
 
+        // Restore progress set and the learner flag.
         let next_idx = self.raft_log.last_index() + 1;
         let mut prs =
             ProgressSet::restore_snapmeta(meta, next_idx, self.max_inflight, &self.logger);
         prs.get_mut(self.id).unwrap().matched = next_idx - 1;
-        if self.is_learner && prs.configuration().voters().contains(&self.id) {
+        if prs.configuration().learners().contains(&self.id) {
```
nolouch: So we can change from `vote` to `learner` if the apply an old or wrong snapshot? What situation does peer think itself is learner but the `voters` in PorgressSet contains it?

hicqu: It's already checked at https://github.com/pingcap/raft-rs/pull/269/files#diff-18f3dab6275bbc16b4a102e73378c08bR2235.

hicqu: > What situation does peer think itself is learner but the voters in PorgressSet contains it?
The peer could be promoted in the raft group although it doesn't know.

src/raft.rs
```diff
@@ -2257,14 +2257,18 @@ impl<T: Storage> Raft<T> {
...
+            self.is_learner = true;
+        } else if prs.configuration().voters().contains(&self.id) {
             self.is_learner = false;
         }
```
overvenus: What if it is neither in voter nor learner?

hicqu: So the peer will treat itself as a voter. It's ok because the leader knows it's removed.

overvenus: What if other peers do not know? Will it lead to split-brain?

hicqu: The peer is not in the configuration means it's removed, which must be commited by a majority of the cluster.

PR #270, opened 2019-08-08
support to specify a logger when creating RawNode

hicqu:
As the title described.
And some `with_logger` APIs removed because they are not necessary now.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -241,9 +241,12 @@ impl<T: Storage> Raft<T> {
     /// Creates a new raft for use on the node.
     #[allow(clippy::new_ret_no_self)]
     pub fn new(c: &Config, store: T) -> Result<Raft<T>> {
-        let logger = default_logger().new(o!(
-            "id" => c.id,
-        ));
+        Self::new_with_logger(c, store, &default_logger())
+    }
+
+    /// Creates a new raft for use on the node.
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new_with_logger(c: &Config, store: T, logger: &Logger) -> Result<Raft<T>> {
         c.validate()?;
         let raft_state = store.initial_state()?;
         let conf_state = &raft_state.conf_state;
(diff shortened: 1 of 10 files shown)


src/raft.rs
```diff
@@ -253,10 +256,10 @@ impl<T: Storage> Raft<T> {
         let mut r = Raft {
             id: c.id,
             read_states: Default::default(),
-            raft_log: RaftLog::new(store, c.tag.clone()),
+            raft_log: RaftLog::new_with_logger(store, c.tag.clone(), logger),
             max_inflight: c.max_inflight_msgs,
             max_msg_size: c.max_size_per_msg,
-            prs: Some(ProgressSet::with_capacity(peers.len(), learners.len())),
+            prs: Some(ProgressSet::new_with_logger(logger)),
```
nrc: Seems like we lose some information here

src/raft.rs
```diff
@@ -241,9 +241,12 @@ impl<T: Storage> Raft<T> {
...
     pub fn new(c: &Config, store: T) -> Result<Raft<T>> {
-        let logger = default_logger().new(o!(
-            "id" => c.id,
-        ));
+        Self::new_with_logger(c, store, &default_logger())
+    }
+
+    /// Creates a new raft for use on the node.
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new_with_logger(c: &Config, store: T, logger: &Logger) -> Result<Raft<T>> {
```
BusyJay: Perhaps `with_logger` is enough?

src/raft.rs
BusyJay: How about requiring logger when initialing it with a capacity?

hicqu:
@nrc currently if we want to give `Raft` a logger, we can only use `with_logger` API. So if there are some logs in `Raft::new`, they could be print into the default logger, which is not expected.

hicqu:
PTAL @BusyJay @nrc thanks!

nrc:
> @nrc currently if we want to give Raft a logger, we can only use with_logger API. So if there are some logs in Raft::new, they could be print into the default logger, which is not expected.

Ah, that makes perfect sense, thanks!

PR #271, opened 2019-08-09
Correct the instructions of regenerating protobufs

Fullstop000:

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -5,18 +5,4 @@
 
 This crate contains the protobuf structs used by raft.
 
-## Regenerate Protobufs
-
-If the protobuf file `eraftpb.proto` changed,
-run this command to regenerate `eraftpb.rs`:
-
-```bash
-cargo build
-```
-
-or using Cargo watch:
-
-```bash
-cargo watch -s "cargo check"
-```
-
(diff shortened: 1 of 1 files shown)


Fullstop000:
Maybe just remove the `Regenerate Protobufs` section ? 

And one more thing is that the actual `eraftpb.rs` file is a little hard to located. The `eraftpb.rs` will be created under `target/debug/build/raft-proto-***/out` but there are usually a lot of `raft-proto-***` so that I don't know which is the newest being generated.

nrc:
> Maybe just remove the Regenerate Protobufs section ?

sgtm

> And one more thing is that the actual eraftpb.rs file is a little hard to locate

The easiest way is to run `cargo clean` before you build. There will then only be two raft-proto directories and only one will have an `out` sub-directory.

Fullstop000:
Using `cargo clean` before re-building could be frustrating because it takes time re-compiling without caches :( .

I tried `cargo clean -p` but it only removes things in `deps`.

Fullstop000:
> And one more thing is that the actual eraftpb.rs file is a little hard to locate

For anyone who has the same problem, I find that RLS is useful to deal with this.

PR #273, opened 2019-08-15
Use hex formatting for bytes

nrc:
This resurrects the formatting of bytes as hex using a fork of rust-protobuf. That was removed recently when we updated rust-protobuf.

PTAL @BusyJay @Hoverbear

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ prost-codec = ["raft-proto/prost-codec"]
 # Make sure to synchronize updates with Harness.
 [dependencies]
 log = ">0.2"
-protobuf = "2"
+protobuf = { git = "https://github.com/nrc/rust-protobuf", branch = "v2.8" }
 slog = "2.2"
 quick-error = "1.2.2"
 raft-proto = { path = "proto", version = "0.6.0-alpha", default-features = false }
@@ -56,3 +56,7 @@ path = "examples/single_mem_node/main.rs"
 [[example]]
 name = "five_mem_node"
 path = "examples/five_mem_node/main.rs"
+
+[replace]
+"protobuf-codegen:2.8.0" = { git = "https://github.com/nrc/rust-protobuf", branch = "v2.8" }
+"protobuf:2.8.0" = { git = "https://github.com/nrc/rust-protobuf", branch = "v2.8" }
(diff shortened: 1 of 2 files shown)


BusyJay:
I don't think you need to update this repository to get the functionality back. The original patch doesn't modify raft-rs either.

nrc:
> The original patch doesn't modify raft-rs either.

The original patch had impls for Raft in kvproto. We need to have the impls for Raft somewhere and if we don't make this update, then we need replace/patch wherever we want to use Raft proto and kvprotos together (i.e., at the moment in TiKV and kvproto). So, there is a trade-off having other Raft users use mainline rust-proto vs nicer Cargo.tomls for TiKV and kvproto. @BusyJay which do you prefer?

My adaption of the original patch has all the impls in the rust-proto fork, rather than some there and some in kvproto (I reasoned it was better to have all the changes in one place if we are already going to have a fork), so it doesn't make sense to have impls in kvproto.

BusyJay:
I think it's a bad idea to make it follow the format what TiKV prefers instead of a standard one. The whole point of make raft a standalone library is to make it generic enough for other usage, so I would prefer not touching the raft crate and do all the magic somewhere else. For example, how about still let raft crate impls special trait, but only enable it in kvproto (using some flag etc.)?

Cargo.toml
```diff
@@ -25,7 +25,7 @@ prost-codec = ["raft-proto/prost-codec"]
 # Make sure to synchronize updates with Harness.
 [dependencies]
 log = ">0.2"
-protobuf = "2"
+protobuf = { git = "https://github.com/nrc/rust-protobuf", branch = "v2.8" }
```
Hoverbear: This unfortunately renders Raft unpublishable, if we merge this it will require a future cleanup PR.

Cargo.toml
```diff
@@ -56,3 +56,7 @@ path = "examples/single_mem_node/main.rs"
 [[example]]
 name = "five_mem_node"
 path = "examples/five_mem_node/main.rs"
+
+[replace]
```
Hoverbear: Please use `[patch]` in all cases as `[replace]` doesn't work with `cargo vendor` :(

nrc:
OK, let's close this PR and use crate replacement in KvProto and TiKV.

PR #278, opened 2019-08-22
Update some deps

nrc:
To reduce the number of duplicated deps in TiKV

PTAL @Hoverbear @hicqu

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ prost-codec = ["raft-proto/prost-codec"]
 
 # Make sure to synchronize updates with Harness.
 [dependencies]
-log = ">0.2"
+log = "0.4"
 protobuf = "2"
 slog = "2.2"
 quick-error = "1.2.2"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -24,7 +24,7 @@ prost-codec = ["raft-proto/prost-codec"]
 
 # Make sure to synchronize updates with Harness.
 [dependencies]
-log = ">0.2"
+log = "0.4"
```
BusyJay: I think `>0.2` includes `0.4`?

Hoverbear: Perhaps it is necessary for `slog-stdlog`?

src/lib.rs
```diff
@@ -374,6 +374,8 @@ before taking old, removed peers offline.
 #![recursion_limit = "128"]
 // This is necessary to support prost and rust-protobuf at the same time.
 #![allow(clippy::identity_conversion)]
+// This lint recommends some bad choices sometimes.
```
Hoverbear: Is there an issue to reference here?

nrc: split off into https://github.com/pingcap/raft-rs/pull/282

nrc:
PTAL @hicqu @brson (there are no material changes, just split out some of it into https://github.com/pingcap/raft-rs/pull/282)

src/lib.rs
Hoverbear: They're still part of this PR so #282 blocks this :(

nrc: I put them in a separate commit so it can still be reviewed. Just needs a rebase now.

PR #279, opened 2019-08-28
Add 'delay' into Network

Fullstop000:
The feature could be used later in the benchmark.

### DIFF
--- a/harness/src/network.rs
+++ b/harness/src/network.rs
@@ -57,8 +57,8 @@ pub struct Network {
     pub storage: HashMap<u64, MemStorage>,
     /// Drop messages from `from` to `to` at a rate of `f64`.
     dropm: HashMap<Connection, f64>,
-    /// Delay sending messages from `from` to `to` at a rate of `f64` by blocking `u64` microseconds.
-    delaym: HashMap<Connection, (f64, u64)>,
+    /// Delay sending messages from `from` to `to` at a rate of `f64` by blocking `Duration` time.
+    delaym: HashMap<Connection, (f64, Duration)>,
     /// Drop messages of type `MessageType`.
     ignorem: HashMap<MessageType, bool>,
 }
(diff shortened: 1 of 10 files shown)


harness/src/network.rs
```diff
@@ -57,6 +57,8 @@ pub struct Network {
     pub storage: HashMap<u64, MemStorage>,
     /// Drop messages from `from` to `to` at a rate of `f64`.
     dropm: HashMap<Connection, f64>,
+    /// Delay sending messages from `from` to `to` at a rate of `f64` by blocking `u64` microseconds.
+    delaym: HashMap<Connection, (f64, u64)>,
```
Hoverbear: Do you think it's worth using https://doc.rust-lang.org/std/time/struct.Duration.html here?

Hoverbear:
Oh my gosh this is such a great idea @Fullstop000 :)

harness/src/network.rs
Fullstop000: It should be a Duration. Thanks!

Fullstop000:
I'm thinking that current designing is single-thread based which might be unfriendly to benchmarks. For example, we have a raft cluster with nodes A, B, and C where A is the leader and B&C is a follower. If a msg from A to B is delayed, the msgs from A to C should be unaffected.

Hoverbear:
@Fullstop000 I think that's a very valid concern. Do you think we could find a design that would permit that?

Fullstop000:
@Hoverbear I'll try to enhance `Network` to achieve that if it is truly needed in the benchmark. The current design is suitable for most cases.

harness/src/network.rs
```diff
@@ -57,6 +57,8 @@ pub struct Network {
     pub storage: HashMap<u64, MemStorage>,
     /// Drop messages from `from` to `to` at a rate of `f64`.
     dropm: HashMap<Connection, f64>,
+    /// Delay sending messages from `from` to `to` at a rate of `f64` by blocking `Duration` time.
+    delaym: HashMap<Connection, (f64, Duration)>,
```
BusyJay: This is a little confusing. So it does not just delay the messages in the connection, it also delays all the pending messages that may dispatch in other connection?

Fullstop000: That's the same concern I described above that connections shouldn't affect each other. But our `Network` seems to be single-thread based now but works fine in most functional test cases.

If I want to add some benches such as benchmarking proposing, this implementation is unsuitable. The design might be reconsidered and could be a lot of work.

Fullstop000:
Close this because the feature is not suitable for current `Network` implementation. I'll try to implement a new design out of `Network` ( or maybe just enhance it ) for benchmarking.

PR #284, opened 2019-09-04
Update criterion to 0.3 (customized version)

YangKeao:
Update criterion from version 0.2 to 0.3 which supports custom measurements.

And a customized version of criterion is used now. Because this customized version will delete outliers (which official criterion will never not) and quantile regression was used to get a more accurate statistic result.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ slog-envlogger = "2.1.0"
 
 [dev-dependencies]
 criterion = "0.3"
+criterion-cpu-time = { git = "https://github.com/YangKeao/criterion-cpu-time.git", branch = "master" }
 regex = "1.1"
 slog-async = "2.3.0"
 slog-term = "2.4.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ slog-stdlog = "3.0.2"
 slog-envlogger = "2.1.0"
 
 [dev-dependencies]
-criterion = ">0.2.4"
+criterion = "0.3"
 regex = "1.1"
 slog-async = "2.3.0"
 slog-term = "2.4.0"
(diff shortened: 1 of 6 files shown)


brson:
I appreciate the work you've done to improve criterion @YangKeao.

Maybe I shouldn't say it since I'm not going to fight merging this, but I am not enthusiastic about PingCAP using a criterion fork. That PingCAP uses so many forks seems like a bad code smell. This isn't a critique specifically of this fork, but I feel like I should voice this concern that has been creeping for a while.

YangKeao:
@brson This fork of criterion deletes all outliers from result. However the guide of criterion said we should pay attention to outliers but not delete them. So this fork of criterion maybe will never be merged with official one.

Your concerns are reasonable. To be honest, several days ago our fork of log v0.3 has compile errors while `Cargo.lock` locks the version of it and hides this error for weeks.

These `git` dependencies are unstable and hard to manage. But some forks will never be merged back with upstream. How should we treat these forks? Should we directly publish another crate (with a different name) ?

BusyJay:
Is the changes necessary? Can it be solved by shell script?

YangKeao:
@BusyJay If we handle raw benchmark output by a shell/python script, we also need to modify criterion (in another way). Because the format of raw data output (which contains time costed by every samples) of criterion is not stable.

Maybe after when criterion supports customizing analysis/report, we can use official criterion with custom analysis/report plugin. But these features are still in design and will not be released in near future.

BusyJay:
Can this meet your need? https://bheisler.github.io/criterion.rs/book/user_guide/csv_output.html

I don't think this will be merged as it introduces git dependency which can't be published to crates.io.

YangKeao:
@BusyJay Great. This document has changed since last time I saw it. 

> The format of this file is expected to remain stable between different versions of Criterion.rs, so this file is suitable for external tools to depend on.

As it guarantees the stability of format of csv file, I can use it without hesitation! :smile:

BusyJay:
Can this be closed?

PR #286, opened 2019-09-09
Update some deps

nrc:
This avoids a security vuln in memoffset and moves us to the same version of protobuf-build as the rest of our ecosystem, which reduces deps in TiKV. I also downgrade rand to 0.6.5, every other Crate we use uses rand 0.6, so this lets us avoid a lot of duplicate deps in TiKV. I don't see any reason to rush into using rand 0.7.

PTAL @Hoverbear @hicqu

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,12 +29,17 @@ protobuf = "2"
 slog = "2.2"
 quick-error = "1.2.2"
 raft-proto = { path = "proto", version = "0.6.0-alpha", default-features = false }
-rand = "0.7.0"
+# Stick with 0.6 (rather than 0.7) for now, since most of the ecosystem has not
+# upgraded yet and this saves duplicate deps.
+rand = "0.6.5"
 hashbrown = "0.5"
 fail = { version = "0.3", optional = true }
 getset = "0.0.7"
 slog-stdlog = "4"
 slog-envlogger = "2.1.0"
+# We don't actually use this crate, but by pinning to 0.5.1 it forces Cargo to
+# avoid 0.5.2 and thus pulling in a transitive dep with a security vulnerability.
+term = "=0.5.1"
 
 [dev-dependencies]
 criterion = ">0.2.4"
--- a/harness/Cargo.toml
+++ b/harness/Cargo.toml
@@ -19,7 +19,7 @@ prost-codec = ["raft/prost-codec"]
 # Make sure to synchronize updates with Raft.
 [dependencies]
 raft = { path = "..", default-features = false }
-rand = "0.7.0"
+rand = "0.6.5"
 slog = "2.2"
 slog-term = "2.4.0"
 slog-envlogger = "2.1.0"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -29,12 +29,15 @@ protobuf = "2"
...
 raft-proto = { path = "proto", version = "0.6.0-alpha", default-features = false }
-rand = "0.7.0"
+rand = "0.6.5"
 hashbrown = "0.5"
 fail = { version = "0.3", optional = true }
 getset = "0.0.7"
 slog-stdlog = "4"
 slog-envlogger = "2.1.0"
+# We don't actually use this crate, but by pinning to 0.5.1 it forces Cargo to
+# avoid 0.5.2 and thus pulling in a transitive dep with a security vulnerability.
```
Hoverbear: Can you note the cause of `rand`?

nrc: done!

zhangjinpeng1987:
@BusyJay PTAL

BusyJay:
What's the changes of protobuf-build?

Cargo.toml
```diff
@@ -29,12 +29,17 @@ protobuf = "2"
...
+# Stick with 0.6 (rather than 0.7) for now, since most of the ecosystem has not
+# upgraded yet and this saves duplicate deps.
+rand = "0.6.5"
 hashbrown = "0.5"
 fail = { version = "0.3", optional = true }
 getset = "0.0.7"
 slog-stdlog = "4"
 slog-envlogger = "2.1.0"
+# We don't actually use this crate, but by pinning to 0.5.1 it forces Cargo to
+# avoid 0.5.2 and thus pulling in a transitive dep with a security vulnerability.
```
BusyJay: If we don't use this crate, why pinning it here? Doesn't it introduce extra dependency?

nrc: No. It is a transitive dependency already so it is always depended on. If we don't pin it, then Cargo will select the more recent 0.5.2, and that has a security vulnerability. Note that in TiKV we stay on 0.5.1 because of the lock file, so its not an issue there.

nrc:
> What's the changes of protobuf-build?

We already use the more up to date version in kvproto and other repos. The changes between versions are that some generated methods return `&[T]` rather than `&Vec<T>` (closer to protobuf) and that the way we escape field names with "type" in the name changed slightly, again making the Prost wrappers closer to protobuf. Neither change is meaningful to Raft, but it means fewer deps in TiKV's dependency graph.

Cargo.toml
BusyJay: I think it's a bad idea to solve it in raft library. It should be the application's responsibility to choose a good version. Besides, it also prevent downstream from choosing a good version like 0.5.3 or 0.4.0.

BusyJay:
Get it. Thanks!

Cargo.toml
nrc: I agree it is not ideal, but it seems better than not doing it and getting a vulnerable dep (note that this is currently a problem, so this change only makes things better). I don't mind taking this out though if you prefer.

PR #288, opened 2019-09-12
*: clean up the usage of slog

BusyJay:
This commit brings 2 big changes to the API:

1. No more default logger.

    Users are required to give a logger to use the most of structs in this library. This makes the interface simple and reduces unnecessary dependencies.

2. No more tag scope value.

    Some APIs insert tag into KVs, some don't. It's confusing and easily lead to duplicate KV definitions. Users are required to insert tag into KVs before passing the logger to the library, and there is no more explicit configuration of tag.

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -47,19 +47,26 @@ use slog::{Drain, Logger};
 /// Currently, this is a terminal log. It ensures it is only initialized once to prevent clobbering.
 // This is `pub` so that testing and benching functions can use it.
 // `#[cfg(test)]` doesn't work for benching.
-#[doc(hidden)]
-pub fn testing_logger() -> &'static Logger {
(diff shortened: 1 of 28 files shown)


harness/src/lib.rs
```diff
@@ -47,19 +47,26 @@ use slog::{Drain, Logger};
 /// Currently, this is a terminal log. It ensures it is only initialized once to prevent clobbering.
 // This is `pub` so that testing and benching functions can use it.
 // `#[cfg(test)]` doesn't work for benching.
-#[doc(hidden)]
```
nrc: Why remove `doc(hidden)`? Seems like we don't want to make this public API

BusyJay: Because harness is not a public library, will never be published to crates.io, doc hidden is unnecessary.

PR #291, opened 2019-09-18
implement fast append rejection

### LINKED ISSUE #246: Fast log append rejection
When leader sends append to a follower, follower may reject the messages. However, it only sets its own last index, so there is no way leader can find what's the conflict they have. So leader has to decrease the progress one by one and detect whether it's accepted by the follower. The progress can be very slow and wasteful if there are too many pending logs in the old leader, which is just recovered from a long isolation, because the logs need be sent over and over again.

I suggest to add a hint for the term of last index. If the new leader know the term of the follower's last index, it can jump to the last log in the given term directly and start replicated. So there will be just one round of RPC to figure out the conflict.

For example,

OL: 1222333333333333333
NL: 124444444

The number is the term of the log. In the old implement, new leader has to decrease the progress one by one to walk through all the '4's. But if it can know that the term of OL's last index, it can jump to 2 directly, and starts replicate logs from 2.

hicqu:

Implement fast append rejection machinasm. Close #246 .

### DIFF
--- a/src/progress/mod.rs
+++ b/src/progress/mod.rs
@@ -28,8 +28,6 @@ use std::cmp;
 
 use self::inflights::Inflights;
 use crate::raft::INVALID_INDEX;
-use crate::raft_log::RaftLog;
-use crate::storage::Storage;
 pub mod inflights;
 pub mod progress_set;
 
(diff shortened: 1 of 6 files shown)


src/progress/mod.rs
```diff
@@ -195,7 +197,14 @@ impl Progress {
     /// Returns false if the given index comes from an out of order message.
     /// Otherwise it decreases the progress next index to min(rejected, last)
     /// and returns true.
-    pub fn maybe_decr_to(&mut self, rejected: u64, last: u64, request_snapshot: u64) -> bool {
+    pub fn maybe_decr_to<T: Storage>(
```
BusyJay: This looks ugly to me. `maybe_decr_to` intends to decrease replicate index to min(rejected, last). Can it be figured outside the function?

src/progress/mod.rs
```diff
@@ -227,6 +236,15 @@ impl Progress {
             if self.next_idx < 1 {
                 self.next_idx = 1;
             }
+            if last_term != 0 {
```
BusyJay: @overvenus PTAL

src/raft.rs
```diff
@@ -1427,11 +1427,24 @@ impl<T: Storage> Raft<T> {
                 self.logger,
                 "received msgAppend rejection";
                 "last index" => m.reject_hint,
+                "last term" => m.log_term,
                 "from" => m.from,
                 "index" => m.index,
             );
 
             if pr.maybe_decr_to(m.index, m.reject_hint, m.request_snapshot) {
+                // Before we implement the fast-resolve-conflict algorithm, `log_term` is 0.
```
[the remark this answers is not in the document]
hicqu: It's described in "In Search of an Understandable Consensus Algorithm", section 5.3. I will comment it in code.

BusyJay:
This is implemented in #367.

PR #293, opened 2019-09-23
`maybe_append` returns appended entries.

hicqu:

Just an engineering change, seperated from #202 .

### DIFF
--- a/src/raft_log.rs
+++ b/src/raft_log.rs
@@ -199,21 +199,22 @@ impl<T: Storage> RaftLog<T> {
         self.term(idx).map(|t| t == term).unwrap_or(false)
     }
 
+    // TODO: revoke pub when there is a better way to append without proposals.
     /// Returns None if the entries cannot be appended. Otherwise,
-    /// it returns Some(last index of new entries).
+    /// it returns Some(real appended entries).
     ///
     /// # Panics
     ///
     /// Panics if it finds a conflicting index.
-    pub fn maybe_append(
+    pub fn maybe_append<'a>(
         &mut self,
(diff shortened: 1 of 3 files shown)


src/raft_log.rs
```diff
@@ -503,6 +503,8 @@ impl<T: Storage> RaftLog<T> {
             snapshot_index = snapshot.get_metadata().index,
             snapshot_term = snapshot.get_metadata().term,
         );
+        // NOTE: here we don't change applied index because some applications apply snapshot
+        // asynchronous.
```
BusyJay: It's not about asynchronous, it's just about correctness. Apply index should be updated after data is restored, no matter if it's done asynchronously no not.

src/raft_log.rs
```diff
@@ -199,21 +199,22 @@ impl<T: Storage> RaftLog<T> {
...
+    pub fn maybe_append<'a>(
         &mut self,
         idx: u64,
         term: u64,
         committed: u64,
-        ents: &[Entry],
-    ) -> Option<u64> {
-        let last_new_index = idx + ents.len() as u64;
+        ents: &'a [Entry],
+    ) -> Option<&'a [Entry]> {
```
BusyJay: Why returning entries?

hicqu: Because in #202 , we need to update the current configuration after entries are appended. However the passed in `entries` could conflict with the peer' raft log, so we need to know which entries are really appended. For example:
the peer's current raft_log:
```
11122333
```
however it received:
```
11133333
```
So the returned `entries` will be `33333`.

PR #294, opened 2019-09-24
reduce the use of unwrap()

MaiCw4J:
This PR is try to use pattern matching to reduce unwrap() usage


src/raft.rs
```diff
@@ -516,42 +516,47 @@ impl<T: Storage> Raft<T> {
 
         m.set_msg_type(MessageType::MsgSnapshot);
         let snapshot_r = self.raft_log.snapshot(pr.pending_request_snapshot);
-        if let Err(e) = snapshot_r {
-            if e == Error::Store(StorageError::SnapshotTemporarilyUnavailable) {
+        match snapshot_r {
```
BusyJay: In my option, using match here makes the code ugly.

src/raft.rs
```diff
@@ -942,8 +944,7 @@ impl<T: Storage> Raft<T> {
         let change_start_index = self
             .pending_membership_change
             .as_ref()
-            .map(|v| Some(v.start_index))
```
BusyJay: I can't believe this was merged. 🤣

src/log_unstable.rs
```diff
@@ -99,25 +99,25 @@ impl Unstable {
...
     pub fn stable_snap_to(&mut self, idx: u64) {
-        if self.snapshot.is_none() {
-            return;
-        }
-        if idx == self.snapshot.as_ref().unwrap().get_metadata().index {
-            self.snapshot = None;
+        if let Some(snapshot_idx) = self
+            .snapshot
+            .as_ref()
+            .map(|snapshot| snapshot.get_metadata().index)
```
BusyJay: Why not compare them in `map`?

MaiCw4J: because I want it to look clearer and simpler. :)

src/raft.rs
MaiCw4J: > In my option, using match here makes the code ugly.

not bad for me, Does anyone else have any other opinions? It's necessary to rollback the code？

src/log_unstable.rs
Hoverbear: I'm not sure it's an improvement, but it's definitely not worse. :)

src/raft.rs
Hoverbear: I think this was me. :( Sorry.

src/log_unstable.rs
nrc: I find this combination of `if let` combinators and then `if` to be hard to read. I would prefer everything to use combinators or to use a variable for the combinator expression

src/raft.rs
nrc: I think both are OK - using `if let` and an early return feels a bit more clear, but a bit more C/Go/Swift style, whereas `match` seems more like idiomatic Rust, but a bit harder to read.

src/raft.rs
```diff
@@ -617,17 +621,16 @@ impl<T: Storage> Raft<T> {
         } else {
             let term = self.raft_log.term(pr.next_idx - 1);
             let ents = self.raft_log.entries(pr.next_idx, self.max_msg_size);
-            if term.is_err() || ents.is_err() {
-                // send snapshot if we failed to get term or entries.
-                if !self.prepare_send_snapshot(&mut m, pr, to) {
+            if let (Ok(term), Ok(mut ents)) = (term, ents) {
```
nrc: I prefer to use `match` rather than `if let ... else` since the former is fewer lines of code and the latter is just sugar for the former

src/util.rs
```diff
@@ -85,11 +85,12 @@ pub fn limit_size<T: PbMessage + Clone>(entries: &mut Vec<T>, max: Option<u64>)
 /// Check whether the entry is continuous to the message.
 /// i.e msg's next entry index should be equal to the first entries's index
 pub fn is_continuous_ents(msg: &Message, ents: &[Entry]) -> bool {
-    if !msg.entries.is_empty() && !ents.is_empty() {
-        let expected_next_idx = msg.entries.last().unwrap().index + 1;
-        return expected_next_idx == ents.first().unwrap().index;
+    if let (Some(msg_last), Some(ents_first)) = (msg.entries.last(), ents.first()) {
```
nrc: I'd prefer `match` here too.

src/log_unstable.rs
MaiCw4J: > I find this combination of `if let` combinators and then `if` to be hard to read. I would prefer everything to use combinators or to use a variable for the combinator expression

thanks for your help, But I still don't know how to modify this code. sorry ;(

MaiCw4J:
out of date close it

PR #296, opened 2019-09-25
Add benches for 'propose' and 'ready' of RawNode

Fullstop000:
Add benches for `RawNode::propose` and `RawNode::ready` as they are two of the most essential API to upper layers.

Part of https://github.com/pingcap/raft-rs/issues/109

@Hoverbear @hicqu 


benches/suites/raw_node.rs
```diff
@@ -20,3 +26,93 @@ pub fn bench_raw_node_new(c: &mut Criterion) {
...
+            node.raft.become_leader();
+            let mut total = Duration::from_nanos(0);
+            for _ in 0..iters {
+                // TODO: Maybe simulate more environments as input. For now, just preparing a raft node after stepping a proposal
+                node.propose(vec![], vec![]).expect("");
+                if node.has_ready() {
+                    let now = Instant::now();
+                    let ready = node.ready();
+                    total = total.add(now.elapsed());
+                    handle_ready(&mut node, ready);
```
BusyJay: Handle_ready usually depends on user usage, may not be proper counted in benchmark time.

Fullstop000: `handle_ready` is excluded from the benchmark as it's called after `total = total.add(now.elapsed())`.

Fullstop000:
I refactored the bench for `RawNode::propose` by `Benchmark` which makes configuring measurement for single benchmark become possible.

@Hoverbear @BusyJay

benches/suites/raw_node.rs
```diff
@@ -1,13 +1,18 @@
-use criterion::{Bencher, Criterion};
-use raft::{storage::MemStorage, Config, RawNode};
+use criterion::{Bencher, Benchmark, Criterion, Throughput};
```
Hoverbear: I note that Criterion 0.3.0 (the new version) has no `Benchmark` struct, so maybe the previous way was more forwards compatible? https://docs.rs/criterion/0.3.0/criterion/index.html

Fullstop000: Done. `BenchmarkGroup` works perfectly on this :)

benches/suites/raw_node.rs
bheisler: Criterion.rs author here! Obviously I don't know much about your code, so feel free to ignore me, but doing `Instant::now()` and `Instant::elapsed()` for every iteration will introduce some nasty measurement overhead and skew your measurements. If you can, setting up a batch of nodes and calling `ready` on all of them using `iter_batched` would give more accurate measurements.

Fullstop000: Thank you for your guidance! @bheisler
`iter_batched` seems very suitable for the situations I met.

BusyJay:
@hicqu @gengliqi PTAL

benches/suites/raw_node.rs
Hoverbear: Thanks for your amazing library, @bheisler!

benches/suites/raw_node.rs
```diff
@@ -20,3 +25,107 @@ pub fn bench_raw_node_new(c: &mut Criterion) {
...
+    group
+        // TODO: The proper measurement time could be affected by the system and machine.
+        .measurement_time(Duration::from_secs(60))
+        .bench_function("Default", |b: &mut Bencher| {
+            b.iter_batched(
+                || test_ready_raft_node(&logger),
+                |mut node| {
+                    let _ = node.ready();
+                },
+                // NOTICE: SmallInput accumulates (iters + 10 - 1) / 10 smaples per batch
```
Hoverbear:
```suggestion
                // NOTICE: SmallInput accumulates (iters + 10 - 1) / 10 samples per batch
```

benches/suites/raw_node.rs
```diff
@@ -20,3 +25,107 @@ pub fn bench_raw_node_new(c: &mut Criterion) {
...
+            );
+    }
+}
+
+pub fn bench_raw_node_new_ready(c: &mut Criterion) {
+    let logger = crate::default_logger();
+    let mut group = c.benchmark_group("RawNode::ready");
+    group
+        // TODO: The proper measurement time could be affected by the system and machine.
+        .measurement_time(Duration::from_secs(60))
```
Hoverbear: 60 seems like a long time, can you help me understand the reasoning?

benches/suites/raw_node.rs
```diff
@@ -20,3 +25,107 @@ pub fn bench_raw_node_new(c: &mut Criterion) {
...
+        .bench_function("Default", |b: &mut Bencher| {
+            b.iter_batched(
+                || test_ready_raft_node(&logger),
+                |mut node| {
+                    let _ = node.ready();
```
Hoverbear: From reading https://github.com/bheisler/criterion.rs/blob/master/book/src/faq.md#when-should-i-use-criterionblack_box I think we don't need to use it here. :)

benches/suites/raw_node.rs
Fullstop000: If the measurement time isn’t long enough, the benchmark result could be inaccurate. Maybe using a 'lighter' node(e.g. with fewer committed entries and messages) ?

benches/suites/raw_node.rs
Fullstop000: Do you mean the macro black_box? I haven’t used it in our benches.

Hoverbear: Yes I think the code is fine, I just was investigating for myself to make sure we didn't need to consider it. Good job. :)

benches/suites/raw_node.rs
Hoverbear: Ok, I think this is a good reason.

benches/suites/raw_node.rs
```diff
@@ -20,3 +25,107 @@ pub fn bench_raw_node_new(c: &mut Criterion) {
...
+        16 * KB,
+        128 * KB,
+        512 * KB,
+        KB * KB,
+    ];
+    let mut group = c.benchmark_group("RawNode::leader_propose");
+    for size in test_sets.drain(..) {
+        // Calculate measurement time in seconds according to the input size.
+        // The approximate time might not be the best but should work fine.
+        let mtime = if size < KB {
```
Hoverbear: Maybe it's better to just use `7` for everything?

Fullstop000: I'm worried that the total measurement time will be too long. We can still get enough analysing samples from the benchmark with tiny throughput by using shorter benching time.

hicqu:
Seems it can't work with criterion `0.2`. In my environment it only works after I update criterion to 0.3. Could you check it @Fullstop000 ? If you also meet the same problem, please update the depencency in Cargo.toml.

hicqu:
rest LGTM. Good job!

Fullstop000:
Can I update `Criterion` to `0.3.0` directly? I'm worried that there might be some dependency concerns involved.
@brson @hicqu

hicqu:
According to my test, it's ok to update criterion to 0.3.

Hoverbear:
Seems the CLA is stuck but you have already signed it. xD

PR #298, opened 2019-09-27
remove is_learner flag because it confuses readers when introduce joint consensus

hicqu:

Remove the `is_learner` flag from `Raft`, because after jonit consus is introduced, some peers could be learner in the old configuration, but voter in the new one. So we use `promotable` instead: if one given peer is voter in one of new or old configuration, it will be `promotable`.

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -1942,7 +1942,7 @@ fn test_non_promotable_voter_which_check_quorum() {
     // overwritten some internal states
     nt.peers.get_mut(&2).unwrap().mut_prs().remove(2).unwrap();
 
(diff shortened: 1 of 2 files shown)


harness/tests/integration_cases/test_raft.rs
```diff
@@ -3892,11 +3892,11 @@ fn test_add_voter_peer_promotes_self_sets_is_learner() -> Result<()> {
     let mut n1 = new_test_raft(1, vec![1], 10, 1, new_storage(), &l);
     // Node is already voter.
     n1.add_learner(1).ok();
-    assert_eq!(n1.is_learner, false);
+    assert_eq!(n1.promotable(), true);
```
BusyJay: Why not use `assert`

hicqu:
PTAL @BusyJay

PR #299, opened 2019-09-27
a little improvement about test network

hicqu:

Add more options about messages in `Network`, to make it works more fine with filters.


harness/src/network.rs
```diff
@@ -156,14 +158,19 @@ impl Network {
...
 
+    /// Read out all messages generated by peers in the `Network`.
+    ///
+    /// Note: messages are not filtered by any configured filters.
     pub fn read_messages(&mut self) -> Vec<Message> {
         self.peers
             .iter_mut()
             .flat_map(|(_peer, progress)| progress.read_messages())
-            .collect()
+            .collect::<Vec<_>>()
```
BusyJay: `Vec<>` is unnecessary.

PR #300, opened 2019-09-29
changes about proto to implement conf change on received

hicqu:

All changes are back compatible.


src/progress/progress_set.rs
```diff
@@ -660,6 +650,12 @@ impl ProgressSet {
             }
         }
     }
+
+    fn shrink_progress(&mut self) {
+        let mut id_list = self.voter_ids();
+        id_list.extend(self.learner_ids());
+        self.progress.retain(|id, _| id_list.contains(id));
```
BusyJay: Should it call `shrint_to_fit`?

PR #302, opened 2019-10-07
Optional logger

### LINKED ISSUE #301: Allow users to not pass a `slog` logger for `RawNode`
Currently, on the master branch, it is required that uses pass a slog logger to create a `RawNode`. This is a hindrance to people who want to get started but have not learned the slog crate.

Hoverbear:
This builds off #288 and fixes #301 .

Could you have a look @LucioFranco ?

Exposes a (default) feature called `default-logger` which is our normal `test_logger` just exposed publicly, it's used when `None` is passed as a logger to `RawNode` or `Raft`.

I am not sure about the exact details about what `default_logger` should look like,  but it seems to work fine if I use it as the logger in `five_mem_node`.


src/raw_node.rs
```diff
@@ -216,10 +216,14 @@ pub struct RawNode<T: Storage> {
     prev_hs: HardState,
 }
 
-impl<T: Storage> RawNode<T> {
+impl<'a, T: Storage> RawNode<T> {
```
LucioFranco: Do you want to tie this lifetime to the fn not the impl?

Hoverbear: Oh, yes, you're right! Thanks!

BusyJay:
I think just showing how to construct a default logger in example is enough. The purpose to require slog logger is to simplify dependencies and interfaces. This PR just defeats the goal.

hicqu:
LGTM.

Hoverbear:
@BusyJay The case @LucioFranco originally brought up was that they did not wish to use slog in their own program. I think that's a valid case and our current design prevents this. Choice in consensus library should dictate your choice in logger.

This allows downstreams like TiKV to opt out of this part of the dependency graph easily, while still allowing users who prefer libraries like tokio-tracing to use it without hacks by falling back to the standard log adapters.

I'd *prefer* to see this PR be an overall migration to tokio-tracing, but that's a bigger task in general.

src/raft.rs
```diff
@@ -235,9 +235,26 @@ pub fn vote_resp_msg_type(t: MessageType) -> MessageType {
 impl<T: Storage> Raft<T> {
     /// Creates a new raft for use on the node.
     #[allow(clippy::new_ret_no_self)]
-    pub fn new(c: &Config, store: T, logger: &Logger) -> Result<Raft<T>> {
+    pub fn new<'a>(c: &Config, store: T, logger: impl Into<Option<&'a Logger>>) -> Result<Raft<T>> {
         c.validate()?;
-        let logger = logger.new(o!("raft_id" => c.id));
+
+        let logger = logger.into();
+        #[cfg(feature = "default-logger")]
```
BusyJay: How about providing a new function `with_default_logger` instead?

Hoverbear:
Absolutely agree Jay. :)

Hoverbear:
@BusyJay I was tracking support right now, and it seems like log -> slog structured conversion may do mallocs right now according to https://github.com/rust-lang-nursery/log/issues/328#issue-438146093. It may be worth us using this interm solution for now.

src/raft.rs
Hoverbear: Yes! I think having `new(config, store).with_logger(x)` or `new(config, store).with_default_logger()` is good :)

LucioFranco:
@Hoverbear I think you're right. Sounds like we are very close to getting it in but its not ready yet. I would say this solution seems fine by me. I will also say for now I've written a simple `slog -> tracing` logger that works just fine and this should also work well for going from `slog -> log`.

src/raft.rs
BusyJay: Oh, I mean `with_default_logger` works like `with_capacity` as a constructor.

Hoverbear:
OK, great, I'll update this PR with https://github.com/pingcap/raft-rs/pull/302#discussion_r333126157 and we can merge this :)

src/raft.rs
Hoverbear: I don't think this is a better solution but since it doesn't really matter (We'll move off slog later anyways) letr's do it your way and get this merged.

PR #310, opened 2019-10-31
proto: introduce joint-consensus from etcd/raft

hicqu:
This PR introduces some proto changes about etcd/raft's joint-consensus implementation.
And, because their implementation is very different from ours, I removed all code about joint consensus in this PR, I will add those code back in following PRs.

The new `ConfStates`:
```
message ConfState {
    repeated uint64 voters = 1;
    repeated uint64 learners = 2;

    // The voters in the outgoing config. It's not empty means it's in joint consensus.
    repeated uint64 voters_outgoing = 3;
    // The nodes that will become learners when the outgoing config is removed.
    // These nodes are necessarily currently in nodes_joint (or they would have
    // been added to the incoming config right away).
    repeated uint64 learners_next = 4;
    // If set, the config is joint and Raft will automatically transition into
    // the final config (i.e. remove the outgoing config) when this is safe.
    bool auto_leave = 5;
}
```

For example, If a Raft wants to switch configuration from 

`V: [1, 2, 3], L: [4, 5]` to 

`V: [1, 2, 5], L: [3, 4]`, its conf state should be:

```
voters: vec[1, 2, 5]
learners: vec[4]
voters_outgoing: vec[1, 2, 3]
learners_next: vec[3]
```

@Hoverbear PTAL, especial the current proto file.


proto/proto/eraftpb.proto
```diff
@@ -94,31 +87,94 @@ message HardState {
...
+    // to use a custom payload (via the Context field).
+    Explicit = 2;
+}
+
 message ConfState {
-    repeated uint64 nodes = 1;
+    repeated uint64 voters = 1;
     repeated uint64 learners = 2;
+
+    // The voters in the outgoing config. It's not empty means it's in joint consensus.
```
Hoverbear:
```suggestion
    // The voters in the outgoing config. If not empty the node is in joint consensus.
```

proto/proto/eraftpb.proto
```diff
@@ -94,31 +87,94 @@ message HardState {
...
     ConfChangeType change_type = 2;
-    // Used in `AddNode`, `RemoveNode`, and `AddLearnerNode`.
     uint64 node_id = 3;
-
-    // Some extra information used in configuration change.
     bytes context = 4;
 
-    // For `BeginMembershipChange`, the target configuration is stored in the field.
-    ConfState configuration = 5;
+    uint64 id = 1;
```
Hoverbear: Why this move?

hicqu: Just to keep same order with etcd/raft. In fact the field `id` is only used by etcd, maybe we can remove it later.

src/raw_node.rs
```diff
@@ -330,10 +330,9 @@ impl<T: Storage> RawNode<T> {
     /// For a safe interface for these directly call `this.raft.begin_membership_change(entry)` or
     /// `this.raft.finalize_membership_change(entry)` respectively.
```
gengliqi: Maybe this annotation also should be removed

gengliqi: Maybe this annotation also should be removed

src/lib.rs
```diff
@@ -434,39 +435,6 @@ let mut node = RawNode::new(&mut config, store, &logger).unwrap();
 node.raft.become_candidate();
 node.raft.become_leader();
 
-// Call this on the leader, or send the command via a normal `MsgPropose`.
-node.raft.propose_membership_change((
```
BusyJay: Why remove the example?

src/progress/progress_set.rs
```diff
@@ -544,12 +444,7 @@ impl ProgressSet {
             }
             pr.recent_active = false;
         }
-        for (&_id, pr) in self.learners_mut() {
```
BusyJay: Why skip setting them to false?

src/lib.rs
hicqu: joint consensus is removed in this PR.

src/progress/progress_set.rs
hicqu: `recent_active` is only used for checking quorum on leader. So `quorum_recently_active` will never care about leaners. You can find the same logic in etcd:
```go
// QuorumActive returns true if the quorum is active from the view of the local
// raft state machine. Otherwise, it returns false.
func (p *ProgressTracker) QuorumActive() bool {
    votes := map[uint64]bool{}
    p.Visit(func(id uint64, pr *Progress) {
        if pr.IsLearner {
            return
        }
        votes[id] = pr.RecentActive
    })
    return p.Voters.VoteResult(votes) == quorum.VoteWon
}
```
And
```go
        if !r.prs.QuorumActive() {
            r.logger.Warningf("%x stepped down to follower since quorum is not active", r.id)
            r.becomeFollower(r.Term, None)
        }
        // Mark everyone (but ourselves) as inactive in preparation for the next
        // CheckQuorum.
        r.prs.Visit(func(id uint64, pr *tracker.Progress) {
            if id != r.id {
                pr.RecentActive = false
            }
        })
```

hicqu: Oh, I find the mistake. Learners still to be marked in the second sector.

src/lib.rs
BusyJay: But it still needs an example to do conf change, right?

hicqu: How about add it after the new API is introduced? It will be
```rust
raw_node.propose_conf_change(cc: &ConfChangeV2)
```
We don't use `ConfChangeV2` for now.

src/progress/progress_set.rs
hicqu: Fixed.

src/progress/progress_set.rs
```diff
@@ -544,12 +444,10 @@ impl ProgressSet {
...
+        for (_, pr) in self.learners_mut() {
```
BusyJay: Why not use `value_mut()`?

proto/proto/eraftpb.proto
```diff
@@ -94,31 +87,94 @@ message HardState {
     uint64 commit = 3;
 }
 
+enum ConfChangeTransition {
+    // Automatically use the simple protocol if possible, otherwise fall back
+    // to ConfChangeJointImplicit. Most applications will want to use this.
```
Hoverbear:
```suggestion
    // to `ConfChangeTransition::Implicit`. Most applications will want to use this.
```

PR #317, opened 2019-11-05
support joint consensus in etcd/raft style

hicqu:
This PR introduces joint consensus like etcd/raft style, and add test cases from etcd/raft and our old `test_membership_changes.rs`.

## Main changes:
* `ProgressSet` is almost refactored like `raft.ProgressTracker`.
* replace `add_voter`, `add_learner` with `apply_conf_change` for `Raft`.
* Add test cases in `test_membership_changes.rs`, based on our original work.

## Some differences from etcd/raft
* Demote voter to learner directly is not allowed. However, remove voter and then add it back as learner is ok.

## Prev work
Please review these PRs before review this:
* #318


BusyJay:
Joint consensus is now fully implemented.

PR #318, opened 2019-11-07
refactor progress set to etcd/raft style

hicqu:
It's extracted from #317 , only refactor `ProgressSet` in this PR, so it should be more easily to review.

### DIFF
--- a/src/progress/progress_set.rs
+++ b/src/progress/progress_set.rs
@@ -26,13 +26,14 @@
 // limitations under the License.
 
 use std::cell::RefCell;
+use std::{cmp, fmt, iter, slice, u64};
 
 use slog::Logger;
 
 use crate::eraftpb::{ConfState, SnapshotMetadata};
 use crate::errors::{Error, Result};
 use crate::progress::Progress;
-use crate::{DefaultHashBuilder, HashMap, HashSet};
+use crate::{HashMap, HashSet};
 
 /// Get the majority number of given nodes count.
 #[inline]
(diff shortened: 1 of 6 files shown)


src/progress/progress_set.rs
```diff
@@ -43,91 +44,98 @@ pub fn majority(total: usize) -> usize {
...
+    learners: Vec<u64>,
 
-    /// Create a new `Configuration` from a given `ConfState`.
-    pub fn from_conf_state(conf_state: &ConfState) -> Self {
-        Self {
-            voters: conf_state.voters.iter().cloned().collect(),
-            learners: conf_state.learners.iter().cloned().collect(),
-        }
+    // Demoted learners in joint consensus, because `learners` shouldn't intersect with `voters`.
+    learners_next: Vec<u64>,
```
Hoverbear: In `voters` we use `[Vec<u64>; 2]`, can we not do this with `learners`?

src/progress/progress_set.rs
```diff
@@ -149,114 +157,59 @@ pub enum CandidacyStatus {
...
+        let mut prs = HashMap::default();
+        for id in self.voters().chain(self.learners()) {
+            prs.insert(id, pr.clone());
         }
-
+        self.progress = prs;
         self.assert_progress_and_configuration_consistent();
     }
 
-    /// Returns the status of voters.
```
Hoverbear: 👍

src/progress/progress_set.rs
hicqu: Of course we can. However, etcd/raft uses `learners_next`, so I guess it's better to keep it.

src/progress/progress_set.rs
```diff
@@ -43,91 +44,98 @@ pub fn majority(total: usize) -> usize {
...
-                "There must be at least one voter.".into(),
-            ))
-        } else {
-            Ok(())
+    /// Create a new `Configuration` from a given `ConfState`.
+    pub fn from_conf_state(conf_state: &ConfState) -> Self {
+        Self::from(conf_state.clone())
+    }
+
+    // Test the configuration is valid or not. It's invalid when
```
Hoverbear: I suggest you use `///` for better IDE support

Fullstop000:
What's the progress of this? It looks like this has been paused for a long time

BusyJay:
Joint consensus is fully implemented.

PR #319, opened 2019-11-11
Update some dependencies

### LINKED ISSUE #335: Upgrading Prost
Hi, are there any plans to upgrade prost to it's 0.6.1 version? It includes a security fix as in addition to lots of other updates. https://github.com/danburkert/prost/releases/tag/v0.6.1 

I know there are many upstream places that it's used as well in raft's toolchain, if someone could tell me where they all are I can open PRs upgrading all the specific components. I just want to make sure I get them all in one pass.

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

* Update some dependencies
* Upgrade MSRV to 1.39.0 (required by `bytes` v0.5)

Close #335 (waiting for [grpcio-compiler/426](https://github.com/tikv/grpc-rs/pull/426)  and [protobuf-build/34](https://github.com/tikv/protobuf-build/pull/34) to release new version)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ rust:
   - stable
   - nightly
   # Officially the oldest compiler we support.
-  - 1.36.0
+  - 1.39.0
 matrix:
   include:
   - os: windows
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ fxhash = "0.2"
 fail = { version = "0.3", optional = true }
 getset = "0.0.9"
 protobuf = "2"
+quick-error = "1"
 raft-proto = { path = "proto", version = "0.6.0-alpha", default-features = false }
 rand = "0.7"
 slog = "2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,9 +4,9 @@ version = "0.6.0-alpha"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
 keywords = ["raft", "distributed-systems", "ha"]
-repository = "https://github.com/pingcap/raft-rs"
+repository = "https://github.com/tikv/raft-rs"
 readme = "README.md"
-homepage = "https://github.com/pingcap/raft-rs"
+homepage = "https://github.com/tikv/raft-rs"
 documentation = "https://docs.rs/raft"
 description = "The rust language implementation of Raft algorithm."
 categories = ["algorithms", "database-implementations"]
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ rust:
   - stable
   - nightly
   # Officially the oldest compiler we support.
-  - 1.36.0
+  - 1.39.0
```
BusyJay: Why change this?

koushiro: required by bytes v0.5

Fullstop000:
And why remove `quick-error`?

koushiro:
Why not use std::error::Error + std::fmt::Display? 
The description and cause methods have been deprecated.

BusyJay:
I can see the master of quick_error has resolved the problem. Let's wait for its next release.

And we may switch to other library depends on discussion in tikv/rfcs#38.

koushiro:
Okay, let's set aside the error handling part and wait for the discussion of RFC to have a result that most people agree with.
BTW, I prefer to use the library `thiserror` :smile:

Cargo.toml
```diff
@@ -25,32 +25,32 @@ default-logger = ["slog-stdlog", "slog-envlogger", "slog-term"]
...
-slog-envlogger = { version = "2.1.0", optional = true }
-slog-term = { version = "2.4.0", optional = true }
-quick-error = "1.2.2"
+quick-error = "1"
 raft-proto = { path = "proto", version = "0.6.0-alpha", default-features = false }
 rand = "0.7"
-fxhash = "0.2.1"
-fail = { version = "0.3", optional = true }
-getset = "0.0.7"
+slog = "2"
```
BusyJay: I suggest to keep all these versions as their original values. For example, it's different between "2" and "2.2". Libraries work with 2.2.0 don't have to work with 2.0.0. Changing it to 2 gives a false assumption that we have tested and guarantee compatible against 2.1 and 2.0.

proto/Cargo.toml
```diff
@@ -15,14 +15,12 @@ build = "build.rs"
 [features]
 default = ["protobuf-codec"]
 protobuf-codec = ["protobuf-build/protobuf-codec"]
-prost-codec = ["prost", "prost-derive", "bytes", "lazy_static", "protobuf-build/prost-codec"]
+prost-codec = ["protobuf-build/prost-codec", "prost", "lazy_static"]
```
BusyJay: Interesting. Isn't prost requires bytes dependency?

koushiro: prost requires bytes dependency, but raft does not explicitly use the methods in Buf and BufMut traits

nrc: Prost 0.6 fixed that requirement, it's only required with 0.5 unless you use Buf/BufMut, etc.

PR #327, opened 2019-12-09
reduce useless Vec allocate

MaiCw4J:
Signed-off-by: mak <610546069@qq.com>

in `ready()` when `committed_entries` is `None` will always return an empty `vec`, i think it's unnecessary, this PR is try to reduce useless Vec allocate in `ready()` function

### DIFF
--- a/src/raw_node.rs
+++ b/src/raw_node.rs
@@ -123,13 +123,10 @@ impl Ready {
         if !raft.msgs.is_empty() {
             mem::swap(&mut raft.msgs, &mut rd.messages);
         }
-        rd.committed_entries = Some(
-            (match since_idx {
-                None => raft.raft_log.next_entries(),
-                Some(idx) => raft.raft_log.next_entries_since(idx),
-            })
-            .unwrap_or_else(Vec::new),
-        );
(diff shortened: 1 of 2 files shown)


src/raw_node.rs
```diff
@@ -141,8 +138,8 @@ impl Ready {
             }
             rd.hs = Some(hs);
         }
-        if raft.raft_log.unstable.snapshot.is_some() {
-            rd.snapshot = raft.raft_log.unstable.snapshot.clone().unwrap();
+        if let Some(snapshot) = &raft.raft_log.unstable.snapshot {
+            rd.snapshot = snapshot.clone();
```
nrc: It seems that if the snapshot is `Some<[]>` then it will still be cloned where it doesn't need to be.

MaiCw4J:
> A Vec doesn't allocate until something is put into it, so empty Vecs are very cheap. I think we should only use `Option<Vec>` if there is a distinction between `None` and `Some<[]>`, which in this case there doesn't seem to be

I got it, thanks for your explanation.

MaiCw4J:
this PR is unnecessary

PR #330, opened 2019-12-27
add stale check bot

BusyJay:
Add a bot to close stale issues/PRs automatically.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,5 +1,5 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 60
+daysUntilStale: 180
 # Number of days of inactivity before a stale issue is closed
 daysUntilClose: 7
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,20 @@
...
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - Help Wanted
+  - Proposal
+  - Pinned
+  - Security
+  - Request for Comment
+# Label to use when marking an issue as stale
+staleLabel: wontfix
```
Hoverbear: `wontfix` seems to mean "We do not ever plan to fix this for any reason", perhaps `stale` is more appropriate? Since it's possible there was an issue we wanted to address but were not satisfied with the solutions.

tisonkun:
Closed as itself stale :D

PR #336, opened 2020-02-06
support to custom quorum function.

hicqu:

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -14,9 +14,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp;
 use std::collections::HashMap;
 use std::panic::{self, AssertUnwindSafe};
+use std::{cmp, mem};
 
 use harness::*;
 use protobuf::Message as PbMessage;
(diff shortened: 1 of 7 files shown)


BusyJay:
Where is the usage of custom quorum function? Besides, I think it's necessary to wrap the output into safe range. Custom quorum is expected to be safer.

hicqu:
> Custom quorum is expected to be safer.
The key is not only `quorum_fn(len(self.peers))` needs to be checked. Could we check `quorum_fn` for `range(1, 10)`? Or we can do the check every time when `quorum_fn` is called. Which do you prefer? Or just keep it unchecked but leave it in doc.

BusyJay:
Check it every time to avoid abuse of the function:
```
cmp::min(cmp::max(n / 2 + 1, quorum_fn(n)), n)
```

BusyJay:
Should check for commit index too.

BusyJay:
Any test case to ensure safety when unsafe function is provided?

src/progress/progress_set.rs
```diff
@@ -385,8 +385,15 @@ impl ProgressSet {
         });
         // Reverse sort.
         matched.sort_by(|a, b| b.cmp(a));
-        let offset = quorum_fn(matched.len()) - 1;
-        matched[offset]
+
+        let mut quorum = quorum_fn(matched.len());
+        if quorum_fn != crate::majority {
```
BusyJay: How about make this a global function to reduce duplicated code?

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4591,3 +4591,36 @@ fn test_request_snapshot_on_role_change() {
...
+        // Test quorum_fn in `has_quorum`.
+        for i in 0..2 {
+            raft.mut_prs().get_mut(i + 1).unwrap().recent_active = true;
+        }
+        assert!(!raft.mut_prs().quorum_recently_active(1, *f));
+
+        for i in 0..5 {
+            raft.mut_prs().get_mut(i + 1).unwrap().recent_active = true;
+        }
+        assert!(raft.mut_prs().quorum_recently_active(1, *f));
```
BusyJay: What about election and log commit?

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4591,3 +4591,78 @@ fn test_request_snapshot_on_role_change() {
...
+    fn unsafe_quorum_fn_1(_: usize) -> usize {
+        1
+    }
+
+    fn unsafe_quorum_fn_2(_: usize) -> usize {
+        100
+    }
+
+    let cases: Vec<(fn(usize) -> usize, usize)> =
+        vec![(unsafe_quorum_fn_1, 3), (unsafe_quorum_fn_2, 5)];
```
BusyJay: Seems miss a customize safe quorum case.

src/raft.rs
```diff
@@ -883,7 +887,7 @@ impl<T: Storage> Raft<T> {
             "term" => self.term
         );
         self.register_vote(self_id, acceptance);
-        if let CandidacyStatus::Elected = self.prs().candidacy_status(&self.votes) {
+        if let CandidacyStatus::Elected = self.prs().candidacy_status(&self.votes, self.quorum_fn) {
```
gengliqi: For the election, I think it's better to use `majority` as before because the majority is enough and has no effect on the case described in https://github.com/etcd-io/etcd/issues/11506.

hicqu: Suppose a Raft group in 2 data centers meets 1 data center failure. If the Raft group is recovered from the rest data center manually, there could be 2 leaders if the failed data center also recovers. So it's a disturbance to use 2 quorum algorithm for election and replication I guess. It's better to keep them as same.

gengliqi: 2 leader case is no problem. As long as the quorum algorithm for election and replication are both greater than the majority, the safety is guaranteed.
(This condition can be weaker according to the flexible paxos paper)

hicqu: This won't introduce inconsistent data, but could introduce unavailable (if clients connect to the partition in failure-recover data center).

PR #339, opened 2020-02-21
*: use unsafe cell to remove options

BusyJay:
So that `step_leader` is implemented more similar to etcd's. It helps a lot to avoid implement bugs or find missing patches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ matrix:
   - rust: nightly
     install:
       - rustup component add clippy-preview --toolchain nightly
-    before_script:
-      - cargo clippy --all -- -D clippy::all
+    # before_script:
+    #   - cargo clippy --all -- -D clippy::all
 
 
 script:
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -25,12 +25,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp;
-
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -19,8 +19,8 @@ matrix:
   - rust: nightly
     install:
       - rustup component add clippy-preview --toolchain nightly
-    before_script:
-      - cargo clippy --all -- -D clippy::all
+    # before_script:
```
hicqu: Why disble it?

BusyJay: It reports a lot of warnings that I'm going to solve in another PR.

src/raft.rs
```diff
@@ -111,7 +111,7 @@ pub struct Raft<T: Storage> {
     /// needs it to be included in a snapshot.
     pub pending_request_snapshot: u64,
 
-    prs: Option<ProgressSet>,
+    prs: UnsafeCell<ProgressSet>,
```
overvenus: How about RefCell? `unsafe` sounds scary.

BusyJay: `RefCell` is not zero cost. We already check the implement carefully that it can't be mutate at the same time.

overvenus: What about `Cell`?

BusyJay:
The PR may break alias rule in Rust, close it for now until a better approach is available.

PR #340, opened 2020-02-27
*: port coreos/etcd#9985

BusyJay:
The patch is to speed up log replication when a node is way behind than leader and logs are not compacted yet.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -560,13 +560,14 @@ impl<T: Storage> Raft<T> {
                 return false;
             }
             let term = self.raft_log.term(pr.next_idx - 1);
-            if term.is_err() || ents.is_err() {
-                // send snapshot if we failed to get term or entries.
-                if !self.prepare_send_snapshot(&mut m, pr, to) {
-                    return false;
+            match (term, ents) {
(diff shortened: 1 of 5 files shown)


src/raft.rs
```diff
@@ -1269,7 +1301,7 @@ impl<T: Storage> Raft<T> {
                 to_send.set_msg_type(MessageType::MsgReadIndexResp);
                 to_send.set_index(rs.index);
                 to_send.set_entries(req.take_entries());
-                *more_to_send = Some(to_send);
+                ctx.more_to_send.push(to_send);
```
hicqu: How about `more_to_send: HashMap<u64, Message>`? So we can send only 1 read index response to one peer.

hicqu:
rest LGTM. BTW could you send a PR to master? Let's think about how to fix test cases about initializing with configuration.

src/raft.rs
BusyJay: `HashMap` is slower than `Vec`, it may not be a good idea. On the other hand, the optimization assumes there is a queue in follower side, which may not be a valid case.

BusyJay:
I will port to master when tests issues are fixed.

PR #341, opened 2020-03-03
*: bump 0.4.4

BusyJay:
- Support requesting snapshot (https://github.com/pingcap/raft-rs/pull/243)
- Make flow control more aggressive (https://github.com/pingcap/raft-rs/pull/340)


BusyJay:
Test failure due to missing clippy on nightly rust. It has pass the check on stable rust. So I'm going to merge it.

PR #342, opened 2020-03-07
update link to raft paper

stoksc:
looking for a good first issue.. just fixed this for now, nothing I see that's not already being worked on. any suggestions?


Fullstop000:
@stoksc Pls sign DCO otherwise the CI is always failing.

PR #344, opened 2020-03-10
support update quorum function

BusyJay:
So we can ease the constraint after disaster.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -389,6 +389,20 @@ impl<T: Storage> Raft<T> {
         self.batch_append = batch_append;
     }
 
+    /// Uses a new quorum function.
+    pub fn set_quorum(&mut self, quorum_fn: fn(usize) -> usize) {
+        self.quorum_fn = quorum_fn;
+        match self.state {
+            StateRole::Leader => {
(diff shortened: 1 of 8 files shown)


src/raft.rs
```diff
@@ -389,6 +389,20 @@ impl<T: Storage> Raft<T> {
         self.batch_append = batch_append;
     }
 
+    /// Uses a new quorum function.
+    pub fn set_quorum(&mut self, quorum_fn: fn(usize) -> usize) {
```
hicqu: Could you also add `get_quorum`? So we can use the field in raftstore directly for `check_conf_change`.

src/raft.rs
```diff
@@ -389,6 +389,20 @@ impl<T: Storage> Raft<T> {
...
+        self.quorum_fn = quorum_fn;
+        match self.state {
+            StateRole::Leader => {
+                if self.maybe_commit() {
```
hicqu: Is it necessary to check pending reads?

BusyJay: It's an existing bug. Not only should check when quorum_fn is changed, but also should check when a voter is removed.

PR #346, opened 2020-03-11
advance commit index by MsgReadIndexResp

### LINKED ISSUE #343: Read index request should advance the commit index if possible
When a follower/learner requests a read index, it's possible that it has new enough logs but its commit index is smaller than this read index.
If there is no proposal, the follower/learner's commit index is advanced by the leader's heartbeat, which is too slow and increases the latency time.
So the read index request should advance the commit index if possible.

**Is your feature request related to a problem? Please describe.**
https://github.com/tikv/tikv/issues/7047
Actually, the issue above is caused by the leader will not broadcast the empty entry when there is only one voter. But this issue inspired us to do this optimization.

NingLin-P:
Signed-off-by: linning <linningde25@gmail.com>
Leader only handle `MsgReadIndex` when it has committed log entry in its term and the read index this leader returned always be its commit index , so log entries at the read index are always proposed and committed by this leader at its term, so if the follower's log entry match the read index and term on `MsgReadIndexResp` it can advance its commit index to the read index.
close #343

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -2453,20 +2453,18 @@ fn test_advance_commit_index_by_read_index_response() {
     tt.cut(1, 3);
     tt.cut(1, 4);
     tt.cut(1, 5);
-
     tt.send(vec![new_message(1, 1, MessageType::MsgPropose, 1)]);
     tt.send(vec![new_message(1, 1, MessageType::MsgPropose, 1)]);
 
-    // network recovery
     tt.recover();
-    // node 2' commit index will not be advanced by these msgs
-    tt.ignore_to(2, MessageType::MsgAppend);
-    tt.ignore_to(2, MessageType::MsgHeartbeat);
+    tt.cut(1, 2);
 
-    tt.peers.get_mut(&1).unwrap().raft_log.commit_to(4);
+    // commit entries for leader but not node 2
+    tt.send(vec![new_message(3, 1, MessageType::MsgReadIndex, 1)]);
     assert_eq!(tt.peers[&1].raft_log.committed, 4);
     assert_eq!(tt.peers[&2].raft_log.committed, 2);
 
+    tt.recover();
     tt.send(vec![new_message(2, 1, MessageType::MsgReadIndex, 1)]);
     assert_eq!(tt.peers[&2].raft_log.committed, 4);
 }
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -2465,6 +2465,8 @@ fn test_advance_commit_index_by_read_index_response() {
     assert_eq!(tt.peers[&2].raft_log.committed, 2);
 
     tt.recover();
+    // use LeaseBased so leader won't send MsgHeartbeat to advance node 2's commit index
+    tt.peers.get_mut(&1).unwrap().read_only.option = ReadOnlyOption::LeaseBased;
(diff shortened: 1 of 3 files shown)


src/raft.rs
```diff
@@ -1826,6 +1826,7 @@ impl<T: Storage> Raft<T> {
                     request_ctx: m.take_entries()[0].take_data(),
                 };
                 self.read_states.push(rs);
+                self.raft_log.maybe_commit(m.commit, m.term);
```
BusyJay: Any tests to cover it?

NingLin-P: Added, PATL.

NingLin-P: BTW `maybe_commit(m.commit, m.term)` was mistake, had changed to `maybe_commit(m.index, m.term)`

src/raft.rs
```diff
@@ -1826,6 +1826,7 @@ impl<T: Storage> Raft<T> {
                     request_ctx: m.take_entries()[0].take_data(),
                 };
                 self.read_states.push(rs);
+                self.raft_log.maybe_commit(m.index, m.term);
```
BusyJay: Term may be wrong.

NingLin-P: `index` and `term` in MsgReadIndexResp is the leader's commit index and ite current term, and log entry in the leader's commit index will always have the leader's current term, because the leader only handle MsgReadIndex after it has committed log entry in its term.

BusyJay: Get it. Can you also comment it in the file?

BusyJay: Remember to also send the patch to etcd.

NingLin-P: > Get it. Can you also comment it in the file?

Added.


> Remember to also send the patch to etcd.

Will do.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -2441,6 +2441,36 @@ fn test_read_only_for_new_leader() {
...
+    tt.send(vec![new_message(1, 1, MessageType::MsgPropose, 1)]);
+    tt.send(vec![new_message(1, 1, MessageType::MsgPropose, 1)]);
+
+    // network recovery
+    tt.recover();
+    // node 2' commit index will not be advanced by these msgs
+    tt.ignore_to(2, MessageType::MsgAppend);
+    tt.ignore_to(2, MessageType::MsgHeartbeat);
+
+    tt.peers.get_mut(&1).unwrap().raft_log.commit_to(4);
```
BusyJay: It's wrong to commit the log. You can send a `MsgReadIndex` on behalf of 3 to resume log replications.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -2441,6 +2441,34 @@ fn test_read_only_for_new_leader() {
...
+    tt.recover();
+    tt.cut(1, 2);
+
+    // commit entries for leader but not node 2
+    tt.send(vec![new_message(3, 1, MessageType::MsgReadIndex, 1)]);
+    assert_eq!(tt.peers[&1].raft_log.committed, 4);
+    assert_eq!(tt.peers[&2].raft_log.committed, 2);
+
+    tt.recover();
+    tt.send(vec![new_message(2, 1, MessageType::MsgReadIndex, 1)]);
```
BusyJay: Should make leader use lease-base read option, otherwise heartbeat is still sent out.

NingLin-P: Fixed

PR #351, opened 2020-03-24
Remove StatusRef

hhbti:
This effectively revert part of 065c37d653daf654b8ef466e8f1658411264e3eb.

Because it is not necessary after b86e9e00067e5dcce10385c65506a38817d22430.

Normal Status also uses borrow now.


hhbti:
@BusyJay Wait I need two approvals for this change? Can you help me find one more? Thanks! 😀

BusyJay:
@gengliqi PTAL

PR #353, opened 2020-03-25
*: allow index start from 0

BusyJay:
This PR removes the modifications of index during initialization, which will allow raft's log index goes from 0 again. This is important as many existing test cases rely on it, especially those related to flow control. Those cases pass on master in an accidental way, checking what are exact broken will be too much work. Besides upstream patches from etcd also rely on the same assumption, only forbid growing from 0 in raft-rs will make it hard to port patches.

Most of the changes are basically revert of index changes introduced from #196 and #214.

### DIFF
--- a/harness/tests/integration_cases/test_raw_node.rs
+++ b/harness/tests/integration_cases/test_raw_node.rs
@@ -57,7 +57,7 @@ fn new_raw_node(
     logger: &Logger,
 ) -> RawNode<MemStorage> {
     let config = new_test_config(id, election, heartbeat);
-    if storage.initial_state().unwrap().initialized() && !peers.is_empty() {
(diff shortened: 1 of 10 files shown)


gengliqi:
Seems some test cases failed.

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -57,11 +57,14 @@ fn new_raw_node(
     logger: &Logger,
 ) -> RawNode<MemStorage> {
     let config = new_test_config(id, election, heartbeat);
-    if storage.initial_state().unwrap().initialized() && peers.is_empty() {
+    if storage.initial_state().unwrap().initialized() && !peers.is_empty() {
```
hicqu: Why add `!`?

PR #356, opened 2020-04-02
implement committed entries pagination

Fullstop000:
Finish the remaining work in https://github.com/tikv/raft-rs/pull/161


Fullstop000:
@BusyJay @hicqu PTAL. Sorry for delaying this for sooooooo long :(.

src/raw_node.rs
```diff
@@ -134,11 +134,19 @@ impl Ready {
         if &ss != prev_ss {
             rd.ss = Some(ss);
         }
-        let hs = raft.hard_state();
+        let mut hs = raft.hard_state();
         if &hs != prev_hs {
             if hs.vote != prev_hs.vote || hs.term != prev_hs.term {
                 rd.must_sync = true;
             }
+            // If we hit a size limit when loading committed_entries, clamp
```
BusyJay, line 142: Seems etcd-io/etcd#10063 is also need to be picked.

src/config.rs
```diff
@@ -49,9 +49,14 @@ pub struct Config {
     /// Limit the max size of each append message. Smaller value lowers
     /// the raft recovery cost(initial probing and message lost during normal operation).
     /// On the other side, it might affect the throughput during normal replication.
-    /// Note: math.MaxUusize64 for unlimited, 0 for at most one entry per message.
+    /// Note: raft::NO_LIMIT for unlimited, 0 for at most one entry per message.
     pub max_size_per_msg: u64,
 
+    /// MaxCommittedSizePerReady limits the size of the committed entries which
+    /// can be applied.
+    /// If not set, this is same as `max_size_per_msg`
```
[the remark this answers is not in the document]
Fullstop000: fixed

Fullstop000:
@BusyJay @hicqu PTAL

src/raw_node.rs
```diff
@@ -408,25 +427,6 @@ impl<T: Storage> RawNode<T> {
...
-            // incorporated into the snapshot, even if rd.CommittedEntries is
-            // empty). Therefore we mark all committed entries as applied
-            // whether they were included in rd.HardState or not.
-            self.advance_apply(commit_idx);
-        }
-    }
-
-    /// Appends and commits the ready value.
-    #[inline]
-    pub fn advance_append(&mut self, rd: Ready) {
```
BusyJay, line 429: This is different from `advance`. Separating advancing commit and apply allow users to handle them separately without blocking each other.

src/config.rs
```diff
@@ -177,6 +183,12 @@ impl Config {
             )));
         }
 
+        // default MaxCommittedSizePerReady to MaxSizePerMsg because they were
+        // previously the same parameter.
+        if self.max_committed_size_per_ready == 0 {
+            self.max_committed_size_per_ready = self.max_size_per_msg
```
BusyJay, line 189: However, it may not be backward compatible. I suggest to just leave it to `NO_LIMIT` instead.

src/raft.rs
```diff
@@ -217,18 +217,19 @@ pub fn vote_resp_msg_type(t: MessageType) -> MessageType {
 impl<T: Storage> Raft<T> {
     /// Creates a new raft for use on the node.
     #[allow(clippy::new_ret_no_self)]
-    pub fn new(c: &Config, store: T, logger: &Logger) -> Result<Self> {
+    pub fn new(mut c: Config, store: T, logger: &Logger) -> Result<Self> {
```
BusyJay, line 220: Better not change the interface. I don't see any point to make it owned.

Fullstop000: Reverted. It's because `validate` need a `&mut self`

src/raw_node.rs
```diff
@@ -408,26 +419,21 @@ impl<T: Storage> RawNode<T> {
...
-            self.advance_apply(commit_idx);
+        // If entries were applied (or a snapshot), update our cursor for
+        // the next Ready. Note that if the current HardState contains a
+        // new Commit index, this does not mean that we're also applying
+        // all of the new entries due to commit pagination by size.
+        let index = rd.applied_cursor();
+        if index > 0 {
+            self.advance_apply(index)
         }
+        self.advance_append(rd);
```
BusyJay, line 430: Any reasons to change the order of `advance_append` and `advance_apply`?

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -484,3 +484,170 @@ fn test_set_priority() {
...
+    persisted_hard_state.set_term(1);
+    persisted_hard_state.set_vote(1);
+    persisted_hard_state.set_commit(10);
+    let s = IgnoreSizeHintMemStorage::default();
+    s.inner.wl().set_hardstate(persisted_hard_state);
+    let ents_count = 10;
+    let mut ents = Vec::with_capacity(ents_count);
+    let mut size = 0u64;
+    for i in 0..ents_count {
+        let e = new_entry(1, i as u64 + 1, Some("a"));
```
BusyJay, line 604: Why not `i +1 `?

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -484,3 +484,170 @@ fn test_set_priority() {
...
+        }
+        highest_applied = rd
+            .committed_entries
+            .as_ref()
+            .unwrap()
+            .last()
+            .unwrap()
+            .get_index();
+        raw_node.advance(rd);
+        let mut m = new_message(1, 1, MessageType::MsgHeartbeat, 0);
```
BusyJay, line 648: What's the purpose of the heartbeat?

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -484,3 +484,170 @@ fn test_set_priority() {
...
+// - Commit index gets bumped to 10
+// - the node persists the HardState, but crashes before applying the entries
+// - upon restart, the storage returns the same entries, but `slice` takes a
+//   different code path and removes the last entry.
+// - Raft does not emit a HardState, but when the app calls advance(), it bumps
+//   its internal applied index cursor to 10 (when it should be 9)
+// - the next Ready asks the app to apply index 11 (omitting index 10), losing a
+//    write.
+#[test]
+fn test_commit_pagination_after_restart() {
```
BusyJay, line 593: The test case doesn't seem to be the same as upstream.

harness/tests/integration_cases/test_raw_node.rs
Fullstop000: `ents_count` is a `usize`, though I could use `0..ents_count as u64`

harness/tests/integration_cases/test_raw_node.rs
Fullstop000: It is silently a little different from the one in etcd due to the diff of `MemStorage`

harness/tests/integration_cases/test_raw_node.rs
Fullstop000: Let the node learns the current `commit_index` is 11. Altered to use `commit_to` instead.

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -492,3 +492,168 @@ fn test_set_priority() {
...
+        size += u64::from(e.compute_size());
+        ents.push(e);
+    }
+    s.inner.wl().append(&ents).unwrap();
+
+    let mut cfg = new_test_config(1, 10, 1);
+    // Set a max_size_per_msg that would suggest to Raft that the last committed entry should
+    // not be included in the initial rd.committed_entries. However, our storage will ignore
+    // this and *will* return it (which is how the Commit index ended up being 10 initially).
+    cfg.max_size_per_msg = size - 1;
```
BusyJay, line 622: I think it should be `size - uint64(s.ents[len(s.ents)-1].Size()) - 1`. The purpose is to let raft return 9 entries instead of 10, so that entry at 10 will get missed.

Fullstop000: Actually, the etcd codes of this test case are kind of buggy because the `firstIndex` here is `2` because the codes in this test *replace* the `ents` with start index `1` and the `firstIndex` of `MemoryStorage` is always `s.ents[0].Index + 1` due to the dummy entry (the index `0`).  Therefore, `len(rd.CommittedEntries)` in the first loop is 8 in etcd version. 

And since `MemStorage` in `raft-rs` has no such a dummy entry approach, we can correctly use `size - 1` here to return 9 committed entries in the first loop.

Though I found I should use `cfg.max_committed_size_per_ready` here 🤣

BusyJay:
@hicqu has re-implemented it as #440 with slight adjust. Close it now. Thanks for your contributions!

PR #357, opened 2020-04-05
Update comments

hhbti:
I updated some comments (and a little bit code) while reading the library (it was a fun journey.)

I'm new to both rust and raft. Not sure whether this can be useful for others. Let me know if I made anything wrong. :)


src/raw_node.rs
```diff
@@ -385,8 +385,8 @@ impl<T: Storage> RawNode<T> {
         false
     }
 
-    /// HasReady called when RawNode user need to check if any Ready pending.
-    /// Checking logic in this method should be consistent with Ready.containsUpdates().
+    /// Checks if any Ready is pending.
+    /// This method should be consistent with Ready.containsUpdates().
```
Fullstop000, line 389: Just for a hint, `containsUpdates()` seems not being implemented in this codebase.
It comes from https://github.com/etcd-io/etcd/blob/master/raft/node.go#L106

Fullstop000:
@BusyJay @hicqu PTAL

src/raft.rs
```diff
@@ -1072,9 +1072,22 @@ impl<T: Storage> Raft<T> {
                 let to_send = new_message(m.from, MessageType::MsgAppendResponse, None);
                 self.send(to_send);
             } else if m.get_msg_type() == MessageType::MsgRequestPreVote {
-                // Before pre_vote enable, there may be a recieving candidate with higher term,
-                // but less log. After update to pre_vote, the cluster may deadlock if
-                // we drop messages with a lower term.
+                // Unlike normal election, both sender and receiver of MsgReqPreVote{Resp}
```
BusyJay, line 1075: I think the original comment is quite sufficient. The comment in this PR may give a wrong impression that the check is for a rarely happen situation.

src/raft.rs
```diff
@@ -1121,15 +1135,14 @@ impl<T: Storage> Raft<T> {
...
-                    // When responding to Msg{Pre,}Vote messages we include the term
-                    // from the message, not the local term. To see why consider the
-                    // case where a single node was previously partitioned away and
-                    // it's local term is now of date. If we include the local term
-                    // (recall that for pre-votes we don't update the local term), the
-                    // (pre-)campaigning node on the other end will proceed to ignore
-                    // the message (it ignores all out of date messages).
-                    // The term in the original message and current local term are the
-                    // same in the case of regular votes, but different for pre-votes.
+                    // When responding to MsgReq{Pre}Vote, we include the term from the
```
BusyJay, line 1138: "," is required.

src/raft.rs
hhbti: Maybe just for me but it took me a while to figure out what the deadlock is. :)
Surely the example is misleading a verbose. How about I remote that part?

src/raft.rs
hhbti: May I know why? Is these comment semantically parsed? I thought only the comments start with /// are used in docs..

src/raw_node.rs
hhbti: Hmm maybe remove this line?

hhbti:
Thanks for taking time to read this boring change. 😀

By the way I have two questions about the design:
1. What's the purpose of local messages? Seems the same functionality can be implemented by a direct invocation of corresponding functions.

2. If implement this from scratch, an intuitive idea is to make it like a server: it waits for events, creates coroutines / fibers to process, and send out responses. This way request / response code can be well organized together. Timers can be used naturally. So what's wrong with this design?

Some reasons I can think of:
a) Needs to mocks more things to test.
b) Coroutines may introduce some non deterministic.

Is this correct?

Thanks a lot for your time.

PR #359, opened 2020-04-15
raft: introduce group commit

BusyJay:
Every progress is assigned to a group ID. only logs replicated to at
least two different groups will be committed. The algorithm won't change
the quorum definition, it works more like delaying committing.

Close #347


src/raft.rs
```diff
@@ -395,6 +395,51 @@ impl<T: Storage> Raft<T> {
...
+    /// Checks whether the raft group is using group commit and consistent
+    /// over group.
+    ///
+    /// If it can't get a correct answer, `None` is returned.
+    pub fn check_group_commit_consistent(&mut self) -> Option<bool> {
+        if self.state != StateRole::Leader {
+            return None;
+        }
+        if self.raft_log.term(self.raft_log.committed).unwrap_or(0) != self.term {
+            // Reject read only request when this leader has not committed any log entry
```
hicqu, line 435: Is the comment necessary? BTW it's better to extract the logic as a function.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4724,3 +4724,59 @@ fn test_request_snapshot_on_role_change() {
...
+        (vec![2, 2, 1], vec![1, 2, 1], 2),
+        (vec![2, 2, 1], vec![1, 1, 2], 1),
+        (vec![2, 2, 1], vec![1, 0, 1], 1),
+        (vec![2, 2, 1], vec![0, 0, 0], 2),
+        // even
+        (vec![4, 2, 1, 3], vec![0, 0, 0, 0], 2),
+        (vec![4, 2, 1, 3], vec![1, 0, 0, 0], 1),
+        (vec![4, 2, 1, 3], vec![0, 1, 0, 2], 2),
+        (vec![4, 2, 1, 3], vec![0, 2, 1, 0], 1),
+        (vec![4, 2, 1, 3], vec![1, 1, 1, 1], 1),
```
hicqu, line 4749: So if 4 peers are in one group, to commit a log needs 4 ACKs? I think we can ignore group ID if there is only 1 group.

src/progress/progress_set.rs
```diff
@@ -369,21 +368,49 @@ impl ProgressSet {
...
+            if checked_group_id == *group_id {
+                continue;
+            }
+            return (cmp::min(*index, quorum_commit_index), true);
+        }
+        if single_group {
+            (matched[quorum - 1].0, false)
+        } else {
+            (matched.last().unwrap().0, false)
+        }
```
gengliqi, line 413: I think this algorithm can't calculate the max commit index.
If the progress is (1, 1) (2, 2) (3, 1), this algorithm's answer is 1 but the correct answer is 2.

gengliqi:
I think the algorithm of calculating the commit index can be:
min(quorum_commit_index, min(max commit index of each group))

BusyJay:
PTAL

src/raft.rs
```diff
@@ -395,6 +395,75 @@ impl<T: Storage> Raft<T> {
...
+        if self.state != StateRole::Leader {
+            return None;
+        }
+        // Previous leader may have reach consistency already.
+        //
+        // check applied_index instead of committed_index to avoid pending conf change.
+        if !self.apply_to_current_term() {
+            return None;
+        }
+        let (index, use_group_commit) = self.prs.as_mut().unwrap().maximal_committed_index();
```
gengliqi, line 440: Using `self.mut_prs()` is shorter

src/raft.rs
```diff
@@ -652,7 +721,7 @@ impl<T: Storage> Raft<T> {
     /// Attempts to advance the commit index. Returns true if the commit index
     /// changed (in which case the caller should call `r.bcast_append`).
     pub fn maybe_commit(&mut self) -> bool {
-        let mci = self.prs().maximal_committed_index();
+        let mci = self.prs.as_mut().unwrap().maximal_committed_index().0;
```
gengliqi, line 724: `self.mut_prs()`

src/raft.rs
```diff
@@ -395,6 +395,75 @@ impl<T: Storage> Raft<T> {
...
+    /// groups are committed.
+    ///
+    /// The tuple is (`peer_id`, `group_id`). `group_id` should be larger than 0.
+    pub fn assign_commit_groups(&mut self, ids: &[(u64, u64)]) {
+        for (peer_id, group_id) in ids {
+            assert!(*group_id > 0);
+            if let Some(pr) = self.mut_prs().get_mut(*peer_id) {
+                pr.commit_group_id = *group_id;
+            } else {
+                return;
```
gengliqi, line 408: Is it valid that the `peer_id` in `ids` can not exist in this raft group?

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4724,3 +4724,170 @@ fn test_request_snapshot_on_role_change() {
...
+        (vec![2, 2, 1], vec![1, 0, 1], 1),
+        (vec![2, 2, 1], vec![0, 0, 0], 2),
+        // even
+        (vec![4, 2, 1, 3], vec![0, 0, 0, 0], 2),
+        (vec![4, 2, 1, 3], vec![1, 0, 0, 0], 1),
+        (vec![4, 2, 1, 3], vec![0, 1, 0, 2], 2),
+        (vec![4, 2, 1, 3], vec![0, 2, 1, 0], 1),
+        (vec![4, 2, 1, 3], vec![1, 1, 1, 1], 2),
+        (vec![4, 2, 1, 3], vec![1, 1, 2, 1], 1),
+        (vec![4, 2, 1, 3], vec![4, 3, 2, 1], 2),
```
gengliqi, line 4751: It makes me confused about the answer of `(vec![4, 2, 1, 3], vec![1, 0, 0, 0], 1)` and `(vec![4, 2, 1, 3], vec![1, 1, 1, 1], 2)`.
Maybe we can make the logic more straightforward, such as "if the kind of group number is greater than 2, we should use group commit"

src/raft.rs
BusyJay: Yes, but here should be continue instead of returning.

src/raft.rs
```diff
@@ -418,9 +439,6 @@ impl<T: Storage> Raft<T> {
         for (_, pr) in self.mut_prs().iter_mut() {
             pr.commit_group_id = 0;
         }
-        if StateRole::Leader == self.state && self.maybe_commit() {
```
hicqu: Why remove it?

BusyJay: If using group commit, then clearing group id will make commit index become smallest one, so no need to commit; if not using group commit, clearing group id has no effect.

PR #360, opened 2020-04-15
replace is_empty_snap free function

laizy:
`is_empty_snap` as a free function in `raw_node` module looks very strange. this pr add an `is_empty` method for `Snapshot` instead.


BusyJay:
Cool, can you also keep the old one for backward compatibility?

PR #361, opened 2020-04-20
Consider priority during election

### LINKED ISSUE #68: Consider priority during election
There may be differences for several nodes of a cluster, like some nodes are in a far away data center, or some nodes have poor hardware configurations. In such case we would want to make them have less chance to become a leader.

After pingcap/raft-rs#63, we support configuring different randomized timeout range for different nodes, which can make it hard to become leader with a larger timeout. However it can still cause noises. Because as long as the nodes have enough logs, it can still become leader even if other preferred nodes have the same logs.

So I propose to add a priority for every node. A follower votes for a candidate only when one of following prerequisites is met:

1. Candidate has more logs;
2. Candidate has the same logs and its priority is not less than the follower.

This policy is expected to work well when the nodes that are preferred to become leaders can form a quorum. In such case a non-preferred node can become leader only when it has more logs. However a node is not preferred usually means that it takes more time to catch up log for high network latency or poor performance as described above.

dimstars:
Fixes #68 
Add a priority to Raft. Election rules have also changed. Followers vote first for the candidate with more logs. With the same log, the followers vote for the candidate whose priority was no less than their own. The default priority is 0 and can be set through the interface within RawNode. And some tests have been added to examine several election scenarios.


nolouch:
PTAL @BusyJay

src/raft.rs
```diff
@@ -186,19 +186,24 @@ pub struct Raft<T: Storage> {
...
+    pub priority: u64,
 }
 
 trait AssertSend: Send {}
 
 impl<T: Storage + Send> AssertSend for Raft<T> {}
 
-fn new_message(to: u64, field_type: MessageType, from: Option<u64>) -> Message {
+/// ljr adds a parameter: priority.
+fn new_message(to: u64, field_type: MessageType, from: Option<u64>, priority: u64) -> Message {
```
BusyJay, line 199: I don't think it's necessary. Priority should only be set for request vote.

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -472,3 +472,15 @@ fn test_skip_bcast_commit() {
...
     assert_eq!(nt.peers[&3].raft_log.committed, 6);
 }
+
+/// test_set_priority checks the set_priority function in RawNode.
+#[test]
+fn test_set_priority() {
+    let l = default_logger();
+    let mut raw_node = new_raw_node(1, vec![1], 10, 1, new_storage(), &l);
+    let prioritys = vec![0, 1, 5, 10, 10000];
+    for &p in prioritys.iter() {
```
BusyJay, line 482: `for p in priorities` is preferred.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4920,3 +4920,132 @@ fn test_group_commit_consistent() {
...
+            network.peers[&id].state,
+            state,
+            "peer {} state (test {})",
+            id,
+            i + 1
+        );
+    }
+}
+
+/// test_election_with_priority_log varifies the correctness
```
BusyJay, line 4961: s/varifies/verifies/

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4920,3 +4920,132 @@ fn test_group_commit_consistent() {
         }
     }
 }
+
+/// test_election_with_priority verifies that a peer with a higher priority
+/// can win an election, but not when its priority is lower than most peers.
+#[test]
+fn test_election_with_priority() {
```
BusyJay, line 4927: This seems be covered by `test_election_with_priority_log`.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4920,3 +4920,132 @@ fn test_group_commit_consistent() {
...
+fn test_election_after_change_priority() {
+    let l = default_logger();
+    let mut n1 = new_test_raft(1, vec![1, 2, 3], 10, 1, new_storage(), &l);
+    let mut n2 = new_test_raft(2, vec![1, 2, 3], 10, 1, new_storage(), &l);
+    let mut n3 = new_test_raft(3, vec![1, 2, 3], 10, 1, new_storage(), &l);
+    // priority of n1 is 0 in default.
+    n2.set_priority(2);
+    n3.set_priority(3);
+    let entries = vec![new_entry(1, 1, SOME_DATA), new_entry(1, 1, SOME_DATA)];
+    n1.raft_log.append(&entries);
```
BusyJay, line 5019: The step seems unnecessary.

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -472,3 +472,15 @@ fn test_skip_bcast_commit() {
...
     assert_eq!(nt.peers[&3].raft_log.committed, 6);
 }
+
+/// test_set_priority checks the set_priority function in RawNode.
+#[test]
+fn test_set_priority() {
+    let l = default_logger();
+    let mut raw_node = new_raw_node(1, vec![1], 10, 1, new_storage(), &l);
+    let priorities = vec![0, 1, 5, 10, 10000];
+    for &p in priorities.iter() {
```
BusyJay, line 482: What I mean is `iter` is unnecessary.

dimstars: Sorry, I have modified it but did not commit.

dimstars:
Should I add “priority” in src/config.rs?

BusyJay:
Yes.

src/raft.rs
```diff
@@ -1181,7 +1195,10 @@ impl<T: Storage> Raft<T> {
                     // ...or this is a PreVote for a future term...
                     (m.get_msg_type() == MessageType::MsgRequestPreVote && m.term > self.term);
                 // ...and we believe the candidate is up to date.
-                if can_vote && self.raft_log.is_up_to_date(m.index, m.log_term) {
+                if can_vote
+                    && self.raft_log.is_up_to_date(m.index, m.log_term)
+                    && (m.index > self.raft_log.last_index() || self.priority <= m.priority)
```
hicqu, line 1200: priority should only work when `m.index == self.raft_log.last_index()`?

dimstars: It seems like this. A node with fewer logs cannot be a leader. The node with more logs should naturally be the leader.

hicqu: @dimstars would you like to update it?

dimstars: @hicqu Suggestions are welcome. I will make it better. (Or you mean this part needs to be modified? )

PR #362, opened 2020-04-23
proto: look for sources relative to CARGO_MANIFEST_DIR

jsgf:
This is a little more reliable than assuming the current dir is
the manifest dir.

Also override the default includes to also use the same path prefix, otherwise protoc complains.

### DIFF
--- a/proto/build.rs
+++ b/proto/build.rs
@@ -3,5 +3,9 @@
 use protobuf_build::Builder;
 
 fn main() {
-    Builder::new().search_dir_for_protos("proto").generate()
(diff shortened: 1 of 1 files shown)


BusyJay:
Hmm, but cargo always runs the build script where project locates. Maybe just a comment is enough?

jsgf:
We're importing this into our internal third-party repo which is also converted to build with Buck rather than Cargo. The environment buildscripts get run in is somewhat like real Cargo (most of the same env vars), but not entirely. This makes it easier to get working.

jsgf:
@nrc Does this need another review approval?

PR #363, opened 2020-05-14
check pending reads when quorum becomes small

### LINKED ISSUE #345: Quorum change should check read index
When quorum changes due to configuration change or quorum function change, read index should be checked eagerly. Otherwise read index will not be advanced until next heartbeat.

hicqu:
Close #345 .

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -4921,8 +4921,10 @@ fn test_group_commit_consistent() {
     }
 }
 
+// `test_read_when_quorum_becomes_small` tests read requests could be handled earlier
+// if quorum becomes less in configuration changes.
 #[test]
-fn test_read_when_quorum_becomes_small() {
+fn test_read_when_quorum_becomes_less() {
     let l = default_logger();
     let mut network = Network::new(vec![None, None], &l);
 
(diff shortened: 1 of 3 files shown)


src/raft.rs
```diff
@@ -2325,4 +2305,21 @@ impl<T: Storage> Raft<T> {
         m.request_snapshot = self.pending_request_snapshot;
         self.send(m);
     }
+
+    fn response_ready_read(&mut self, mut req: Message, index: u64) -> Option<Message> {
```
BusyJay, line 2309: s/response_ready_read/handle_ready_read_index/

src/raft.rs
```diff
@@ -2214,6 +2175,25 @@ impl<T: Storage> Raft<T> {
         if self.maybe_commit() {
             self.bcast_append();
         }
+
+        // The quorum size is now smaller, consider to response some read requests.
+        // If there is only one peer, all pending read requests must be responsed.
+        if let Some(ctx) = self.read_only.last_pending_request_ctx() {
```
BusyJay, line 2181: What if some reads in the front of the queue are ready to be processed?

harness/tests/integration_cases/test_raft.rs
```diff
@@ -4920,3 +4920,36 @@ fn test_group_commit_consistent() {
         }
     }
 }
+
+#[test]
+fn test_read_when_quorum_becomes_small() {
```
BusyJay, line 4925: Please add comment what it is testing.

src/raft.rs
hicqu: Without iterating the pending queue, they can only become ready when more `heartbeat responses` are received. Do you have any solution to handle them without iterating?

PR #364, opened 2020-05-19
add committed index in Progress

### LINKED ISSUE #325: Add a way to figure out followers commit index
Raft doesn't requires follower reports its commit index. But aggregated them can help us find out whether a log is committed in all nodes, which is quite useful in some cases.

For example, in TiKV, when enabling hibernate regions, it's better to let all the followers to commit to latest logs before going to sleep, otherwise follower can wait for a long time before committing and applying logs, which can hurt follower reads.

It can also be used in merge too.

accelsao:
Close #325

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -1411,28 +1411,14 @@ impl<T: Storage> Raft<T> {
             }
         }
 
-        match self.read_only.recv_ack(m) {
+        match self.read_only.recv_ack(m.from, &m.context) {
             Some(acks) if prs.has_quorum(acks) => {}
             _ => return,
         }
 
-        let rss = self.read_only.advance(m, &self.logger);
-        for rs in rss {
-            let mut req = rs.req;
-            if req.from == INVALID_ID || req.from == self.id {
-                // from local member
-                let rs = ReadState {
-                    index: rs.index,
-                    request_ctx: req.take_entries()[0].take_data(),
-                };
-                self.read_states.push(rs);
-            } else {
-                let mut to_send = Message::default();
-                to_send.set_msg_type(MessageType::MsgReadIndexResp);
-                to_send.to = req.from;
-                to_send.index = rs.index;
-                to_send.set_entries(req.take_entries());
-                ctx.more_to_send.push(to_send);
+        for rs in self.read_only.advance(&m.context, &self.logger) {
+            if let Some(m) = self.handle_ready_read_index(rs.req, rs.index) {
+                ctx.more_to_send.push(m);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/raft.rs
```diff
@@ -740,7 +740,15 @@ impl<T: Storage> Raft<T> {
     /// changed (in which case the caller should call `r.bcast_append`).
     pub fn maybe_commit(&mut self) -> bool {
         let mci = self.mut_prs().maximal_committed_index().0;
-        self.raft_log.maybe_commit(mci, self.term)
+        let ok = self.raft_log.maybe_commit(mci, self.term);
+        let self_id = self.id;
+        self.update_committed(self_id);
```
Fullstop000, line 745: Why use `self.id` as a `committed_index`?

accelsao: I am trying to pass `self.raft_log.committed` to progress set for updating committed index

BusyJay:
Progress committed index should be the index recorded in followers.

accelsao:
maybe let `Progress committed index` be `HashMap` for recording every raft node's committed index?

Fullstop000:
In the leader's step routine, maybe you can get a follower's `commit_index` from msgs like `MsgAppendResp` or `MsgHeartBeatResp` and update the related follower's `Progress`.

accelsao:
travis build error message with `The command "if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo fmt --all -- --check; fi" exited with 1.` 

I have no idea why `cargo fmt` failed.

And, do I need all the commit messages signed?

src/raft_log.rs
```diff
@@ -201,7 +201,7 @@ impl<T: Storage> RaftLog<T> {
         term: u64,
         committed: u64,
         ents: &[Entry],
-    ) -> Option<(u64, u64)> {
+    ) -> Option<(u64, u64, u64)> {
```
BusyJay, line 204: I don't think it's necessary to be changed.

src/raft.rs
```diff
@@ -1345,6 +1346,9 @@ impl<T: Storage> Raft<T> {
             return;
         }
 
+        // update committed index for the append responder
+        pr.update_committed(m.commit);
```
BusyJay, line 1350: It can be moved to L1319.

src/progress/mod.rs
```diff
@@ -179,6 +185,11 @@ impl Progress {
         need_update
     }
 
+    /// update committed_index.
+    pub fn update_committed(&mut self, committed_index: u64) {
+        self.committed_index = committed_index
```
BusyJay, line 190: Should skip smaller value.

accelsao: I think the committed index is non-decreasing, so we wont get the smaller value.
What I miss here?

BusyJay: Message can be reordered.

Fullstop000:
@accelsao DCO is required for every single commit in the PR :( 
One approach is to squash all the commits so far into one commit with your sign-off and force push it

accelsao:
whoops..

PR #366, opened 2020-05-22
add committed index in Progress

### LINKED ISSUE #325: Add a way to figure out followers commit index

accelsao:
Close #325

- [X] Commit index won't step backward.
- [X] Log rejection should not change commit index
- [X] Successfully Append should update commit index
- [x] Heartbeat should update commit index

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -219,9 +219,9 @@ fn test_progress_committed_index() {
     nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
     assert_eq!(nt.peers[&1].state, StateRole::Leader);
 
-    assert_raft_log(&format!("#1: "), &nt.peers[&1].raft_log, (1, 0, 1));
-    assert_raft_log(&format!("#2: "), &nt.peers[&2].raft_log, (1, 0, 1));
-    assert_raft_log(&format!("#3: "), &nt.peers[&3].raft_log, (1, 0, 1));
+    assert_raft_log(&"#1: ".to_string(), &nt.peers[&1].raft_log, (1, 0, 1));
+    assert_raft_log(&"#2: ".to_string(), &nt.peers[&2].raft_log, (1, 0, 1));
+    assert_raft_log(&"#3: ".to_string(), &nt.peers[&3].raft_log, (1, 0, 1));
 
     assert_eq!(
         nt.peers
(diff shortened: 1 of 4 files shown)


accelsao:
I think my test_case is useless..

accelsao:
error `The command "if [[ $TRAVIS_RUST_VERSION == "stable" && $TRAVIS_OS_NAME == "linux" ]]; then cargo fmt --all -- --check; fi" exited with 1.` in travis build is left

harness/tests/integration_cases/test_raft.rs
```diff
@@ -212,6 +212,121 @@ fn test_progress_update() {
     }
 }
 
+#[test]
+fn test_progress_committed() {
```
BusyJay, line 216: I think you can use `Network` to test the commit index change. It should validate following case:
1. Log rejection should not change commit index
2. Successfully Append should update commit index
3. Heartbeat should update commit index
4. Commit index won't step backward.

accelsao: I still do not know when will we get the smaller commit index

BusyJay: For example, leader sends two `MsgAppend` to follower and gets `MsgAppendResponse` in reverse order.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -212,6 +211,350 @@ fn test_progress_update() {
...
+        4
+    );
+
+    nt.isolate(2);
+    let m = new_message_with_entries(2, 2, MessageType::MsgPropose, vec![test_entries]);
+    nt.send(vec![m.clone(), m]);
+    nt.recover();
+
+    // test append entries rejection (update committed index = false)
+    let mut m = new_message(2, 1, MessageType::MsgAppend, 0);
```
BusyJay, line 385: Can `dispatch` and `read_messages` be used to avoid creating messages?

accelsao: thank advice, it makes life easier

harness/tests/integration_cases/test_raft.rs
```diff
@@ -212,6 +211,166 @@ fn test_progress_update() {
...
+
+    // #4 pass a smaller committed index, it occurs when the append response delay
+    let mut msg_append_response = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+    msg_append_response.term = 3;
+    msg_append_response.index = 10;
+    msg_append_response.commit = 10;
+
+    nt.send(vec![msg_append_response.clone()]);
+
+    assert_eq!(nt.peers[&1].prs().get(2).unwrap().committed_index, 10);
```
BusyJay, line 364: This seems a bug.

accelsao: shouldnt that commiitted index is update as 10?

BusyJay: If leader doesn't have log 10, it shouldn't commit to 10. Giving that raft-rs doesn't take much effort in byzantine fault tolerant, I think remove the case is OK for now.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -212,6 +211,192 @@ fn test_progress_update() {
...
+            nt.peers[&1].prs().get(1).unwrap().committed_index,
+            nt.peers[&1].prs().get(2).unwrap().committed_index,
+            nt.peers[&1].prs().get(3).unwrap().committed_index
+        ),
+        (0, 3, 1)
+    );
+
+    // #2 test heart beat
+    let heart_beat = new_message(1, 1, MessageType::MsgBeat, 0);
+    nt.send(vec![heart_beat]);
```
gengliqi, line 260: `heartbeat` is a word.

src/raft.rs
```diff
@@ -1939,6 +1945,7 @@ impl<T: Storage> Raft<T> {
             .maybe_append(m.index, m.log_term, m.commit, &m.entries)
         {
             to_send.set_index(last_idx);
+            to_send.set_commit(self.raft_log.committed);
             self.send(to_send);
```
gengliqi, line 1949: How about moving these two lines and 1964,1965 lines outside this conditional branch?

PR #367, opened 2020-05-23
fast log append rejection

### LINKED ISSUE #246: Fast log append rejection

accelsao:
Close #246 #291 .

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5569,3 +5569,86 @@ fn test_uncommitted_state_advance_ready_from_last_term() {
     // uncommitted size should be 12(remain unchanged since there's only one uncommitted entries)
     assert_eq!(nt.peers.get_mut(&2).unwrap().uncommitted_size(), data.len());
 }
+
+#[test]
+fn test_fast_log_rejection() {
+    let mut tests = vec![(
+        vec![
+            empty_entry(1, 1),
+            empty_entry(2, 2),
+            empty_entry(2, 3),
+            empty_entry(4, 4),
+            empty_entry(4, 5),
+            empty_entry(4, 6),
+            empty_entry(4, 7),
+        ],
+        vec![
+            empty_entry(1, 1),
+            empty_entry(2, 2),
+            empty_entry(2, 3),
+            empty_entry(3, 4),
+            empty_entry(3, 5),
+            empty_entry(3, 6),
+            empty_entry(3, 7),
+            empty_entry(3, 8),
+            empty_entry(3, 9),
+            empty_entry(3, 10),
+            empty_entry(3, 11),
+        ],
+        3,
+        7,
+        2,
+        3,
+    ),
+
+    ];
+    for (
+        i,
+        (leader_log, follower_log, reject_hint_term, reject_hint_index, next_append_term, next_append_index),
+    ) in tests.drain(..).enumerate()
+    {
+        // TODO(accelsao): maybe we need `Copy` of Message
+        let l = default_logger();
+        let s1 = MemStorage::new_with_conf_state((vec![1, 2, 3], vec![]));
+        s1.wl().append(&leader_log).unwrap();
+        let s2 = MemStorage::new_with_conf_state((vec![1, 2, 3], vec![]));
+        s2.wl().append(&follower_log).unwrap();
+        let mut n1 = new_test_raft(1, vec![1,2,3], 10, 1, s1, &l);
+        let mut n2 = new_test_raft(2, vec![1,2,3], 10, 1, s2, &l);
+        n1.become_candidate();
+        n1.become_leader();
+        n2.step(new_message(2, 2, MessageType::MsgHeartbeat, 0)).unwrap();
+        let msgs = n2.read_messages();
+        assert_eq!(msgs.len(), 1, "#{}", i);
+        assert_eq!(msgs[0].get_msg_type(), MessageType::MsgHeartbeatResponse, "#{}", i);
+        n1.step(new_message(2, 2, MessageType::MsgHeartbeatResponse, 0)).unwrap();
+        let msgs = n1.read_messages();
+        assert_eq!(msgs.len(), 1, "#{}", i);
+        assert_eq!(msgs[0].get_msg_type(), MessageType::MsgAppend, "#{}", i);
+        let mut m = new_message_with_entries(1, 2, MessageType::MsgAppend, vec![empty_entry(1, 8)]);
+        m.term = 1;
+        m.log_term = 4;
+        m.index = 7;
+        // println!("msgs: {:?}", msgs[0]);
+        n2.step(m).unwrap();
+        let msgs = n2.read_messages();
+        assert_eq!(msgs.len(), 1, "#{}", i);
+        assert_eq!(msgs[0].get_msg_type(), MessageType::MsgAppendResponse, "#{}", i);
+        assert!(msgs[0].reject, "#{}", i);
+        assert_eq!(msgs[0].reject_hint, reject_hint_index, "#{}", i);
+        assert_eq!(msgs[0].log_term, reject_hint_term, "#{}", i);
+        println!("msgs: {:?}", msgs[0]);
+        let mut m = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+        m.term = 1;
+        m.index = 7;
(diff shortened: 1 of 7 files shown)


src/raft.rs
```diff
@@ -1323,9 +1323,15 @@ impl<T: Storage> Raft<T> {
                 "last index" => m.reject_hint,
                 "from" => m.from,
                 "index" => m.index,
+                "last term" => m.log_term,
             );
 
             if pr.maybe_decr_to(m.index, m.reject_hint, m.request_snapshot) {
+                let conflict_index = self.raft_log.find_conflict_by_term(pr.next_idx, m.log_term);
```
hicqu, line 1330: For backward compatibility, please check `m.log_term == 0`.

hicqu:
Thanks for your contribution! Have you already read #291 ? Seems two implementations are same. Or you get some new discoveries?

accelsao:
> Thanks for your contribution! Have you already read #291 ? Seems two implementations are same. Or you get some new discoveries?

Yes, I had. I rewrite the test case to test the append response but not `test_find_conflict_by_term` only, because I think it is not enough. Also get conflict_index directly without checking `m.log_term != 0 && pr.next_idx == m.reject_hint + 1`.

src/raft.rs
accelsao: > For backward compatibility, please check `m.log_term == 0`.

I can fix this, but I wonder when does `m.log_term == 0` occur? After become `leader`, the term is at least 1, unless we send message with log_term 0 for testing.

hicqu: Because `log_term` is not filled for `MsgAppendResponse` before this PR, so it will be 0. So if a Raft group is in rolling update, it could become a problem.

hicqu:
> pr.next_idx == m.reject_hint + 1
It's ok to remove this check. You can see if the expression is true, `conflict_idx` will always be 0.

src/raft_log.rs
```diff
@@ -183,6 +183,19 @@ impl<T: Storage> RaftLog<T> {
...
 
+    /// Finds the index of the conflict with the given term
+    ///
+    /// It returns the first index that the term is not
+    /// greater than given term
+    pub fn find_conflict_by_term(&self, mut index: u64, term: u64) -> u64 {
+        let mut conflict_index = 0;
+        while self.term(index).ok().map_or(false, |t| t > term) {
+            index -= 1;
+            conflict_index = index
```
hicqu, line 194: better to add a `;` at end.

hicqu:
rest LGTM. Thanks!

hicqu (dismissed):

BusyJay:
Can you send the change to etcd-io/etcd first?

accelsao:
> Can you send the change to etcd-io/etcd first?

there is no similar issue in etcd

hicqu:
@accelsao yes, to keep 2 repo sync is helpful. If you are not familiar to golang, I can help you to port to etcd/raft.

accelsao:
> @accelsao yes, to keep 2 repo sync is helpful. If you are not familiar to golang, I can help you to port to etcd/raft.

can you help me? I want write more rust code :D

hicqu:
Ported to etcd/raft: https://github.com/etcd-io/etcd/pull/11964.

hicqu (approved):

NingLin-P (dismissed):
Rest LGTM

NingLin-P (approved):

src/raft_log.rs
```diff
@@ -194,6 +194,41 @@ impl<T: Storage> RaftLog<T> {
...
+                index,
+                self.last_index()
+            );
+        }
+
+        loop {
+            match self.term(conflict_index) {
+                Ok(t) => {
+                    if t > term {
+                        conflict_index -= 1
```
BusyJay, line 222: Is it possible to underflow?

accelsao: if index is out of range, we get `0`, as long as the given `term` is non-negative,  we will return immediately, so no underflow.

harness/tests/integration_cases/test_raft_paper.rs
```diff
@@ -647,23 +647,27 @@ fn test_follower_check_msg_append() {
     let ents = vec![empty_entry(1, 1), empty_entry(2, 2)];
     let mut tests = vec![
         // match with committed entries
-        (0, 0, 1, false, 0, 1),
-        (ents[0].term, ents[0].index, 1, false, 0, 1),
+        (0, 0, 1, 1, false, 0, 0),
```
BusyJay, line 650: should still keep original `w_commit` field.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -5567,3 +5567,280 @@ fn test_uncommitted_state_advance_ready_from_last_term() {
...
+            .unwrap();
+        let msgs = n2.read_messages();
+        assert_eq!(msgs.len(), 1, "#{}", i);
+        assert_eq!(
+            msgs[0].get_msg_type(),
+            MessageType::MsgHeartbeatResponse,
+            "#{}",
+            i
+        );
+        n1.step(new_message(2, 2, MessageType::MsgHeartbeatResponse, 0))
```
BusyJay, line 5811: Why not just `n1.step(msgs[0])`?

BusyJay (commented):
Can you also add the comment for proto file?

BusyJay:
CI fail

harness/tests/integration_cases/test_raft.rs
accelsao: No Copy for Message, I am not familiar with proto custom implement yet

BusyJay: If `msgs` is a vec, then it's move instead of copy. Besides, copy works the same as recreate the message and more readable.

harness/tests/integration_cases/test_raft_paper.rs
accelsao: I reorder `w_commit` before group of `reject`

harness/tests/integration_cases/test_raft.rs
accelsao: But the error says it is because of `Copy`
```
error[E0507]: cannot move out of index of `Vec<raft::prelude::Message>`
    --> harness/tests/integration_cases/test_raft.rs:5811:17
     |
5811 |         n1.step(msgs[0]).unwrap();
     |                 ^^^^^^^ move occurs because value has type `raft::prelude::Message`, which does not implement the `Copy` trait

error: aborting due to previous error
```

accelsao: I found that I can use `pop()` for a workaround

src/raft_log.rs
```diff
@@ -194,6 +194,41 @@ impl<T: Storage> RaftLog<T> {
...
+    /// Return (index, term)
+    pub fn find_conflict_by_term(&self, index: u64, term: u64) -> (u64, Option<u64>) {
+        let mut conflict_index = index;
+
+        if index > self.last_index() {
+            warn!(
+                self.unstable.logger,
+                "index({}) is out of range [0, last_index({})] in find_conflict_by_term",
+                index,
+                self.last_index()
```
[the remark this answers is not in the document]
accelsao: > Can you also add the comment for proto file?

Which one?

BusyJay:
I still don't see the comment for proto file.

Rest LGTM

accelsao:
> I still don't see the comment for proto file.

which file?

BusyJay (approved):

BusyJay:
@NingLin-P PTAL again, since some changes are included.

hicqu (approved):

PR #368, opened 2020-05-25
advance read-index queue if configuration change reduces quorum

hicqu:


tests/integration_cases/test_raft.rs
```diff
@@ -3914,7 +3914,7 @@ fn test_remove_learner() {
     let mut n1 = new_test_learner_raft(1, vec![1], vec![2], 10, 1, new_storage());
     n1.remove_node(2);
     assert_eq!(n1.prs().nodes(), vec![1]);
-    assert_eq!(n1.prs().learner_nodes(), vec![]);
+    assert_eq!(n1.prs().learner_nodes(), Vec::default());
```
hicqu, line 3917: It should be an old bug fixed in master branch, but not picked to 0.4.x.

tests/integration_cases/test_raft.rs
```diff
@@ -2467,7 +2467,7 @@ fn test_read_only_option_lease_without_check_quorum() {
     let read_states = &nt.peers[&2].read_states;
     assert!(!read_states.is_empty());
     let rs = &read_states[0];
-    assert_eq!(rs.index, INVALID_ID);
+    assert_eq!(rs.index, 1);
```
hicqu, line 2470: It should be an old bug fixed in master branch, but not picked to 0.4.x.

hicqu:
Seems check quorum doesn't work correctly in 0.4.x. So this PR makes it be more easily to return an uncommitted index. Close it.

PR #371, opened 2020-06-10
skip broadcasting heartbeat to up-to-date learner

### LINKED ISSUE #33: Skip broadcasting heartbeat to learner
Since learner don't vote, so it's meaningless to send heartbeat to learner if it is up to date already. If learner is not up to date, heartbeat is still needed for updating progress and commit.

accelsao:
close #33


BusyJay:
Thanks @accelsao. Our practice finds the optimization is not practical any more. Detail explanation is written in the original issue. Sorry to waste your time.

If you want to implement another optimization, better discuss about it on the issue, so that we can confirm whether it's worthy to give it a try.

PR #375, opened 2020-06-12
update progress.committed_index for leader

NingLin-P:
Signed-off-by: linning <linningde25@gmail.com>

Do few thing for https://github.com/tikv/raft-rs/pull/366 :
- update `committed_index` for leader

### DIFF
--- a/src/progress/mod.rs
+++ b/src/progress/mod.rs
@@ -116,7 +116,6 @@ impl Progress {
     pub(crate) fn reset(&mut self, next_idx: u64) {
         self.matched = 0;
         self.next_idx = next_idx;
-        self.committed_index = 0;
         self.state = ProgressState::default();
         self.paused = false;
         self.pending_snapshot = 0;
--- a/src/progress/mod.rs
+++ b/src/progress/mod.rs
@@ -116,6 +116,7 @@ impl Progress {
     pub(crate) fn reset(&mut self, next_idx: u64) {
         self.matched = 0;
         self.next_idx = next_idx;
+        self.committed_index = 0;
         self.state = ProgressState::default();
         self.paused = false;
         self.pending_snapshot = 0;
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -754,7 +754,15 @@ impl<T: Storage> Raft<T> {
     /// changed (in which case the caller should call `r.bcast_append`).
     pub fn maybe_commit(&mut self) -> bool {
         let mci = self.mut_prs().maximal_committed_index().0;
-        self.raft_log.maybe_commit(mci, self.term)
+        if self.raft_log.maybe_commit(mci, self.term) {
+            let (self_id, committed) = (self.id, self.raft_log.committed);
+            self.mut_prs()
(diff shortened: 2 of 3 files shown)


NingLin-P:
PTAL @accelsao

src/progress/mod.rs
```diff
@@ -116,6 +116,7 @@ impl Progress {
     pub(crate) fn reset(&mut self, next_idx: u64) {
         self.matched = 0;
         self.next_idx = next_idx;
+        self.committed_index = 0;
```
BusyJay, line 119: Why reset it?

NingLin-P: Reset it to 0 so we can know whether `committed_index` is stale.

BusyJay: Actually you can never know if `committed_index` is stale, because follower can update committed index without informing leader. The only thing is for sure is follower's commit index >= `committed_index`. And the guarantee holds no matter reseting it or not.

src/raft.rs
```diff
@@ -2082,6 +2092,7 @@ impl<T: Storage> Raft<T> {
         let next_idx = self.raft_log.last_index() + 1;
         prs.restore_snapmeta(meta, next_idx, self.max_inflight);
         prs.get_mut(self.id).unwrap().matched = next_idx - 1;
+        prs.get_mut(self.id).unwrap().committed_index = self.raft_log.committed;
```
NingLin-P, line 2095: When handling snapshot, we will come here to set the `progress` for itself first, then restore the raft log, and the `progress` we set here will be stale. Should we move these two lines of code after `self.raft_log.restore(snap);` ?

src/progress/mod.rs
NingLin-P: Yes, it's true, but if we did not reset it, the value will be much staler, and I think it will be quite strange to see it at the log (if people add it to the log).

BusyJay: I think 0 is as strange as other numbers. Even stale commit indexes are informative. For example, applications can still rely on the commit indexes from a frequently re-elected leader without waiting for election to complete and a round of heartbeat.

src/raft.rs
BusyJay: Yes, it should. It seems a bug exists for a long time.

NingLin-P: I think we can just delete them as this peer must not be leader when handling snapshot, and `progress` for the follower is useless.

src/progress/mod.rs
NingLin-P: OK, let's keep it since it's still useful for the applications.

src/raft.rs
BusyJay: How about leave it as it is? We can sync the changes with etcd/raft later.

gengliqi: 🤣 It's no harm because this peer is a follower and follower's progress is useless.

gengliqi:
The ci failed due to https://github.com/rust-lang/cargo/pull/8353.

PR #380, opened 2020-06-26
*: add quorum package

BusyJay:
Adds quorum package and majority configuration. Configuration in tracker
is also updated. The quorum package is ported from etcd master.

### DIFF
--- a/src/quorum.rs
+++ b/src/quorum.rs
@@ -1,0 +1,56 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
+
+pub mod majority;
+
+use std::collections::HashMap;
+use std::fmt::{self, Debug, Display, Formatter};
+
+/// VoteResult indicates the outcome of a vote.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum VoteResult {
+    /// Pending indicates that the decision of the vote depends on future
+    /// votes, i.e. neither "yes" or "no" has reached quorum yet.
+    Pending,
+    // Lost indicates that the quorum has voted "no".
+    Lost,
+    // Won indicates that the quorum has voted "yes".
+    Won,
+}
+
+/// Index is a Raft log position.
+#[derive(Default, Clone, Copy)]
+pub struct Index {
+    pub index: u64,
+    pub group_id: u64,
+}
+
+impl Display for Index {
+    #[inline]
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        if self.index != u64::MAX {
+            write!(f, "[{}]{}", self.group_id, self.index)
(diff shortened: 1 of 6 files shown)


BusyJay:
@Fullstop000 Do you have time to review this? You can use command "LGTM" to give your approval.

src/quorum/majority.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// Eg. If the matched indexes are [2,2,2,4,5], it will return 2.
+    /// If the matched indexes and groups are `[(1, 1), (2, 2), (3, 2)]`, it will return 1.
+    pub fn committed_index(&self, use_group_commit: bool, l: &impl AckedIndexer) -> (u64, bool) {
+        if self.voters.is_empty() {
+            // This plays well with joint quorums which, when one half is the zero
+            // MajorityConfig, should behave like the other half.
+            return (u64::MAX, false);
+        }
+
+        let mut stack_arr: [MaybeUninit<Index>; 7] = unsafe { MaybeUninit::uninit().assume_init() };
```
hicqu, line 49: I prefer `SmallVec` for this.

BusyJay: I didn't use `SmallVec` on purpose. On the one hand, it's such a small piece of code that doesn't need to introduce another dependency; on the other hand, it has less condition branch as every operation is inlined and optimized when using array.

src/tracker.rs
```diff
@@ -199,7 +165,7 @@ impl ProgressSet {
         let pr = Progress::new(next_idx, max_inflight);
         for id in &meta.conf_state.as_ref().unwrap().voters {
             self.progress.insert(*id, pr.clone());
-            self.configuration.voters.insert(*id);
+            self.configuration.voters.voters.insert(*id);
```
Fullstop000, line 168: This looks wired... Is there potential extra fields in `Configuration` besides `voters` ?

src/quorum.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+impl Debug for Index {
+    #[inline]
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        Display::fmt(self, f)
+    }
+}
+
+pub trait AckedIndexer {
```
Fullstop000, line 45: It could be better to add some comments for this public trait

src/tracker.rs
BusyJay: Yes. It will be added when Joint configuration is ported.

src/quorum.rs
BusyJay: It's not a public trait. It can only be accessed by the crate. We have `#![deny(missing_docs)]` check.

Fullstop000:
LGTM

PR #381, opened 2020-06-29
remove storage error other

hicqu:

`StorageError::Other` is not used in raft-rs, it can be removed.


BusyJay:
It maybe used by implementer of storage trait.

PR #382, opened 2020-07-02
*: introduce joint quorum

BusyJay:
This PR ports joint quorum. It rename ProgressSet to ProgressTracker
and keep both name and field layout just like etcd.

In addition, the PR fixes a performance issue discovered earlier that
campaign prints too much logs.

joint.rs is ported from etcd/raft/quorum/joint.go and tracker is from
etcd/raft/tracker/tracker.go.

### DIFF
--- a/src/quorum/joint.rs
+++ b/src/quorum/joint.rs
@@ -1,0 +1,80 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
+
+use super::{AckedIndexer, VoteResult};
+use crate::util::Union;
+use crate::HashSet;
+use crate::MajorityConfig;
+use std::cmp;
+
+/// A configuration of two groups of (possibly overlapping) majority configurations.
+/// Decisions require the support of both majorities.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Configuration {
+    pub(crate) incoming: MajorityConfig,
+    pub(crate) outgoing: MajorityConfig,
+}
+
+impl Configuration {
+    /// Creates a new configuration using the given IDs.
+    pub fn new(voters: HashSet<u64>) -> Configuration {
+        Configuration {
+            incoming: MajorityConfig::new(voters),
+            outgoing: MajorityConfig::default(),
+        }
+    }
+
+    /// Creates an empty configuration with given capacity.
+    pub fn with_capacity(cap: usize) -> Configuration {
+        Configuration {
+            incoming: MajorityConfig::with_capacity(cap),
+            outgoing: MajorityConfig::default(),
+        }
+    }
+
+    /// Returns the largest committed index for the given joint quorum. An index is
+    /// jointly committed if it is committed in both constituent majorities.
+    pub fn committed_index(&self, use_group_commit: bool, l: &impl AckedIndexer) -> (u64, bool) {
+        let i_idx = self.incoming.committed_index(use_group_commit, l);
+        let o_idx = self.outgoing.committed_index(use_group_commit, l);
+        (cmp::min(i_idx.0, o_idx.0), i_idx.1 && o_idx.1)
+    }
+
+    /// Takes a mapping of voters to yes/no (true/false) votes and returns a result
+    /// indicating whether the vote is pending, lost, or won. A joint quorum requires
+    /// both majority quorums to vote in favor.
+    pub fn vote_result(&self, check: impl Fn(u64) -> Option<bool>) -> VoteResult {
+        let i = self.incoming.vote_result(&check);
+        let o = self.outgoing.vote_result(check);
+        match (i, o) {
+            // It won if won in both.
+            (VoteResult::Won, VoteResult::Won) => VoteResult::Won,
+            // It lost if lost in either.
+            (VoteResult::Lost, _) | (_, VoteResult::Lost) => VoteResult::Lost,
+            // It remains pending if pending in both or just won in one side.
+            _ => VoteResult::Pending,
+        }
+    }
+
+    /// Clears all IDs.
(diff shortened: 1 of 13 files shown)


BusyJay:

src/raft.rs
```diff
@@ -1052,37 +1051,29 @@ impl<T: Storage> Raft<T> {
...
-            self.logger,
-            "{id} received message from {from}",
-            id = self.id,
-            from = self_id;
-            "msg" => ?vote_msg,
-            "term" => self.term
-        );
-        self.register_vote(self_id, acceptance);
-        if let Some(true) = self.check_votes() {
+        if VoteResult::Won == self.poll(self_id, vote_msg, true) {
```
Connor1996, line 1054: In etcd, the logic is
```
if VoteResult::Won == self.poll(self_id, vote_msg, true) {
  if campaign_type == CAMPAIGN_PRE_ELECTION {
     self.campaign(campaignEletion);
  else {
     self.becomeLeader();
  }
  return;
}

File: src/raft.rs [hunk 2]
```diff
@@ -1052,37 +1051,29 @@ impl<T: Storage> Raft<T> {
...
-                "[logterm: {log_term}, index: {log_index}] sent request to {id}",
-                log_term = self.raft_log.last_term(),
-                log_index = self.raft_log.last_index(),
-                id = id;
-                "term" => self.term,
-                "msg" => ?vote_msg,
-            );
-            let mut m = new_message(*id, vote_msg, None);
+
+            if voter_cnt < 7 {
```
Connor1996, line 1069: why 7? why just let it print all at the end even it's more than 7.

src/quorum/joint.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::util::Union;
+use crate::HashSet;
+use crate::MajorityConfig;
+use std::cmp;
+
+/// A configuration of two groups of (possibly overlapping) majority configurations.
+/// Decisions require the support of both majorities.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Configuration {
+    pub(crate) incoming: MajorityConfig,
```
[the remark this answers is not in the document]
BusyJay: I adopt the names from [etcd](https://github.com/etcd-io/etcd/blob/15884e90854981494e6889bdb663439612b547a7/raft/confchange/confchange.go#L407-L408).

src/quorum/majority.rs
```diff
@@ -43,7 +48,7 @@ impl Configuration {
         if self.voters.is_empty() {
             // This plays well with joint quorums which, when one half is the zero
             // MajorityConfig, should behave like the other half.
-            return (u64::MAX, false);
+            return (u64::MAX, true);
```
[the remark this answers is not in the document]
BusyJay: When there is zero member, whether it's using group commit depends on other configuration. It's a mistake to set it to false in the past.

src/raft.rs
BusyJay: To avoid allocation.

src/tracker.rs
```diff
@@ -44,6 +45,47 @@ pub struct Configuration {
...
+    ///   voters:   {1 2}
+    ///   learners: {3}
+    ///   next_learners: {}
+    ///
+    /// Note that next_learners is not used while adding a learner that is not
+    /// also a voter in the joint config. In this case, the learner is added
+    /// right away when entering the joint configuration, so that it is caught up
+    /// as soon as possible.
+    #[get = "pub"]
+    learner_next: HashSet<u64>,
```
Fullstop000, line 83: Prefer `learners_next` to follow `learners`

src/raft.rs
```diff
@@ -1052,37 +1051,28 @@ impl<T: Storage> Raft<T> {
...
-                "[logterm: {log_term}, index: {log_index}] sent request to {id}",
-                log_term = self.raft_log.last_term(),
-                log_index = self.raft_log.last_index(),
-                id = id;
-                "term" => self.term,
-                "msg" => ?vote_msg,
-            );
-            let mut m = new_message(*id, vote_msg, None);
+
+            if voter_cnt == 7 {
```
Fullstop000, line 1069: Maybe use `voters.len()` instead

src/quorum/joint.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Returns the largest committed index for the given joint quorum. An index is
+    /// jointly committed if it is committed in both constituent majorities.
+    ///
+    /// The bool flag indicates whether the index is computed by group commit algorithm
+    /// successfully. It's true only when both majorities use group commit.
+    pub fn committed_index(&self, use_group_commit: bool, l: &impl AckedIndexer) -> (u64, bool) {
+        let i_idx = self.incoming.committed_index(use_group_commit, l);
+        let o_idx = self.outgoing.committed_index(use_group_commit, l);
+        (cmp::min(i_idx.0, o_idx.0), i_idx.1 && o_idx.1)
```
Fullstop000, line 42: Maybe using match pattern could be more readable?
```rust
let (i_idx, i_use_group) = self.incoming.committed_index(use_group_commit, l);
let (o_idx, o_use_group) = self.outgoing.committed_index(use_group_commit, l);
(cmp::min(i_idx, o_idx), i_use_group && o_use_group)
```

src/raft.rs
BusyJay: voters is an array, its length is always 7.

Fullstop000: I think it could be better to use `voter_cnt == voters.len()` to replace the magic number

Fullstop000:
LGTM

BusyJay:
Thanks! @Connor1996 @Fullstop000

PR #384, opened 2020-07-13
port joint quorum test

jayzhan211:
# TODO
- [x] joint commit single group
- [x] joint commit multi group
- [x] joint vote
- [x] majority commit single group
- [x] majority commit multi group
- [x] majority vote
- [ ] maybe more test for group commit

## Difference from etcd
1. add group commit
2. no alternative compute test (is it necessary? I think this is the test for computing majority) 
https://github.com/etcd-io/etcd/blob/69f7e163c652cd13e8d2c435ff1dbba47e2930cc/raft/quorum/quick_test.go#L84-L85


BusyJay:
Cool! How about making those test cases in quorum package instead of integration tests? So we can access private fields without adding explicit contructors.

BusyJay:
@HunDunDM @nolouch PTAL

Cargo.toml
```diff
@@ -36,6 +36,7 @@ slog = "2.2"
 slog-envlogger = { version = "2.1.0", optional = true }
 slog-stdlog = { version = "4", optional = true }
 slog-term = { version = "2.4.0", optional = true }
+anyhow = "1.0.32"
```
BusyJay, line 39: This should be `dev-dependencies` too.

BusyJay (commented):
The PR is too large, please add one test (file) at a time.

Cargo.toml
accelsao: > The PR is too large, please add one test (file) at a time.

I will divide them as single group commit , vote and group commit

accelsao:
I would like to replace `cargo clippy --all -- -D clippy::all` to `cargo clippy --all --all-targets -- -D clippy::all` in `Readme.md`, is the latter better?

PR #385, opened 2020-07-13
*: introduce eraftpb fixture

BusyJay:
All changes are from etcd/raft/raftpb.

### DIFF
--- a/proto/src/confstate.rs
+++ b/proto/src/confstate.rs
@@ -1,0 +1,27 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
+
+use crate::eraftpb::ConfState;
+
+fn eq_without_order(lhs: &[u64], rhs: &[u64]) -> bool {
+    for l in lhs {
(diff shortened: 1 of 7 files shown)


proto/src/confstate.rs
```diff
@@ -0,0 +1,27 @@
...
+        if !lhs.contains(r) {
+            return false;
+        }
+    }
+    true
+}
+
+// Returns a nil error if the inputs describe the same configuration.
+// On mismatch, returns a descriptive error showing the differences.
+pub fn conf_state_eq(lhs: &ConfState, rhs: &ConfState) -> bool {
```
Fullstop000, line 21: Seems the behavior and its comments are not matched

BusyJay:
@hundunDM, @nolouch PTAL

proto/src/confstate.rs
BusyJay: Oh, my bad.

Fullstop000:
LGTM

PR #386, opened 2020-07-14
*: use joint consensus

BusyJay:
This PR updates raft to utilize joint consensus. Several outdated methods from ProgressTracker and benches are removed.

Most codes are ported with the same name and same location. Except `switchToConfig` is ported as `post_conf_change` as we mutates progress inline.

Tests are updated to get it passed. But more tests need to be ported later.

@HunDunDM  @nolouch PTAL


src/raft.rs
```diff
@@ -2096,176 +2151,154 @@ impl<T: Storage> Raft<T> {
...
-    pub fn remove_node(&mut self, id: u64) -> Result<()> {
-        self.mut_prs().remove(id)?;
-
-        // do not try to commit or abort transferring if there are no voters in the cluster.
-        if self.prs().voter_ids().is_empty() {
-            return Ok(());
+    /// Updates the in-memory state and, when necessary, carries out additional actions
+    /// such as reacting to the removal of nodes or changed quorum requirements.
+    pub fn post_conf_change(&mut self) -> ConfState {
+        info!(self.logger, "switched to configuration"; "config" => ?self.prs.conf());
```
nolouch, line 2257: In fact, it was already switched during `restore`, a little confused about the log and the function command.

src/raft.rs
```diff
@@ -2096,176 +2151,154 @@ impl<T: Storage> Raft<T> {
...
+        // TODO: instead of creating a conf state, validating conf state inside
+        // progress tracker is better.
+        let cs = self.prs.conf().to_conf_state();
+        let is_voter = self.prs.conf().voters.contains(self.id);
+        self.promotable = is_voter;
```
nolouch, line 2262: Why not use `prs.get()` directly? the commands about promotable say the voter should in progress list.

src/raft.rs
BusyJay: Indeed, that's why the logs say "switched".

src/raft.rs
BusyJay: It's guaranteed all voters are in progress. It's checked by

https://github.com/tikv/raft-rs/blob/master/src/confchange/changer.rs#L293-L300

proto/src/confstate.rs
```diff
@@ -17,7 +17,18 @@ fn eq_without_order(lhs: &[u64], rhs: &[u64]) -> bool {
 }
 
 // Returns true if the inputs describe the same configuration.
+#[must_use]
 pub fn conf_state_eq(lhs: &ConfState, rhs: &ConfState) -> bool {
+    // In most cases, the orders are the same.
```
Fullstop000, line 22: Maybe needs a more precise description like in what cases the orders aren't same

src/raft.rs
```diff
@@ -1688,20 +1711,52 @@ impl<T: Storage> Raft<T> {
...
                         } else {
-                            self.pending_conf_index = self.raft_log.last_index() + i as u64 + 1;
+                            ""
                         }
+                    };
+
+                    if reason.is_empty() {
+                        self.pending_conf_index = self.raft_log.last_index() + i as u64 + 1;
+                    } else {
+                        info!(
```
abbccdda, line 1749: Should all the above reasons be set as info level? For example, does empty conf change actually exist?

src/raft.rs
```diff
@@ -272,20 +278,15 @@ impl<T: Storage> Raft<T> {
...
-                r.promotable = true;
-            }
-        }
-        for p in learners {
-            let pr = Progress::new(1, r.max_inflight);
-            if let Err(e) = r.prs.insert_learner(*p, pr) {
-                fatal!(r.logger, "{}", e);
-            };
+        confchange::restore(&mut r.prs, r.r.raft_log.last_index(), conf_state)?;
+        let new_cs = r.post_conf_change();
```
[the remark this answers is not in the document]
BusyJay: Rust uses abbreviation as well. For example, you can see `Vec` and `U64` instead of `Vector` and `UnsignedInt64`. We generally follows both the style of [Rust](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md) and https://pingcap.github.io/style-guide/rust/.

src/raft.rs
BusyJay: Yes. See https://github.com/tikv/raft-rs/pull/386/files#diff-18f3dab6275bbc16b4a102e73378c08bR828-R832.

src/raft.rs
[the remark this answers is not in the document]
Fullstop000: I think `post_conf_change` is better because it's actually doing some side effects to `r` **after** `r.prs` restored from confstate

src/raft.rs
```diff
@@ -2096,176 +2151,154 @@ impl<T: Storage> Raft<T> {
...
+            // at the time writing but hypothetically we handle them the same way as
+            // removing the leader: stepping down into the next Term.
+            //
+            // TODO(tbg): step down (for sanity) and ask follower with largest Match
+            // to TimeoutNow (to avoid interruption). This might still drop some
+            // proposals but it's better than nothing.
+            //
+            // TODO(tbg): test this branch. It is untested at the time of writing.
+            return cs;
+        }
```
Fullstop000, line 2274: Can do some "safe operations" to evict leader itself e.g. `r.become_follower` first?  How does TiKV handle such a situation currently or maybe PD can just avoid this?

Fullstop000:
@BusyJay Maybe we can add a specialized method, proposing the empty `ConfChangeV2`, into rawnode for leaving `Explict` transition or doc this in somewhere else?

Rest LGTM besides the CI error.

src/raft.rs
BusyJay: TiKV rejects removing leader. It simply drops any proposal that tries to remove itself. However, it's not guaranteed.

BusyJay:
I added a comment to `propose_conf_change`. And now CI passes.

PR #387, opened 2020-07-16
port etcd confchange test

accelsao:
etcd confchange test
no quick test
no restore test


src/confchange/testdata/joint_autoleave.json
```diff
@@ -0,0 +1,96 @@
+{
```
BusyJay, line 1: Json is too verbose and not readable as plain text. 29 line of codes are translated into 96, 661 bytes to 2.11KB. 🤣

BusyJay:
Hi, @accelsao what's the progress? Can this be continued?

accelsao:
> Hi, @accelsao what's the progress? Can this be continued?

Not yet start it, because I am quite busy. But I will continue this.

PR #388, opened 2020-07-22
port cockroachdb datadriven

accelsao:
## TODO
None

## difference between [cockroachdb/datadriven](https://github.com/cockroachdb/datadriven)
1. no rewrite
2. no subtest
3. no skip command (command must exist)
> they are not use in raft test
4. automatic run `testdata` folder
5. no scan (left the works in custom test function)

## parsing rules
`key=String`, `value=Vec<String> `
- argument 
  - key=argument,value=[]
- argument=
  - key=argument,value=[]
- argument=a
  - key=argument,value=[a]
- argument=a,b,c
  - key=argument,value=[a,b,c]
- argument=(a,b,c)
  - key=argument,value=[a,b,c]
- a,b,c
  - key=a,b,c,value=[]

## testdata format
```
<command> [arg | arg=val | arg=(val1, val2, ...)]...
<input to the command>
----
<expected results>
<more expected results>
<blank line>
```
```
<command> [arg | arg=val | arg=(val1, val2, ...)]...
<input to the command>
----
----
<optional blank line>
<expected results>
<optional blank line>
<more expected results>
<optional blank line>
----
----
<blank line>  (This blank line is not necessary, but add it to be the same as above, and also cockroach implement it for the case rewrite)
```


datadriven/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.31"
+regex = "1.3.9"
+lazy_static = "1.4.0"
+log = "0.4.11"
+env_logger = "0.7.1"
+difference = "2.0.0"
```
BusyJay, line 15: Better leave a new line.

datadriven/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.31"
+regex = "1.3.9"
+lazy_static = "1.4.0"
+log = "0.4.11"
+env_logger = "0.7.1"
```
BusyJay, line 14: How about using slog just like the raft crate?

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,217 @@
...
+use anyhow::Result;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+/// The main function to run tests
+///
+/// You need to pass the path of `testdata` where store the test cases, and your function
+/// to output expected result.
+///
+/// It will be compare with crate `difference`
```
BusyJay, line 12: s/compare/compared/

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,217 @@
...
+///
+/// It invokes a data-driven test. The test cases are contained in a
+/// separate test file and are dynamically loaded, parsed, and executed by this
+/// testing framework. By convention, test files are typically located in a
+/// sub-directory called "testdata". Each test file has the following format:
+///
+/// ```txt
+/// <command> [arg | arg=val | arg=(val1, val2, ...)]...
```
BusyJay, line 20: The original version is `<command>[,<command>...]`.

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,217 @@
...
+/// ----
+/// ----
+/// <expected results>
+///
+/// <more expected results>
+/// ----
+/// ----
+/// ````
+/// Note: blank line is required after the second *`----`*.
+pub fn run_test<F>(path: &str, f: F) -> Result<()>
```
BusyJay, line 41: Missing comment for `f`.

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,217 @@
...
+/// <input to the command>
+/// ----
+/// ----
+/// <expected results>
+///
+/// <more expected results>
+/// ----
+/// ----
+/// ````
+/// Note: blank line is required after the second *`----`*.
```
BusyJay, line 40: Do you mean that when using two `---`, there has to be at least one blank line in expected result?

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,217 @@
...
+pub fn run_test<F>(path: &str, f: F) -> Result<()>
+where
+    F: FnOnce(&TestData) -> String + Copy,
+{
+    let entries: Vec<PathBuf> = fs::read_dir(path)?
```
BusyJay, line 45: So test data has to be stored in a directory?

datadriven/src/lib.rs
```diff
@@ -0,0 +1,41 @@
+/*! # datadriven
+**datadriven** is a tool for testing. ported from [cockroachdb/datadriven](https://github.com/cockroachdb/datadriven)
+
+To execute data-driven tests, pass the path of the test file as well as a
+function which can interpret and execute whatever commands are present in
+the test file. The framework invokes the function, passing it information
+about the test case in a TestData struct.
+
+The function must returns the actual results of the case, which
```
BusyJay, line 9: s/returns/return/

datadriven/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+run_test() compares with the expected results. If the two are not
+equal, the test is marked to fail.
+
+`run_test()` will test all the file in `testdata` folder.
+If you put it under `src/`, the path to need to pass is `src/testdata`
+
+The comparision is done by [difference](https://docs.rs/difference/2.0.0/difference/)
+
+The difference between [cockroachdb/datadriven](https://github.com/cockroachdb/datadriven)
+1. no rewrite
```
BusyJay, line 19: Are these two features used in raft test?

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,217 @@
...
+// instead of returned because the testing module implements t.Skip
+// and t.Fatal using panics, and we're not guaranteed to get back to
+// the caller via a return in those cases.
+fn run_directive<F>(r: &TestDataReader, f: F) -> Result<()>
+where
+    F: FnOnce(&TestData) -> String,
+{
+    let d = r.get_data();
+    let actual = f(&d);
+    debug!("actual: {:?}", actual);
```
BusyJay, line 81: This seems unnecessary as they will be printed at L84.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::test_data::TestData;
+use anyhow::Result;
+use std::path::{Path, PathBuf};
+// use for writeln! macro
+use std::fmt::Write;
+
+pub struct TestDataReader<'a> {
+    source_name: PathBuf,
+    data: TestData,
+    scanner: LineScanner<'a>,
```
BusyJay, line 12: Seems the struct can be replaced by `Enumerate<Lines<'a>>`.

datadriven/src/datadriven.rs
accelsao: I think second command is for the case `subtest`, so I skip this, for confchange and quorum test, they are not neccessary

datadriven/src/datadriven.rs
accelsao: yes, and it is also true when using one separator.

datadriven/src/datadriven.rs
accelsao: yes, the same function is used in the directory, and run all of them.

datadriven/src/datadriven.rs
BusyJay: I don't get it. The docs say when using one separator, blank line is not allow. Can you give an example about the illegal case described by " blank line is required after the second *`----`*."?

datadriven/src/datadriven.rs
BusyJay: How about providing two functions, one for directory, one for single file? So users don't have to put test data into a directory.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,151 @@
...
+        P: AsRef<Path>,
+    {
+        Self {
+            source_name: source_name.as_ref().to_path_buf(),
+            scanner: content.lines().enumerate(),
+            data: TestData::default(),
+        }
+    }
+
+    pub fn next(&mut self, logger: &slog::Logger) -> Result<bool> {
```
BusyJay, line 28: How about setting the logger as a field of `TestDataReader`?

datadriven/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+anyhow = "1.0.31"
+regex = "1.3.9"
+lazy_static = "1.4.0"
+slog = "2.5.2"
+slog-term = "2.6.0"
+slog-async = "2.5.0"
+log = "0.4.11"
+env_logger = "0.7.1"
```
BusyJay, line 17: This seems not used.

datadriven/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.31"
+regex = "1.3.9"
+lazy_static = "1.4.0"
+slog = "2.5.2"
+slog-term = "2.6.0"
+slog-async = "2.5.0"
+log = "0.4.11"
```
BusyJay, line 16: Unnecessary when using slog.

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,259 @@
...
+//
+// The stopNow and subTestSkipped booleans are modified by-reference
+// instead of returned because the testing module implements t.Skip
+// and t.Fatal using panics, and we're not guaranteed to get back to
+// the caller via a return in those cases.
+fn run_directive<F>(r: &TestDataReader, f: F) -> Result<()>
+where
+    F: FnOnce(&TestData) -> String,
+{
+    let d = r.data.clone();
```
BusyJay, line 117: Why clone?

datadriven/src/test_data.rs
```diff
@@ -0,0 +1,106 @@
+use std::fmt;
+
+/// CmdArg contains information about an argument on the directive line. An
+/// argument is specified in one of the following forms:
+/// - arg
+///     - key: arg
```
BusyJay, line 6: This is confusing. I thought `key: arg` is a part of the form. I think following should self-explained:
```
- key
- key=
- key=a # single_value
- key=a,b,c # multiple value
- key=(a,b,c) # same as above.
```

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    pub fn next(&mut self) -> Result<bool> {
+        loop {
+            let line = self.scanner.next();
+            if line.is_none() {
+                return Ok(false);
+            }
+
+            self.data = TestData::default();
```
BusyJay, line 37: Can be moved to L51 to reduce initialization.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,153 @@
...
+            self.data.pos = format!(
+                "{} : L{}",
+                self.source_name.as_path().display(),
+                line.unwrap().0
```
BusyJay, line 41: Should be `line.unwrap().0 + 1`.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,153 @@
...
+            if line.starts_with('#') || line.is_empty() {
+                // Skip comment lines.
+                continue;
+            }
+
+            // Support wrapping directive lines using \, for example:
+            // build-scalar \
+            // vars(int)
+            while line.ends_with('\\') {
+                line = line.trim_end_matches('\\').to_string();
```
BusyJay, line 56: This is wrong. For example, "\\\\" will be trimmed as empty string.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,153 @@
...
+
+                let l = self
+                    .scanner
+                    .next()
+                    .expect("expected argument that is not end with '\\'")
```
BusyJay, line 61: It doesn't panic in the original implement.

datadriven/src/test_data_reader.rs
BusyJay: The original implementation points line to the end of continuous line. For example,
```
a \
b \
c
```
The line number is 3 instead of 0.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,115 @@
...
+            bail!("unknown argument format: {}", arg)
+        }
+    }
+
+    Ok((cmd, cmd_args))
+}
+
+lazy_static! {
+    static ref RE: Regex =
+        Regex::new(r"^ *[-a-zA-Z0-9/_,.]+(=[-a-zA-Z0-9_@=+/,.]*|=\([^)]*\)| *)( |$)").unwrap();
```
BusyJay, line 52: I don't think `/` is allowed.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,115 @@
...
+use anyhow::Result;
+use regex::Regex;
+
+// token
+// (1) argument
+// (2) argument=a,b,c
+// (3) argument=
+// (4) argument=(a,b,c,...)
+// (5) a,b,c
+pub fn parse_line(line: &str, logger: &slog::Logger) -> Result<(String, Vec<CmdArg>)> {
```
BusyJay, line 11: How about keeping their original name? So it's easier to review.

datadriven/src/line_sparser.rs
accelsao: They have
 ```var splitDirectivesRE = regexp.MustCompile(`^ *[-a-zA-Z0-9/_,\.]+(|=[-a-zA-Z0-9_@=+/,\.]*|=\([^)]*\))( |$)`)```,
 however I think `a-z` `A-Z` `_` are enough.

datadriven/src/test_data.rs
accelsao: Can I delete argument type `- a,b,c`, I don't think we need this.
Only the above will remain.

datadriven/src/line_sparser.rs
BusyJay: Turned out I had compared with an old version of the file, which is used by etcd. `/` is allowed actually.

But why allow multiple space as indicated by `| *`?

datadriven/src/test_data.rs
BusyJay: LGTM

datadriven/src/line_sparser.rs
accelsao: That is a workaround to catch type `key`, because `rust` can't catch `empty`, so I use white space instead

BusyJay: I think it's because you remove the first `|`. If you keep the original go regex expression, it should match pure key correctly.

datadriven/src/test_data_reader.rs
accelsao: This will happen when the testdata is not complete, or there is no argument end without `\`.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,121 @@
+use crate::test_data::CmdArg;
+use anyhow::Result;
+use regex::Regex;
+
+// Token
+// (1) argument (no value)
+// (2) argument= (no value)
+// (3) argument=a (single value)
+// (4) argument=a,b,c (multiple value)
+// (5) argument=(a,b,c,...) (multiple value)
```
Fullstop000, line 10: How about `argument=()`

datadriven/src/datadriven.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+    Ok(())
+}
+
+// run_directive runs just one directive in the input.
+//
+// The stopNow and subTestSkipped booleans are modified by-reference
+// instead of returned because the testing module implements t.Skip
+// and t.Fatal using panics, and we're not guaranteed to get back to
+// the caller via a return in those cases.
```
Fullstop000, line 115: I think these can be removed

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,159 @@
...
+            if line.starts_with('#') || line.is_empty() {
+                // Skip comment lines.
+                continue;
+            }
+
+            // Support wrapping directive lines using \, for example:
+            // build-scalar \
+            // vars(int)
+            while line.ends_with('\\') {
+                line = line[..line.len() - 1].to_string();
```
BusyJay, line 51: Seems `line.pop()` is better.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,159 @@
...
+                    .1
+                    .trim();
+
+                line.push_str(l);
+
+                // We need the last line number of argument
+                pos += 1;
+            }
+
+            line = line.trim().to_string();
```
BusyJay, line 66: I don't think it's necessary.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,121 @@
...
+//
+// parse_line parses a line of datadriven input language and returns
+// the parsed command and CmdArgs.
+pub fn parse_line(line: &str, logger: &slog::Logger) -> Result<(String, Vec<CmdArg>)> {
+    let fields = split_directives(line)?;
+    if fields.is_empty() {
+        bail!("empty lines occurs, unexpected.");
```
BusyJay, line 17: The original implement is returning empty result.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    debug!(logger, "argument after split: {:?}", fields);
+
+    let cmd = fields[0].clone();
+    let mut cmd_args = vec![];
+
+    for arg in fields[1..].iter() {
```
BusyJay, line 25: `&fields[1..]` is a better style.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,121 @@
...
+        let vals: Vec<String> = arg.split_terminator('=').map(|v| v.to_string()).collect();
```
BusyJay, line 26: go implement allows `=` in value.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,159 @@
...
+                }
+                loop {
+                    let mut line = self.scanner.next().unwrap().1.trim().to_string();
+                    if line.is_empty() {
+                        break;
+                    } else {
+                        writeln!(&mut line).unwrap();
+                        self.data.expected.push_str(line.as_str());
+                    }
+                }
```
Fullstop000, line 155: To keep consisitent with the original go implementation. Maybe we can use a similar approach like this?
```rust
loop {
   if l.is_empty() {
        break;
   }
   writeln!(&mut l).unwrap();
   self.data.expected.push_str(l.as_str());
   l = self.scanner.next().unwrap().1.trim().to_string();
}
```

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,159 @@
...
+        // (1) second separator
+        // (2) non empty output
+
+        // We need to add an end line for each line in order to distinguish the exact output,
+        // otherwise we may consider an unexpected output combination to be correct
+
+        if let Some((_, line)) = self.scanner.next() {
+            if line == "----" {
+                loop {
+                    let mut line = self.scanner.next().unwrap().1.trim().to_string();
```
Fullstop000, line 119: Is `unwrap()` safe here?

datadriven/src/line_sparser.rs
accelsao: But we get empty fields only when there is no command and arguments, which is `nothing` in test cases.
Maybe we just disallow this condition?

datadriven/src/test_data_reader.rs
accelsao: We consider an empty line as the terminator, so we must panic if not. I will add a message.

accelsao:
`empty line` or `blank line`, which is better?

datadriven/src/line_sparser.rs
BusyJay: I suggest to follow the behavior of go.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    debug!(logger, "argument after split: {:?}", fields);
+
+    let cmd = fields[0].clone();
+    let mut cmd_args = vec![];
+
+    for arg in &fields[1..] {
+        let key_value = arg
+            .replacen("=", " ", 1)
```
BusyJay, line 28: Why not `arg.splitn(2, '=')`?

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,136 @@
...
+                    val.pop();
+                    // remove '('
+                    val.remove(0);
+                    // trim because white space is allow.
+                    vals = val
+                        .split_terminator(',')
+                        .map(|v| v.trim().to_string())
+                        .collect();
+                } else {
+                    vals = val.split_terminator(',').map(|v| v.to_string()).collect();
```
BusyJay, line 56: Let's treat "a,b,c" as single value to keep it simple.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,136 @@
...
+                let vals: Vec<String>;
+
+                if val.starts_with('(') && val.ends_with(')') {
+                    // remove ')'
+                    val.pop();
+                    // remove '('
+                    val.remove(0);
+                    // trim because white space is allow.
+                    vals = val
+                        .split_terminator(',')
```
BusyJay, line 52: This is wrong. "," should be split into `["", ""]` instead of `[""]`.

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,136 @@
...
+    debug!(logger, "argument after split: {:?}", fields);
+
+    let cmd = fields[0].clone();
+    let mut cmd_args = vec![];
+
+    for arg in &fields[1..] {
+        let key_value = arg
+            .replacen("=", " ", 1)
+            .split_terminator(' ')
+            .map(|v| v.to_string())
```
BusyJay, line 30: In rust, we tend to allocate only when it's necessary. From what I can see, following code can happily accept `&str` and iterator` instead of `String` and `Vec`.

BusyJay:
Blank line is better, which can contains space or tab.

datadriven/src/line_sparser.rs
accelsao: Do you mean changing `vals` to String instead of Vec<String>?

BusyJay: What I mean is
```
// (1) argument (no value)
// (2) argument= (no value)
// (3) argument=() (no value)
// (4) argument=a (single value)
// (5) argument=a,b,c (single value)
// (6) argument=(a,b,c,...) (multiple value)
```

accelsao: when computing `sum` a=1,2,3, we will get "1,2,3" as value, while if we get `sum` b=(2,3,4), we will get ["2","3","4"].
We will need to add additional code to consider which types are they.
Is this expected?

accelsao: The former, we need to split with a comma and parse the number, the latter we just need to parse the number

BusyJay: I'm not sure if I understand you completely. If users want to use multiple values,

BusyJay: Actually I am not more in favor of either one. The principle is following the upstream convention.

datadriven/src/test_data.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub key: String,
+    /// Values of CmdArg
+    pub vals: Vec<String>,
+}
+
+impl fmt::Display for CmdArg {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.vals.len() {
+            // wrap key with `` for readability (key might contains ',')
```
BusyJay, line 24: I suggest to remove `'` for readability. In most case there will be no ','.

datadriven/src/test_data.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// output.
+    /// This field is provided so that a test can perform an early return
+    /// with "return d.expected" to signal that nothing has changed.
+    pub expected: String,
+}
+
+impl TestData {
+    /// Return `true` if the `cmd_args` contains a value for the specified key.
+    pub fn contains_key(&self, k: &str) -> bool {
+        for cmd_arg in self.cmd_args.iter() {
```
BusyJay, line 64: `&self.cmd_args` is prefered.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,156 @@
...
+                line.pop();
+
+                let l = self
+                    .scanner
+                    .next()
+                    .expect("expect argument ends without '\\'")
+                    .1
+                    .trim();
+
+                line.push_str(l);
```
BusyJay, line 60: Should append a space first.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,156 @@
...
+                    bail!("testdata is not complete, we expected '----' after command");
+                }
+
+                let line = line.unwrap().1;
+
+                if line == "----" {
+                    self.read_expected();
+                    return Ok(true);
+                } else if !line.is_empty() {
+                    bail!("expected '----', found '{}'", line);
```
BusyJay, line 98: It seems all lines are just ignored in go implement.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,156 @@
...
+                        if line2 == "----" {
+                            let line3 = self.scanner.next().unwrap().1.trim().to_string();
+                            assert!(
+                                line3.is_empty(),
+                                "we expected an empty line after second separator, found '{}'",
+                                line3
+                            );
+                            break;
+                        }
+                        if !line2.is_empty() {
```
BusyJay, line 129: Why check if empty?

datadriven/src/line_sparser.rs
```diff
@@ -0,0 +1,136 @@
...
+
+lazy_static! {
+    static ref RE: Regex =
+        Regex::new(r"^ *[-a-zA-Z0-9/_,.]+(|=[-a-zA-Z0-9_@=+/,.]*|=\([^)]*\))( |$)").unwrap();
+}
+
+fn split_directives(line: &str) -> Result<Vec<String>> {
+    let mut res = vec![];
+
+    let mut line = line.to_string();
```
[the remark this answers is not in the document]
accelsao: I had no idea how to assign back to line with `&str`

`L83: line = last.to_string();`

datadriven/src/test_data_reader.rs
accelsao: That is to skip blank line, only add "\n" to others.
Currently, I skip the blank line to expected value.
e.g.
```

a=1
```
`expected string` = "a=1\n"

If we don't skip blank line, 
we will expect to get `expected string`  = "\na=1\n".

datadriven/src/line_sparser.rs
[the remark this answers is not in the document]
BusyJay: Just `line = last`?

datadriven/src/test_data_reader.rs
BusyJay: What's the golang behavior?

datadriven/src/line_sparser.rs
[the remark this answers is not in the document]
accelsao: I found it, I need `let mut line = line;`

datadriven/src/test_data_reader.rs
accelsao: I think they do the latter, I will follow it.

datadriven/src/test_data.rs
```diff
@@ -0,0 +1,95 @@
+use std::fmt;
+
+/// CmdArg contains information about an argument on the directive line. An
+/// argument is specified in one of the following forms:
+///
+/// - key         (no value)
+/// - key=        (no value)
+/// - key=a       (single value)
+/// - key=a,b,c   (multiple value)
```
BusyJay, line 9: Comment needs to be updated.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,166 @@
...
+
+                if line == "----" {
+                    separator = true;
+                    break;
+                } else {
+                    buf.push_str(line);
+                }
+            }
+
+            self.data.input = buf;
```
BusyJay, line 109: May need trim.

datadriven/src/test_data_reader.rs
```diff
@@ -0,0 +1,166 @@
...
+                    break;
+                }
+
+                let line = line.unwrap().1;
+
+                if line == "----" {
+                    separator = true;
+                    break;
+                } else {
+                    buf.push_str(line.trim());
```
BusyJay, line 105: I don't think it's the same as trimming at L109. Besides, it's expected to add a new line here.

accelsao: go impl is really strange, that only the first line and the last line will trim whitespace,
I will follow it...

BusyJay:
@Fullstop000 @NingLin-P PTAL

Fullstop000:
LGTM

BusyJay:
Well done! Thanks @accelsao @Fullstop000

PR #389, opened 2020-07-22
*: port raft raw node tests

BusyJay:
Added two test cases to check if joint consensus is implemented correctly.

### DIFF
--- a/harness/tests/integration_cases/test_raw_node.rs
+++ b/harness/tests/integration_cases/test_raw_node.rs
@@ -19,6 +19,7 @@ use protobuf::{Message as PbMessage, ProtobufEnum as _};
 use raft::eraftpb::*;
 use raft::storage::MemStorage;
 use raft::*;
+use raft_proto::*;
 use slog::Logger;
 
 use crate::test_util::*;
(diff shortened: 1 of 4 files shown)


harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -154,50 +156,278 @@ fn test_raw_node_read_index_to_old_leader() {
...
+        new_conf_change_single(3, ConfChangeType::AddLearnerNode),
+    ]);
+    cc.set_transition(ConfChangeTransition::Implicit);
+    let cs = conf_state_v2(vec![2], vec![3], vec![1], vec![1], true);
+    test_cases.push((Box::new(cc), cs, Some(conf_state(vec![2], vec![1, 3]))));
+
+    for (cc, exp, exp2) in test_cases {
+        let s = new_storage();
+        let mut raw_node = new_raw_node(1, vec![1], 10, 1, s.clone(), &l);
+        raw_node.campaign().expect("");
```
NingLin-P, line 238: Instead of `expect("")`, could we just use `unwrap()`?

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -154,50 +156,278 @@ fn test_raw_node_read_index_to_old_leader() {
...
-            ccdata.reserve_exact(cc.compute_size() as usize);
-            cc.write_to_vec(&mut ccdata).unwrap();
-            raw_node.propose_conf_change(vec![], cc).expect("");
+                // Force it step down.
+                let mut msg = new_message(1, 1, MessageType::MsgHeartbeatResponse, 0);
+                msg.term = raw_node.raft.term + 1;
+                raw_node.step(msg).unwrap();
 
-            proposed = true;
+                cs = Some(raw_node.apply_conf_change(&cc).unwrap());
```
NingLin-P, line 372: Here, the leader step down first then enter joint, should we also test that a leader enter joint then step down?

src/raft.rs
```diff
@@ -823,7 +823,7 @@ impl<T: Storage> Raft<T> {
 
         // TODO: it may never auto_leave if leader steps down before enter joint is applied.
         if self.prs.conf().auto_leave
-            && old_applied < self.pending_conf_index
+            && old_applied <= self.pending_conf_index
```
NingLin-P, line 826: Why changing this?

harness/tests/integration_cases/test_raw_node.rs
BusyJay: Enter joint then step down is the same as L198~L203. In normal flow, an empty entry is proposed immediately when entering joint state.

src/raft.rs
BusyJay: See etcd-io/etcd#12137

PR #391, opened 2020-08-09
datadriven: add rewrite mode

accelsao:
1. add rewrite
2. don't trim on expected value (bug)
3. move the test to independent file


BusyJay:
Is rewrite mode used in the raft tests?

accelsao:
This is like a tool, not really necessary for testing.
If we change the `Display` in the future, we can just modify the test by hand.
I forget we need group commit so I skip this feature at the beginning.

datadriven/src/testdata/multiline
```diff
@@ -0,0 +1,20 @@
...
+
+large
+----
+----
+more
+than
+five
+lines
+of
+
```
BusyJay, line 17: There is no blank line in the original test data.

datadriven/src/test_data_reader.rs
```diff
@@ -10,10 +10,12 @@ pub struct TestDataReader<'a> {
     pub data: TestData,
     scanner: Enumerate<Lines<'a>>,
     pub logger: slog::Logger,
+    pub rewrite: bool,
+    pub rewrite_buffer: String,
```
BusyJay, line 14: `Option<String>` seems a more safe definition. People can forget to check the Boolean flag.

datadriven/src/test_data_reader.rs
```diff
@@ -34,6 +38,7 @@ impl<'a> TestDataReader<'a> {
 
             let mut pos = line.unwrap().0;
             let mut line = line.unwrap().1.to_string();
+            self.emit(&line);
```
BusyJay, line 41: I think you need to trim after emitting. For example, both of following two lines are comments and should be ignored.
```
# comment
   # comment
```

datadriven/src/test_data_reader.rs
```diff
@@ -134,42 +139,66 @@ impl<'a> TestDataReader<'a> {
...
                         }
-                        line2 += "\n";
-                        self.data.expected.push_str(&line2);
+
+                        break;
                     }
-                    line += "\n";
-                    self.data.expected.push_str(&line);
+                    line2 += "\n";
+                    self.data.expected.push_str(&line2);
```
BusyJay, line 175: This seems wrong. Original implement push `line` first and then `line2`.

datadriven/src/test_data.rs
```diff
@@ -22,7 +22,21 @@ impl fmt::Display for CmdArg {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.vals.len() {
             0 => write!(f, "{}", self.key),
-            _ => write!(f, "{}={:?}", self.key, self.vals),
+            1 => {
+                let mut str = format!("{}=", self.key);
```
BusyJay, line 26: Why not just `write!(f, "{}={}", self.key, self.vals[0])`?

datadriven/src/test_data.rs
```diff
@@ -22,7 +22,21 @@ impl fmt::Display for CmdArg {
...
+                str += &self.vals[0];
+                write!(f, "{}", str)
+            }
+            _ => {
+                let mut str = format!("{}=(", self.key);
```
BusyJay, line 31: Why not `write!(f, "{}=({})", self.key, self.vals.join(", "))`?

datadriven/src/testdata/multiline
BusyJay: Any updates?

datadriven/src/datadriven.rs
```diff
@@ -71,233 +78,102 @@ use std::path::{Path, PathBuf};
...
+    let mut r = TestDataReader::new(source_name, content, rewrite, logger);
+
     while r.next()? {
-        run_directive(&r, f)?;
+        run_directive(&mut r, f)?;
+    }
+
+    if r.rewrite {
+        // remove redundant '\n'
+        let data = r.rewrite_buffer.map(|rb| rb.trim_end().to_string() + "\n");
```
BusyJay, line 126: How about:
```
r.rewrite_buffer.map(|mut rb| {
    if rb.ends_with("\n\n") {
        rb.pop();
    }
    rb
}
```

datadriven/src/test_data.rs
accelsao: Cool!

BusyJay:
@NingLin-P @Fullstop000 PTAL

NingLin-P:
LGTM

datadriven/src/datadriven.rs
```diff
@@ -71,233 +78,105 @@ use std::path::{Path, PathBuf};
...
-        run_directive(&r, f)?;
+        run_directive(&mut r, f)?;
     }
-    Ok(())
+
+    // remove redundant '\n'
+    let data = r.rewrite_buffer.map(|mut rb| {
+        if rb.ends_with("\n\n") {
+            rb.pop();
+            rb.pop();
```
BusyJay, line 128: Why? I can see go implement will still keep one newline.

accelsao: There is no "\n" at the end of rewrite data, so I need to pop 2 "\n"

accelsao: the addition "\n" is the reason why I pop 2 "\n"

accelsao: I am not sure why they dont have any blank line at the end

PR #394, opened 2020-08-26
quorum: port single group joint commit

accelsao:

### DIFF
--- a/src/quorum.rs
+++ b/src/quorum.rs
@@ -4,7 +4,6 @@ pub mod datadriven_test;
 pub mod joint;
 pub mod majority;
 
-use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::fmt::{self, Debug, Display, Formatter};
 
@@ -21,28 +20,10 @@ pub enum VoteResult {
 }
 
 /// Index is a Raft log position.
-#[derive(Default, Clone, Copy, Eq)]
+#[derive(Default, Clone, Copy)]
 pub struct Index {
     pub index: u64,
     pub group_id: u64,
-}
-
-impl PartialEq for Index {
-    fn eq(&self, other: &Self) -> bool {
-        self.index == other.index
-    }
-}
-
-impl PartialOrd for Index {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.index.cmp(&other.index))
-    }
-}
-
-impl Ord for Index {
-    fn cmp(&self, other: &Self) -> Ordering {
-        self.index.cmp(&other.index)
-    }
 }
 
 impl Display for Index {
--- a/src/quorum.rs
+++ b/src/quorum.rs
@@ -48,10 +48,15 @@ impl Ord for Index {
 impl Display for Index {
     #[inline]
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        if self.index != u64::MAX {
-            write!(f, "[{}]{}", self.group_id, self.index)
-        } else {
-            write!(f, "[{}]∞", self.group_id)
+        match self.group_id {
+            0 => match self.index {
+                u64::MAX => write!(f, "∞"),
+                index => write!(f, "{}", index),
+            },
+            group_id => match self.index {
+                u64::MAX => write!(f, "[{}]∞", group_id),
+                index => write!(f, "[{}]{}", group_id, index),
+            },
         }
     }
 }
--- a/src/quorum.rs
+++ b/src/quorum.rs
@@ -1,8 +1,10 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
-
+#[cfg(test)]
+pub mod datadriven_test;
 pub mod joint;
 pub mod majority;
 
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::fmt::{self, Debug, Display, Formatter};
 
@@ -19,10 +21,28 @@ pub enum VoteResult {
 }
 
 /// Index is a Raft log position.
-#[derive(Default, Clone, Copy)]
+#[derive(Default, Clone, Copy, Eq)]
 pub struct Index {
     pub index: u64,
     pub group_id: u64,
+}
+
+impl PartialEq for Index {
+    fn eq(&self, other: &Self) -> bool {
+        self.index == other.index
+    }
+}
+
+impl PartialOrd for Index {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.index.cmp(&other.index))
+    }
+}
+
+impl Ord for Index {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.index.cmp(&other.index)
+    }
 }
 
 impl Display for Index {
--- a/src/quorum/datadriven_test.rs
+++ b/src/quorum/datadriven_test.rs
@@ -88,7 +88,7 @@ fn test_quorum(data: &TestData) -> String {
     };
 
     // verify length of voters
-    let input = idxs.clone().len();
+    let input = idxs.len();
     let voters = JointConfig::new_joint_from_majorities(c.clone(), cj.clone())
         .ids()
         .len();
--- a/src/quorum/datadriven_test.rs
+++ b/src/quorum/datadriven_test.rs
@@ -1,0 +1,142 @@
+use crate::quorum::{AckIndexer, Index};
+use crate::{default_logger, HashSet, JointConfig, MajorityConfig};
+use datadriven::{run_test, TestData};
+
+#[test]
+fn test_data_driven_quorum() -> anyhow::Result<()> {
+    let logger = default_logger();
+    fn test_quorum(data: &TestData) -> String {
+        // Two majority configs. The first one is always used (though it may
+        // be empty) and the second one is used iff joint is true.
+        let mut joint = false;
+
+        // The committed indexes for the nodes in the config in the order in
+        // which they appear in (ids,idsj), without repetition. An underscore
+        // denotes an omission (i.e. no information for this voter); this is
+        // different from 0. For example,
+        //
+        // cfg=(1,2) cfgj=(2,3,4) idxs=(_,5,_,7) initializes the idx for voter 2
+        // to 5 and that for voter 4 to 7 (and no others).
+        //
+        let mut ids: Vec<u64> = Vec::new();
+        let mut idsj: Vec<u64> = Vec::new();
+
+        // indexs
+        let mut idxs: Vec<Index> = Vec::new();
+
+        for arg in &data.cmd_args {
+            for val in &arg.vals {
+                match arg.key.as_str() {
+                    "cfg" => {
+                        let n: u64 = val.parse().expect("type of n should be u64");
+                        ids.push(n);
+                    }
+                    "cfgj" => {
+                        joint = true;
+
+                        if val == "zero" {
+                            assert_eq!(arg.vals.len(), 1, "cannot mix 'zero' into configuration")
+                        } else {
+                            let n: u64 = val.parse().expect("type of n should be u64");
+                            idsj.push(n);
+                        }
+                    }
+                    "idx" => {
+                        let mut n: u64 = 0;
+                        if val != "_" {
+                            n = val.parse().expect("type of n should be u64");
+                            if n == 0 {
+                                panic!("use '_' as 0, check {}", data.pos)
+                            }
+                        }
+                        idxs.push(Index {
+                            index: n,
+                            group_id: 0,
+                        });
+                    }
+                    _ => {
+                        panic!("unknown arg: {}", arg.key);
+                    }
+                }
+            }
+        }
+
+        let ids_set: HashSet<u64> = ids.iter().cloned().collect();
+        let idsj_set: HashSet<u64> = idsj.iter().cloned().collect();
+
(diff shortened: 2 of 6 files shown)


src/quorum/testdata/joint_commit.txt
```diff
@@ -0,0 +1,481 @@
...
+# by the framework, so we don't dwell on it more).
+#
+# Note that by specifying cfgj explicitly we tell the test harness to treat the
+# input as a joint quorum and not a majority quorum. If we didn't specify
+# cfgj=zero the test would pass just the same, but it wouldn't be exercising the
+# joint quorum path.
+committed cfg=(1,2,3) cfgj=zero idx=(100,101,99)
+----
+       idx
+x>       [0]100 (id=1)
```
BusyJay, line 12: Is it group id? How about hiding 0 by default? So that the file is the same as etcd.

src/quorum/majority.rs
```diff
@@ -130,6 +131,66 @@ impl Configuration {
             VoteResult::Lost
         }
     }
+
+    /// Describe returns a (multi-line) representation of the commit indexes for the
+    /// given lookuper.
+    pub fn describe(&self, l: &impl AckedIndexer) -> String {
```
Fullstop000, line 137: Prefer to add a sample output

src/quorum/datadriven_test.rs
```diff
@@ -0,0 +1,142 @@
+use crate::quorum::{AckIndexer, Index};
+use crate::{default_logger, HashSet, JointConfig, MajorityConfig};
+use datadriven::{run_test, TestData};
+
+#[test]
+fn test_data_driven_quorum() -> anyhow::Result<()> {
```
Fullstop000, line 6: Is the returning value necessary?

accelsao: Yes, we return lots of `Result` in `run_test`, such as `read_to_string`,  `OpenOptions::new()`

src/quorum/testdata/joint_commit.txt
accelsao: I set the width larger for the space of group id

BusyJay: Can they be removed?

accelsao: > Can they be removed?

Which one? space, group id or this describe?

src/quorum/majority.rs
```diff
@@ -130,6 +131,78 @@ impl Configuration {
...
+    ///
+    /// e.g.
+    /// ```txt
+    ///             idx
+    /// x>          100 (id=1)
+    /// xx>         101 (id=2)
+    /// >            99 (id=3)
+    /// 100
+    /// ```
+    pub fn describe(&self, l: &impl AckedIndexer) -> String {
```
NingLin-P, line 149: Could you add an unit test for it?

NingLin-P (approved):
LGTM, thanks!

Fullstop000 (approved):
LGTM

accelsao:
@BusyJay PTAL, thanks

src/quorum/testdata/joint_commit.txt
BusyJay: Space. Better make the file identical with etcd's.

src/quorum/majority.rs
```diff
@@ -147,3 +224,123 @@ impl DerefMut for Configuration {
...
+            Index {
+                index: 120,
+                group_id: 1,
+            },
+        );
+        assert_eq!(
+            r#"       idx
+?        0    (id=1)
+x>     120    (id=2)
+>    [1]120    (id=3)
```
accelsao, line 326: Because the width in etcd is 5, so this cant be align if the width of id if over 5 (including group id information)

accelsao:
@BusyJay PTAL, thanks

src/quorum/joint.rs
```diff
@@ -23,6 +23,11 @@ impl Configuration {
         }
     }
 
+    /// Creates a new configuration using the given MajorityConfigs.
+    pub fn new_joint_from_majorities(incoming: MajorityConfig, outgoing: MajorityConfig) -> Self {
```
BusyJay, line 27: Should be `pub(crate)`.

src/quorum/majority.rs
```diff
@@ -130,6 +131,82 @@ impl Configuration {
...
+    ///
+    /// e.g.
+    /// ```txt
+    ///             idx
+    /// x>          100 (id=1)
+    /// xx>         101 (id=2)
+    /// >            99 (id=3)
+    /// 100
+    /// ```
+    pub fn describe(&self, l: &impl AckedIndexer) -> String {
```
BusyJay, line 149: Should be `pub(crate)`.

src/quorum/majority.rs
```diff
@@ -130,6 +131,82 @@ impl Configuration {
...
+            idx: Option<Index>,
+            // length of bar displayed for this Tup
+            bar: usize,
+        }
+
+        // Below, populate .bar so that the i-th largest commit index has bar i (we
+        // plot this as sort of a progress bar). The actual code is a bit more
+        // complicated and also makes sure that equal index => equal bar.
+
+        let mut info = vec![];
```
BusyJay, line 166: You can use `Vec::with_capacity(n)`.

src/quorum/majority.rs
```diff
@@ -130,6 +131,82 @@ impl Configuration {
...
+
+        for id in self.raw_slice() {
```
BusyJay, line 168: Why not `for id in &self.voters`?

src/quorum/majority.rs
```diff
@@ -130,6 +131,82 @@ impl Configuration {
...
+            let idx = l.acked_index(id);
+            info.push(Tup { id, idx, bar: 0 })
+        }
+
+        info.sort_by(|a, b| match a.idx.cmp(&b.idx) {
```
BusyJay, line 173: You mean `(a.idx, a.id).cmp(&(b.idx, b.id))`?

src/quorum/majority.rs
```diff
@@ -130,6 +131,82 @@ impl Configuration {
...
+
+        info.sort_by(|a, b| a.id.cmp(&b.id));
+
+        let mut buf = String::new();
+        buf.push_str(&(" ".repeat(n) + format!("  {:>5}\n", "idx").as_str()));
+
+        for tup in info {
+            let string;
+            if let Some(idx) = tup.idx {
+                string = "x".repeat(tup.bar)
```
BusyJay, line 192: How about using `buf` directly instead of creating a temporary string?

src/quorum/majority.rs
```diff
@@ -130,6 +131,82 @@ impl Configuration {
...
+                    + ">"
+                    + " ".repeat(n - tup.bar).as_str()
+                    + format!(" {:>5}    (id={})\n", format!("{}", idx), tup.id).as_str();
```
BusyJay, line 195: Redundant code, can be extracted as `write!(buf, " {:>5}    (id={})\n", tup.idx.unwrap_or_default(), tup.id).unwrap()`.

accelsao: I still use branch to determine the prefix of the string

src/quorum/joint.rs
```diff
@@ -80,4 +88,10 @@ impl Configuration {
     pub fn contains(&self, id: u64) -> bool {
         self.incoming.contains(&id) || self.outgoing.contains(&id)
     }
+
+    /// Describe returns a (multi-line) representation of the commit indexes for the
+    /// given lookuper.
+    pub fn describe(&self, l: &impl AckedIndexer) -> String {
```
BusyJay, line 94: Should be `pub(crate)`.

src/quorum/majority.rs
```diff
@@ -130,6 +131,78 @@ impl Configuration {
...
+        info.sort_by(|a, b| a.id.cmp(&b.id));
+
+        let mut buf = String::new();
+        write!(buf, " ".repeat(n) + "    idx\n");
+
+        for tup in info {
+            match tup.idx {
+                Some(idx) => write!(
+                    buf,
+                    "x".repeat(tup.bar)
```
BusyJay, line 190: Don't mix `write!` and `+`. Just use `write!` is easy to read.

src/quorum.rs
```diff
@@ -19,19 +21,42 @@ pub enum VoteResult {
...
 
 /// Index is a Raft log position.
-#[derive(Default, Clone, Copy)]
+#[derive(Default, Clone, Copy, Eq)]
 pub struct Index {
     pub index: u64,
     pub group_id: u64,
 }
 
+impl PartialEq for Index {
```
BusyJay, line 30: It doesn't feel right to implement `PartialEq` and its family traits for the `Index`. As the same index with different group id should not be treat as the same `Index`. The comparison rule is limited to context and should not apply universally.

accelsao: Maybe remove Ord and PartialOrd?

src/quorum/majority.rs
accelsao: I fail to get width with 
`writeln!(buf, " {:>5} (id={})", idx, tup.id)
                        .expect("Error occurred while trying to write in String");
                ` and 
`writeln!(buf, " {:>5} (id={})", Index::default(), tup.id)
                        .expect("Error occurred while trying to write in String");
                `

accelsao: It works if passing `format("{}", idx)`

BusyJay: Can you talk more about the error you meet? Or you can show me a picture in slack.

accelsao: the following is the result with this code `writeln!(buf, " {:>5}    (id={})", idx, tup.id)`, 
the green one is the expected output.

accelsao: with this code `writeln!(buf, " {:>5}    (id={})", format!("{}", idx), tup.id)`, I can get the green output

accelsao: It seems the problem may be in `Display`
https://stackoverflow.com/questions/56826464/why-is-the-width-ignored-for-my-custom-formatter-implementation

src/quorum/datadriven_test.rs
```diff
@@ -0,0 +1,142 @@
...
+    let logger = default_logger();
+    fn test_quorum(data: &TestData) -> String {
+        // Two majority configs. The first one is always used (though it may
+        // be empty) and the second one is used iff joint is true.
+        let mut joint = false;
+
+        // The committed indexes for the nodes in the config in the order in
```
BusyJay, line 13: I think the comment is for `idxs`.

src/quorum/datadriven_test.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        // Build the two majority configs.
+        let c = MajorityConfig::new(ids_set);
+        let cj = MajorityConfig::new(idsj_set);
+
+        let make_lookuper = |idxs: &[Index], ids: &[u64], idsj: &[u64]| -> AckIndexer {
+            let mut l = AckIndexer::default();
+            // next to consume from idxs
+            let mut p: usize = 0;
+            let mut ids = ids.to_vec();
```
BusyJay, line 75: How about `for id in ids.iter().chain(idsj)`?

src/quorum/datadriven_test.rs
```diff
@@ -0,0 +1,142 @@
...
+            // them in the lookuper because "no entry" is different from
+            // "zero entry". Note that we prevent tests from specifying
+            // zero commit indexes, so that there's no confusion between
+            // the two concepts.
+            l.retain(|_, index| index.index > 0);
+            l
+        };
+
+        let input = idxs.clone();
+        let voters: HashSet<u64> = JointConfig::new_joint_from_majorities(c.clone(), cj.clone())
```
BusyJay, line 94: `.ids().len()` should be enough.

src/quorum/datadriven_test.rs
```diff
@@ -0,0 +1,142 @@
+use crate::quorum::{AckIndexer, Index};
+use crate::{default_logger, HashSet, JointConfig, MajorityConfig};
+use datadriven::{run_test, TestData};
+
+#[test]
+fn test_data_driven_quorum() -> anyhow::Result<()> {
+    let logger = default_logger();
+    fn test_quorum(data: &TestData) -> String {
```
BusyJay, line 8: Better move it out of function for readability.

accelsao: resolved.

src/quorum/joint.rs
BusyJay: Any updates?

src/quorum/majority.rs
```diff
@@ -130,6 +132,87 @@ impl Configuration {
...
+    ///
+    /// e.g.
+    /// ```txt
+    ///             idx
+    /// x>          100 (id=1)
+    /// xx>         101 (id=2)
+    /// >            99 (id=3)
+    /// 100
+    /// ```
+    pub(crate) fn describe(&self, l: &impl AckedIndexer) -> String {
```
BusyJay, line 150: Perhaps we can mark it with `#[cfg(test)]` so that we don't need another test case to test it.

BusyJay (commented):
Rest LGTM

src/quorum/majority.rs
accelsao: Resolved

BusyJay (approved):

NingLin-P (approved):

PR #395, opened 2020-08-30
Store heartbeat messages in a independent vector

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

Store heartbeat messages in a independent vector so that we do not need to create a `Ready` struct for  raft-group even if there is no entries or propose data in state-machine. And we can send these message at once because there is no `Ready` need to be persisted.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -186,6 +186,8 @@ pub struct Raft<T: Storage> {
 
     /// The list of messages.
     pub msgs: Vec<Message>,
+    /// The list of heartbeat messages, which could be sent before persist Ready.
+    pub heartbeats: Vec<Message>,
     /// Internal raftCore.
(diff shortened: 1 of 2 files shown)


BusyJay:
I think you can do the optimization in application. Application can detect if the ready only contains messages and advance quickly.

maxsitu:
> I think you can do the optimization in application. Application can detect if the ready only contains messages and advance quickly.

Isn't this idea actually exposing the internal logic to the application level? Shouldn't it be better off hidden inside the library?

BusyJay:
> ...internal logic...

Exposing a field to caller instead of going through the ready protocol is exact what you call "exposing internal logic". 😆 

The library is about raft algorithm, we try to make it as clear and simple as possible, which is the key of correctness.

PR #397, opened 2020-09-17
quorum: port majority commit test

accelsao:


src/quorum/datadriven_test.rs
```diff
@@ -123,7 +123,76 @@ fn test_quorum(data: &TestData) -> String {
...
+
+                // Joining a majority with itself should give same result.
+                let a_idx = JointConfig::new_joint_from_majorities(c.clone(), c.clone())
+                    .committed_index(use_group_commit, &l);
+                if a_idx != idx {
+                    buf.push_str(&format!("{} <-- via self-joint quorum\n", a_idx.0));
+                }
+
+                // test overlaying
+                for id in c.raw_slice() {
```
BusyJay, line 147: Use `c.ids().iter()` instead.

src/quorum/datadriven_test.rs
```diff
@@ -123,7 +123,76 @@ fn test_quorum(data: &TestData) -> String {
...
+                    if let Some(iidx) = l.acked_index(id) {
+                        if idx.0 > iidx.index {
```
BusyJay, line 149: Can `iidx.index` be zero?

src/quorum/datadriven_test.rs
```diff
@@ -123,7 +123,76 @@ fn test_quorum(data: &TestData) -> String {
...
+                            // try index - 1
+                            l.insert(
```
BusyJay, line 151: Why? It doesn't match the original implement.

accelsao: L154 is a typo, the original implement create a new `map` with `index - 1`, I think we can just insert a new Index instead of creating a new one.

src/quorum/datadriven_test.rs
accelsao: L86 might removes index that equal zero,
`l.retain(|_, index| index.index > 0);`

BusyJay (approved):

accelsao:
@NingLin-P PTAL, thanks

src/quorum/majority.rs
```diff
@@ -26,6 +27,11 @@ impl Configuration {
         }
     }
 
+    /// Returns an iterator over voters.
+    pub fn ids(&self) -> Iter<'_, u64> {
```
NingLin-P, line 31: Could we use the `raw_slice` method?

src/quorum/datadriven_test.rs
```diff
@@ -123,7 +123,76 @@ fn test_quorum(data: &TestData) -> String {
                 }
                 buf.push_str(&format!("{}\n", idx.0));
             } else {
-                // TODO(accelsao): port majority commit
+                use_group_commit = false;
```
NingLin-P, line 126: Seem this line is unnecessary.

src/quorum/datadriven_test.rs
```diff
@@ -123,7 +123,76 @@ fn test_quorum(data: &TestData) -> String {
...
+                        }
+                    }
+                }
+                buf.push_str(&format!(
+                    "{}\n",
+                    Index {
+                        index: idx.0,
+                        group_id: 0
+                    }
+                ));
```
NingLin-P, line 195: Could we just 
```suggestion
                buf.push_str(&format!("{}\n", idx.0));
```

src/quorum/datadriven_test.rs
```diff
@@ -123,7 +123,76 @@ fn test_quorum(data: &TestData) -> String {
...
+                }
+
+                // Joining a majority with itself should give same result.
+                let a_idx = JointConfig::new_joint_from_majorities(c.clone(), c.clone())
+                    .committed_index(use_group_commit, &l);
+                if a_idx != idx {
+                    buf.push_str(&format!("{} <-- via self-joint quorum\n", a_idx.0));
+                }
+
+                // test overlaying
```
NingLin-P, line 146: Better also port the comment here.
`If the committed index was definitely above the currently inspected idx, the result shouldn't change if we lower it further.`

NingLin-P (commented):
Rest LGTM.

src/quorum/majority.rs
BusyJay: `raw_slice` requires cloning.

src/quorum/datadriven_test.rs
accelsao: When index is MAX,  we need to display `oo`, so we better print in `Index`

NingLin-P: But the joint quorum branch just prints `idx.0` should we unify them?

accelsao: ok, I will unify them.

NingLin-P (approved):
LGTM, thanks!

BusyJay (approved):

PR #398, opened 2020-09-20
*: add protection against unlimited log growth(#131)

### LINKED ISSUE #131: raft: provide protection against unbounded Raft log growth

DanteLee:
- Add protection against uncommitted log growth of leader.
  This protection is done by estimate size of entry data
  rather than count of entry. So entry with no data will
  never be dropped
- Add tests for raft and raw_node
- fix wrong clippy command in CONTRIBUTING.md
- close #131 

Signed-off-by: lyzongyuan <lyzongyuan@gmail.com>

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5092,15 +5092,6 @@ fn test_uncommitted_entries_size_limit() {
 
     nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
 
-    // should return ProposalDropped error
-    {
-        let mut entry = Entry::default();
-        entry.data = b"hello world and raft".to_vec();
-        let long_msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
-        let result = nt.dispatch(vec![long_msg].to_vec());
-        assert_eq!(result.unwrap_err(), raft::Error::ProposalDropped);
-    }
-
     // should return ok
     {
         let result = nt.dispatch(vec![msg.clone()].to_vec());
@@ -5134,6 +5125,29 @@ fn test_uncommitted_entries_size_limit() {
             .reduce_uncommitted_size(&[entry]);
 
         let result = nt.dispatch(vec![msg].to_vec());
+        assert!(result.is_ok());
+    }
+}
+
+#[test]
+fn test_at_least_one_uncommitted_entry_should_be_allowed() {
+    let l = default_logger();
+    let config = &Config {
+        id: 1,
+        max_uncommitted_size: 12,
+        ..Config::default()
+    };
+    let mut nt = Network::new_with_config(vec![None, None, None], config, &l);
+
+    nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
+
+    // at least one entry should be accepted,
+    // even its size is overlimit
+    {
+        let mut entry = Entry::default();
+        entry.data = b"hello world and raft".to_vec();
+        let long_msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
+        let result = nt.dispatch(vec![long_msg].to_vec());
         assert!(result.is_ok());
     }
 }
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5075,3 +5075,95 @@ fn test_read_when_quorum_becomes_less() {
         .unwrap();
     assert!(!network.peers[&1].read_states.is_empty());
 }
+
+#[test]
+fn test_uncommitted_entries_size_limit() {
+    let l = default_logger();
+    let config = &Config {
+        id: 1,
+        max_uncommitted_size: 12,
+        ..Config::default()
+    };
+    let mut nt = Network::new_with_config(vec![None, None, None], config, &l);
+    let data = b"hello world!".to_vec();
+
+    nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
+
+    // should return ProposalDropped error
+    {
+        let mut entry = Entry::default();
+        entry.data = b"hello world and raft".to_vec();
+        let mut msg = Message::default();
+        msg.from = 1;
+        msg.to = 1;
+        msg.set_msg_type(MessageType::MsgPropose);
+        msg.set_entries(vec![entry].into());
+
+        let result = nt.dispatch(vec![msg].to_vec());
+        assert_eq!(result.unwrap_err(), raft::Error::ProposalDropped);
+    }
+
+    // should return ok
+    {
+        let mut entry = Entry::default();
+        entry.data = data.clone();
+        let mut msg = Message::default();
+        msg.from = 1;
+        msg.to = 1;
+        msg.set_msg_type(MessageType::MsgPropose);
+        msg.set_entries(vec![entry].into());
+
+        let result = nt.dispatch(vec![msg].to_vec());
+        assert!(result.is_ok());
+    }
+
+    // then next proposal should be dropped
+    {
+        let mut entry = Entry::default();
+        entry.data = b"!".to_vec();
+        let mut msg = Message::default();
+        msg.from = 1;
+        msg.to = 1;
+        msg.set_msg_type(MessageType::MsgPropose);
+        msg.set_entries(vec![entry].into());
+
+        let result = nt.dispatch(vec![msg].to_vec());
+        assert!(!result.is_ok());
+        assert_eq!(result.unwrap_err(), raft::Error::ProposalDropped);
+    }
+
+    // but entry with empty size should be accepted
+    {
+        let entry = Entry::default();
+        let mut msg = Message::default();
+        msg.from = 1;
+        msg.to = 1;
+        msg.set_msg_type(MessageType::MsgPropose);
+        msg.set_entries(vec![entry].into());
+
+        let result = nt.dispatch(vec![msg].to_vec());
+        assert!(result.is_ok());
+    }
+
+    // after reduce, new proposal should be accecpted
+    {
+        let mut entry = Entry::default();
+        entry.data = data.clone();
+        let mut msg = Message::default();
+        msg.from = 1;
+        msg.to = 1;
+        msg.set_msg_type(MessageType::MsgPropose);
+        msg.set_entries(vec![entry].into());
+
+        // consume entry
+        let mut entry = Entry::default();
+        entry.data = data.clone();
+        nt.peers
+            .get_mut(&1)
+            .unwrap()
+            .reduce_uncommitted_size(&vec![entry]);
+
+        let result = nt.dispatch(vec![msg].to_vec());
+        assert!(result.is_ok());
+    }
+}
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5157,11 +5157,11 @@ fn test_uncommitted_entries_size_limit() {
 
         // consume entry
         let mut entry = Entry::default();
-        entry.data = data.clone();
+        entry.data = data;
         nt.peers
             .get_mut(&1)
             .unwrap()
-            .reduce_uncommitted_size(&vec![entry]);
+            .reduce_uncommitted_size(&[entry]);
 
         let result = nt.dispatch(vec![msg].to_vec());
         assert!(result.is_ok());
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5167,3 +5167,46 @@ fn test_uncommitted_entries_size_limit() {
         assert!(result.is_ok());
     }
 }
+
+#[test]
+fn test_uncommitted_entry_after_leader_election() {
+    let l = default_logger();
+    let config = &Config {
+        id: 1,
+        max_uncommitted_size: 12,
+        ..Config::default()
+    };
+    let mut nt = Network::new_with_config(vec![None, None, None, None, None], config, &l);
+    let data = b"hello world!".to_vec();
+
+    nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
+
+    // create a uncommitted entry on node2
+    nt.cut(1, 3);
+    nt.cut(1, 4);
+    nt.cut(1, 5);
+    {
+        let mut entry = Entry::default();
+        entry.data = data.clone();
+        let mut msg = Message::default();
+        msg.from = 1;
+        msg.to = 1;
+        msg.set_msg_type(MessageType::MsgPropose);
+        msg.set_entries(vec![entry].into());
+
+        nt.send(vec![msg].to_vec());
+    }
+
+    // now isolate master and make node2 as master
+    nt.isolate(1);
+    // ignore message append, cluster only work on election
+    nt.ignore(MessageType::MsgAppend);
+    nt.send(vec![new_message(2, 2, MessageType::MsgHup, 0)]);
+
+    // uncommitted log size should be 12 on node2
+    assert_eq!(nt.peers.get_mut(&2).unwrap().state, raft::StateRole::Leader);
+    assert_eq!(
+        nt.peers.get_mut(&2).unwrap().compute_uncommitted_size(),
+        data.len()
+    );
+}
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5093,63 +5093,44 @@ fn test_uncommitted_entries_size_limit() {
     nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
 
     // should return ok
-    {
-        let result = nt.dispatch(vec![msg.clone()].to_vec());
-        assert!(result.is_ok());
-    }
+    let result = nt.dispatch(vec![msg.clone()].to_vec());
+    assert!(result.is_ok());
 
     // then next proposal should be dropped
-    {
-        let result = nt.dispatch(vec![msg.clone()].to_vec());
-        assert!(!result.is_ok());
-        assert_eq!(result.unwrap_err(), raft::Error::ProposalDropped);
-    }
+    let result = nt.dispatch(vec![msg].to_vec());
+    assert!(!result.is_ok());
+    assert_eq!(result.unwrap_err(), raft::Error::ProposalDropped);
 
     // but entry with empty size should be accepted
-    {
-        let entry = Entry::default();
-        let empty_msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
-        let result = nt.dispatch(vec![empty_msg].to_vec());
-        assert!(result.is_ok());
-    }
+    let entry = Entry::default();
+    let empty_msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
+    let result = nt.dispatch(vec![empty_msg].to_vec());
+    assert!(result.is_ok());
 
     // after reduce, new proposal should be accecpted
-    {
-        // consume entry
-        let mut entry = Entry::default();
-        entry.data = data;
-        entry.index = 3;
-        nt.peers
-            .get_mut(&1)
-            .unwrap()
-            .reduce_uncommitted_size(&[entry]);
+    let mut entry = Entry::default();
+    entry.data = data;
+    entry.index = 3;
+    nt.peers
+        .get_mut(&1)
+        .unwrap()
+        .reduce_uncommitted_size(&[entry]);
+    assert_eq!(nt.peers.get_mut(&1).unwrap().uncommitted_size(), 0);
 
-        let result = nt.dispatch(vec![msg].to_vec());
-        assert!(result.is_ok());
-    }
-}
+    // a huge proposal should be accepted when there is no uncommitted entry,
+    // even it's bigger than max_uncommitted_size
+    let mut entry = Entry::default();
+    entry.data = b"hello world and raft".to_vec();
+    let long_msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
+    let result = nt.dispatch(vec![long_msg].to_vec());
+    assert!(result.is_ok());
 
-#[test]
-fn test_at_least_one_uncommitted_entry_should_be_allowed() {
-    let l = default_logger();
-    let config = &Config {
-        id: 1,
-        max_uncommitted_size: 12,
-        ..Config::default()
-    };
-    let mut nt = Network::new_with_config(vec![None, None, None], config, &l);
-
-    nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
-
-    // at least one entry should be accepted,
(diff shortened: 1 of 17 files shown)


DanteLee:
keep PR up-to-date with master. @NingLin-P PTAL

NingLin-P:
Thank you! @DanteLee It will be helpful to have this protection.

src/raft.rs
```diff
@@ -861,6 +869,7 @@ impl<T: Storage> Raft<T> {
 
         let last_index = self.raft_log.last_index();
         let committed = self.raft_log.committed;
+        self.uncommitted_size = 0;
```
NingLin-P, line 872: After a new leader elected, we should also count the uncommitted log of the previous term or `reduce_uncommitted_size` may panic.

DanteLee: Done, also add a test for this situation, PTAL:)

DanteLee: it seems that [etcd-io/etcd](https://github.com/etcd-io/etcd/blob/master/raft/raft.go#L632) has the same issue? If we don't set this value correctly, raft can not estimate uncommitted size percisely.

eg. 
node1 has 20 bytes of uncommitted data and has 30 for its limit.
But when it becomes leader,  node1's uncommitted size is set to 0.
We can now add 30 more bytes to its log where it should only be 10.

src/raft.rs
```diff
@@ -861,6 +869,7 @@ impl<T: Storage> Raft<T> {
 
         let last_index = self.raft_log.last_index();
         let committed = self.raft_log.committed;
+        self.uncommitted_size = self.compute_uncommitted_size();
```
NingLin-P, line 872: `reset` is called by `become_follower`, `become_candidate` and `become_leader`, yet `uncommitted_size` is only useful for the leader, instead of `reset`, I think `become_leader` will be a better place for this line of code.

src/raft.rs
NingLin-P: Yes, etcd have a [workaround](https://github.com/etcd-io/etcd/blob/cc2b4cd05e6b677c9c73d9ccb7333ebf1e1be7ee/raft/raft.go#L1666-L1670) for this issue.

src/raft.rs
DanteLee: Done, PTAL:)

src/raft.rs
```diff
@@ -2457,4 +2477,73 @@ impl<T: Storage> Raft<T> {
...
+        for entry in ents {
+            size += entry.get_data().len()
+        }
+
+        if size > self.uncommitted_size {
+            fatal!(
+                self.r.logger,
+                "try to reduce uncommitted size less than 0, first index of pending ents is {}",
+                ents[0].get_index()
+            );
```
NingLin-P, line 2502: How about set `self.uncommitted_size` to zero and print a warning log in this case like [etcd](https://github.com/etcd-io/etcd/blob/cc2b4cd05e6b677c9c73d9ccb7333ebf1e1be7ee/raft/raft.go#L1666-L1670)

src/raft.rs
```diff
@@ -2457,4 +2477,73 @@ impl<T: Storage> Raft<T> {
...
+        } else {
+            self.uncommitted_size -= size;
+        }
+    }
+
+    /// Increase size of 'ents' to uncommitted size. Return true when size limit
+    /// is satisfied. Otherwise return false and uncommitted size remains unchanged.
+    pub fn maybe_increase_uncommitted_size(&mut self, ents: &[Entry]) -> bool {
```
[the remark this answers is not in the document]
DanteLee: 'maybe_increase_uncommitted_size' only invoked by leader right now.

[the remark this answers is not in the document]
DanteLee: but for future consideration, adding fast path is a good point

DanteLee:
@NingLin-P Done, PTAL. I'm new to rustlang and the project. Thanks for the comments so far.

src/raw_node.rs
```diff
@@ -263,6 +263,11 @@ impl<T: Storage> RawNode<T> {
         if !rd.read_states.is_empty() {
             self.raft.read_states.clear();
         }
+        // update raft uncommitted entries size
+        if rd.committed_entries.is_some() {
+            self.raft
+                .reduce_uncommitted_size(&rd.committed_entries.unwrap_or_default())
+        }
```
NingLin-P, line 270: Unlike etcd, raft-rs has interfaces like `ready_since` which fetch committed entries in range `(user_provided_index, commit_index]`. In the usage of TiKV, the `user_provided_index` is the last committed entry of the last `Ready` it got [code](https://github.com/tikv/tikv/blob/0b05a9947ca1bc29e090a24a58d8838f1574f9f8/components/raftstore/src/store/peer.rs#L1718). And `reduce_uncommitted_size` for `Ready.committed_entries` here may make  `uncommitted_size` inaccurate, even we compute `uncommitted_size` based on the commit index after a peer become leader, because the `user_provided_index` may less than the commit index

NingLin-P (commented):
Sorry for the delay

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -728,3 +746,55 @@ fn test_set_priority() {
...
+#[test]
+fn test_bounded_uncommitted_entries_growth_with_partition() {
+    let l = default_logger();
+    let config = &Config {
+        id: 1,
+        max_uncommitted_size: 12,
+        ..Config::default()
+    };
+    let s = new_storage();
+    let mut raw_node = new_raw_node_with_config(vec![1], config, s.clone(), &l);
```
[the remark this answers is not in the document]
DanteLee: we can not specify `max_uncommitted_size` through `new_raw_node` unless we make `max_uncommitted_size` public.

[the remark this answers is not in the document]
NingLin-P: Yes, you are right. We can make `new_raw_node` call `new_raw_node_with_config` to reduce redundant code though.

src/raw_node.rs
NingLin-P: A workaround to fix this issue is we compute the `uncommitted_size` again if we meet `uncommitted_size < 0` in `reduce_uncommitted_size`

DanteLee: You mean that when using `ready_since`, user may `advance` a `ready` which has entries that has been included in last `ready`?

NingLin-P: No, for example, a peer is a follower and has log entries: [i1, .., c1, i2, .., c2] 
1. the peer's commit index is c1, so it produces a Ready with entries [i1, .., c1]
2. the peer's commit index forward to c2
3. the peer becomes the leader and set `uncommitted_size` to zero because there are not uncommitted entries
4. the peer advance the previous Ready with entries [i1, .., c1], and call `reduce_uncommitted_size` then find `uncommitted_size` < 0

With the abovementioned workaround, the peer set `uncommitted_size` to the size of entries [i2, .., c2], because next time the peer will call `ready_since(c1)` and produce a Ready with entries [i2, .., c2], and later advance it.

DanteLee: What about use an `apply_hint` to keep track of the commit index when a peer becomes leader?

When `reduce_uncommitted_size`, we only need to reduce the size of entry slice from 
`[max(apply_hint+1, first_entry_index), last_entry_index]` 

eg.
For peer which is leader(normal case):
`apply_hint` should always smaller than `first_entry_index` in any `Ready`;

For peer become leader(which is the case we discuss about):
`apply_hint` is set to `raft_log.committed` when peer becomes leader. For any `Ready` created when peer is follower,
`apply_hint` should be greater than the `last_entry_index` in `Ready`. So the size of slice shall be 0 and current
`uncommitted_size` will not be reduced. 

We can update `apply_hint` in `compute_uncommitted_size`

NingLin-P: Good idea! I have a few suggestions:
- the name `apply_hint` may be ambiguous, I suggest something like `init_committed`
- we can update `apply_hint` in `become_leader`
- we should push `max_uncommitted_size`, `uncommitted_size` and `apply_hint` into a struct to make the logic clear

NingLin-P:
Any update? @DanteLee

DanteLee:
> Any update? @DanteLee

Sorry for the delay. Focus on company these days. But i think i can submit a fix today.

DanteLee:
@NingLin-P Done, PTAL :P

src/raft.rs
```diff
@@ -79,6 +80,77 @@ pub struct SoftState {
...
+        }
+
+        // user may advance a 'Ready' which is generated before this node becomes leader
+        if let Some(ent) = ents.last() {
+            if ent.index <= self.last_committed_index {
+                return false;
+            }
+        }
+
+        let size = ents.iter().fold(0, |acc, ent| acc + ent.get_data().len());
```
NingLin-P, line 143:
```suggestion
        let size = ents.iter().skip_while(|ent| ent.index <= self.last_committed_index).fold(0, |acc, ent| acc + ent.get_data().len());
```

We should compute the `size` of entries with `ent.index > self.last_committed_index`

NingLin-P (commented):
The rest LGTM, thank you! @BusyJay PTAL

src/raft.rs
```diff
@@ -79,6 +80,77 @@ pub struct SoftState {
...
+
+    pub fn maybe_increase_uncommitted_size(&mut self, ents: &[Entry]) -> bool {
+        // fast path
+        if self.is_no_limit() {
+            return true;
+        }
+
+        let size = ents.iter().fold(0, |acc, ent| acc + ent.get_data().len());
+
+        if size + self.uncommitted_size > self.max_uncommitted_size {
```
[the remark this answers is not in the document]
DanteLee: can you explain why this is need?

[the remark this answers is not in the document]
NingLin-P: `max_uncommitted_size` aims to protect against log grow unlimited, but not to be the max size limit of a log entry, there is no reason to drop a proposal if there are not uncommitted entry. (BTW, `max_size_per_msg` also allow at least one log entry)

DanteLee:
@NingLin-P Done ^_^

NingLin-P (dismissed):

DanteLee:
@BusyJay PTAL

src/raft.rs
```diff
@@ -79,6 +80,80 @@ pub struct SoftState {
...
+    uncommitted_size: usize,
+
+    /// Record last committed log index when node becomes leader from candidate.
+    /// See https://github.com/tikv/raft-rs/pull/398#discussion_r502417531 for more detail
+    last_committed_index: u64,
+}
+
+impl UncommittedState {
+    #[inline]
+    pub fn set_uncommitted_size(&mut self, size: usize) {
```
BusyJay, line 100: Since it's a private struct, the method is unnecessary.

src/raft.rs
```diff
@@ -2457,4 +2559,53 @@ impl<T: Storage> Raft<T> {
...
+            );
+        }
+    }
+
+    /// Increase size of 'ents' to uncommitted size. Return true when size limit
+    /// is satisfied. Otherwise return false and uncommitted size remains unchanged.
+    /// For raft with no limit(or non-leader raft), it always return true.
+    pub fn maybe_increase_uncommitted_size(&mut self, ents: &[Entry]) -> bool {
+        // fast path for non-leader endpoint
+        if self.state != StateRole::Leader {
```
BusyJay, line 2586: Because the method will only be called when it's leader, the fast path is actually slow path and causes wrong branch detection. I suggest to remove it or add a `debug_assert`.

src/raft.rs
```diff
@@ -79,6 +80,80 @@ pub struct SoftState {
...
+        self.max_uncommitted_size == NO_LIMIT as usize
+    }
+
+    pub fn maybe_increase_uncommitted_size(&mut self, ents: &[Entry]) -> bool {
+        // fast path
+        if self.is_no_limit() {
+            return true;
+        }
+
+        let size = ents.iter().fold(0, |acc, ent| acc + ent.get_data().len());
```
BusyJay, line 120: A clear method is `ents.iter().map(|e| e.get_data().len()).sum()`.

src/raft.rs
```diff
@@ -79,6 +80,80 @@ pub struct SoftState {
...
+
+        // 1. we should never drop an entry with out any data(eg. leader election)
```
BusyJay, line 122:
```suggestion
        // 1. we should never drop an entry without any data(eg. leader election)
```

src/raft.rs
```diff
@@ -871,9 +954,13 @@ impl<T: Storage> Raft<T> {
         }
     }
 
-    /// Appends a slice of entries to the log. The entries are updated to match
-    /// the current index and term.
-    pub fn append_entry(&mut self, es: &mut [Entry]) {
+    /// Appends a slice of entries to the log.
+    /// The entries are updated to match the current index and term.
+    pub fn append_entry(&mut self, es: &mut [Entry]) -> bool {
```
BusyJay, line 959: Should annotate with `#[must_use]`.

src/raft.rs
```diff
@@ -2457,4 +2559,53 @@ impl<T: Storage> Raft<T> {
...
+
+    /// Compute current uncommitted log size. Only called by leader.
+    pub fn compute_uncommitted_size(&self) -> usize {
+        if self.uncommitted_state.is_no_limit() {
+            return 0;
+        }
+
+        // since uncommitted entries of leader won't be compacted,
+        // any error return by slice should cause panic
+        match self.raft_log.entries(self.raft_log.committed + 1, None) {
```
BusyJay, line 2602: I don't think it's a good idea. We have optimized hard for avoiding reading logs from hard disk when becoming leader. Perhaps we can keep the size limit a soft limit instead of a hard one. /cc @NingLin-P

DanteLee: @NingLin-P @BusyJay just like `etcd`, we can simply set `uncommitted_size` to zero rather than re-computing it.

NingLin-P: We have `last_committed_index` tracing where `uncommitted_size` start counting, as long as they are consistent I think it is always okay, for example:
- `last_committed_index = last_index`, `uncommitted_size = 0`
- `last_committed_index = index of the first unstable log `, `uncommitted_size = unstable log szie`

BusyJay: LGTM

NingLin-P: > We have `last_committed_index` tracing where `uncommitted_size` start counting, as long as they are consistent I think it is always okay, for example:
> 
> * `last_committed_index = last_index`, `uncommitted_size = 0`
> * `last_committed_index = index of the first unstable log `, `uncommitted_size = unstable log szie`

We still need to pick a solution @Dantelee

src/raft.rs
```diff
@@ -834,7 +908,9 @@ impl<T: Storage> Raft<T> {
             // (which registers as zero).
             let mut entry = Entry::default();
             entry.set_entry_type(EntryType::EntryConfChangeV2);
-            self.append_entry(&mut [entry]);
+
+            // append_entry will never refuse an empty
+            let _ = self.append_entry(&mut [entry]);
```
BusyJay, line 913: It should panic if false is returned.

src/raft.rs
```diff
@@ -1043,7 +1133,9 @@ impl<T: Storage> Raft<T> {
         // could be expensive.
         self.pending_conf_index = self.raft_log.last_index();
 
-        self.append_entry(&mut [Entry::default()]);
+        // no need to check result becase append_entry never refuse entries
+        // which size is zero
+        let _ = self.append_entry(&mut [Entry::default()]);
```
BusyJay, line 1138: Should panic too.

src/raft.rs
```diff
@@ -79,6 +80,71 @@ pub struct SoftState {
...
+
+        // user may advance a 'Ready' which is generated before this node becomes leader
+        let size: usize = ents
+            .iter()
+            .skip_while(|ent| ent.index <= self.last_committed_index)
+            .map(|ent| ent.get_data().len())
+            .sum();
+
+        if size > self.uncommitted_size {
+            false
```
BusyJay, line 140: Should set it back to 0.

src/raft.rs
```diff
@@ -1114,9 +1117,8 @@ impl<T: Storage> Raft<T> {
         self.state = StateRole::Leader;
 
         // update uncommitted state
-        let uncommitted_size = self.compute_uncommitted_size();
         let committed_idx = self.raft_log.committed;
-        self.uncommitted_state.uncommitted_size = uncommitted_size;
+        self.uncommitted_state.uncommitted_size = 0;
         self.uncommitted_state.last_committed_index = committed_idx;
```
NingLin-P: With `uncommitted_size` set to zore, `last_committed_index` should set to `self.raft_log.last_index()`, also the 
name `last_committed_index` may need to change

src/raft.rs
```diff
@@ -1761,7 +1857,10 @@ impl<T: Storage> Raft<T> {
                         e.set_entry_type(EntryType::EntryNormal);
                     }
                 }
-                self.append_entry(&mut m.mut_entries());
+                if !self.append_entry(&mut m.mut_entries()) {
+                    // return ProposalDropped when uncommitted size limit is reached
+                    return Err(Error::ProposalDropped);
+                }
```
NingLin-P, line 1863: We may need a debug log here /cc @BusyJay

DanteLee: perhaps we could put the debug log inside append_entry?

NingLin-P: I think put the log  near the `ProposalDropped` error is more clear, in other places where `append_entry` return false, we will just panic so doesn't need a log

NingLin-P (dismissed):

DanteLee:
@BusyJay Friendly ping, PTAL

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -52,12 +52,26 @@ fn must_cmp_ready(
...
     id: u64,
     peers: Vec<u64>,
-    election: usize,
-    heartbeat: usize,
+    election_tick: usize,
+    heartbeat_tick: usize,
+    storage: MemStorage,
+    logger: &Logger,
+) -> RawNode<MemStorage> {
+    let config = &Config {
```
BusyJay, line 60: Should keep `new_test_config`.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -5075,3 +5075,174 @@ fn test_read_when_quorum_becomes_less() {
...
+    let mut nt = Network::new_with_config(vec![None, None, None], config, &l);
+    let data = b"hello world!".to_vec();
+    let mut entry = Entry::default();
+    entry.data = data.to_vec();
+    let msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
+
+    nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
+
+    // should return ok
+    {
```
BusyJay, line 5096: We use blank line instead of brackets to separate block of code.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -5075,3 +5075,174 @@ fn test_read_when_quorum_becomes_less() {
         .unwrap();
     assert!(!network.peers[&1].read_states.is_empty());
 }
+
+#[test]
+fn test_uncommitted_entries_size_limit() {
```
BusyJay, line 5080: How about just follow the test case `TestUncommittedEntryLimit` in etcd for better maintaining in the future?

DanteLee:
@BusyJay PTAL :)

harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -728,3 +737,49 @@ fn test_set_priority() {
         assert_eq!(raw_node.raft.priority, p);
     }
 }
+
+#[test]
+fn test_bounded_uncommitted_entries_growth_with_partition() {
```
BusyJay, line 742: Should include docs.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -5075,3 +5075,151 @@ fn test_read_when_quorum_becomes_less() {
...
+    let long_msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
+    let result = nt.dispatch(vec![long_msg].to_vec());
+    assert!(result.is_ok());
+
+    // but another huge one will be dropped
+    let mut entry = Entry::default();
+    entry.data = b"hello world and raft".to_vec();
+    let long_msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
+    let result = nt.dispatch(vec![long_msg].to_vec());
+    assert!(!result.is_ok());
```
BusyJay, line 5133: Missing empty message case. For example, just removing https://github.com/tikv/raft-rs/pull/398/files#diff-80bb34f30c4e632196d442f13524e8f740c12063031bd3c277060e81c560888cR115 the case can still pass.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -5075,3 +5075,151 @@ fn test_read_when_quorum_becomes_less() {
...
+    entry.data = data;
+    let msg = new_message_with_entries(1, 1, MessageType::MsgPropose, vec![entry]);
+
+    nt.send(vec![new_message(1, 1, MessageType::MsgHup, 0)]);
+
+    // create a uncommitted entry on node2
+    nt.cut(1, 3);
+    nt.cut(1, 4);
+    nt.cut(1, 5);
+    nt.send(vec![msg].to_vec());
```
BusyJay, line 5156: `to_vec` is unnecessary.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -5075,3 +5075,151 @@ fn test_read_when_quorum_becomes_less() {
...
+
+    // now isolate master and make node2 as master
+    nt.isolate(1);
+    // ignore message append, cluster only work on election
+    nt.ignore(MessageType::MsgAppend);
+    nt.send(vec![new_message(2, 2, MessageType::MsgHup, 0)]);
+
+    // uncommitted log size should be 12 on node2
```
BusyJay, line 5164: Wrong comment.

DanteLee:
@BusyJay PTAL

BusyJay (approved):

BusyJay:
@NingLin-P PTAL

NingLin-P (approved):
LGTM, thank you!

PR #401, opened 2020-09-24
quorum: add joint group commit

accelsao:
Signed-off-by: accelsao <jayzhan211@gmail.com>

### DIFF
--- a/src/quorum/testdata/joint_group_commit.txt
+++ b/src/quorum/testdata/joint_group_commit.txt
@@ -1,0 +1,70 @@
+# the same result of joint single group commit
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(1,1,1)
+----
+100
+
+# min(quorum_commit_index = 100, first index that appears in second group = 99)
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(1,1,2)
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 101)
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(2,1,1)
+----
+100
+
+# minimum index = 99
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(_,1,1)
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 99)
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(_,1,2)
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 99)
+group_committed cfg=(1,2,3) cfgj=(1,2,3) idx=(100,101,99) gid=(_,1,2)
+----
+99
+
+# cfg 1 = 101
+# cfg 2 = 100
+group_committed cfg=(1,2,3) cfgj=(4,5,6) idx=(99,100,101,99,100,101) gid=(1,1,2,1,2,1)
+----
+100
+
+# cfg 1 = 101
+# cfg 2 = 99
+group_committed cfg=(1,2,3) cfgj=(4,5,6) idx=(99,100,101,99,100,101) gid=(1,1,2,1,1,_)
+----
+99
+
+# min(quorum_commit_index = 101, first index that appears in second group = 103)
+group_committed cfg=(1,2,3,4,5) cfgj=zero idx=(99,100,101,102,103) gid=(1,1,1,1,2)
+----
+101
+
+# cfg 1 = 1
+# cfg 2 = 2
+group_committed cfg=(1,2,3,4,5) cfgj=(2,3,4,5,6) idx=(1,100,101,102,103,2) gid=(1,_,1,1,1,1)
+----
+1
+
+# cfg 1 = 3
+# cfg 2 = 101
+group_committed cfg=(1,2,3,4,5) cfgj=(2,3,4,5,6) idx=(3,100,101,102,103,2) gid=(_,1,1,1,1,1)
+----
+3
+
+# cfg 1 = 3
+# cfg 2 = 101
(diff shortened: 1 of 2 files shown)


accelsao:
[rfc]
1. if we need to print index with its group information, we need to modify `committed_index` from return `u64` to `Index`.
2. `describe` function doesn't seem to be useful for group commit, just skip it or we need another one? (I prefer to skip it)

BusyJay:
> if we need to print index with its group information, we need to modify committed_index from return u64 to Index

I don't think it's necessary. Group id is meaningless for `committed_index`.

> ...just skip it or we need another one? (I prefer to skip it)

I'm OK with skipping it.

src/quorum/testdata/joint_group_commit.txt
```diff
@@ -0,0 +1,70 @@
...
+----
+100
+
+# minimum index = 99
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(_,1,1)
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 99)
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(_,1,2)
```
BusyJay, line 22: An implementation that just ignores all group id when there is 0 can pass the case.

src/quorum/testdata/joint_group_commit.txt
```diff
@@ -0,0 +1,70 @@
...
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 99)
+group_committed cfg=(1,2,3) cfgj=(1,2,3) idx=(100,101,99) gid=(_,1,2)
+----
+99
+
+# cfg 1 = 101
```
BusyJay, line 31: No, `cfg 1 = 100`.

src/quorum/testdata/joint_group_commit.txt
```diff
@@ -0,0 +1,70 @@
...
+# cfg 2 = 101
+group_committed cfg=(1,2,3,4,5) cfgj=(2,3,4,5,6) idx=(3,100,101,102,103,2) gid=(_,1,1,1,1,1)
+----
+3
+
+# cfg 1 = 3
+# cfg 2 = 101
+group_committed cfg=(1,2,3,4,5) cfgj=(2,3,4,5,6) idx=(3,100,101,102,103,2) gid=(_,1,1,1,3,1)
+----
+101
```
BusyJay, line 64: Why? It should use the minimal.

src/quorum/testdata/joint_group_commit.txt
accelsao: I got it.

src/quorum/testdata/joint_group_commit.txt
```diff
@@ -0,0 +1,76 @@
...
+----
+1
+
+# cfg 1 = 100
+# cfg 2 = 100
+group_committed cfg=(1,2,3) cfgj=(4,5,6) idx=(99,100,101,99,100,101) gid=(1,1,2,1,2,1)
+----
+100
+
+# cfg 1 = 101
```
BusyJay, line 43: `cfg 1 = 100`.

src/quorum/testdata/joint_group_commit.txt
```diff
@@ -0,0 +1,76 @@
...
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(_,1,1)
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 99)
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(_,1,2)
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 98)
```
BusyJay, line 26: Actually it should be minimum commit index because there is just one known group.

BusyJay (dismissed):

accelsao:
@NingLin-P PTAL, thanks.

BusyJay:
@NingLin-P PTAL

src/quorum/testdata/joint_group_commit.txt
```diff
@@ -0,0 +1,76 @@
...
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(1,1,1)
+----
+100
+
+# min(quorum_commit_index = 100, first index that appears in second group = 99)
+group_committed cfg=(1,2,3) cfgj=zero idx=(100,101,99) gid=(1,1,2)
+----
+99
+
+# min(quorum_commit_index = 100, first index that appears in second group = 101)
```
NingLin-P, line 11:
```suggestion
# min(quorum_commit_index = 100, first index that appears in second group = 100)
```

src/quorum/testdata/joint_group_commit.txt
```diff
@@ -0,0 +1,76 @@
...
+98
+
+# cfg 1 = min(quorum_commit_index = 100, first index that appears in second group = 99) 
+# cfg 2 = min(quorum_commit_index = 100, first index that appears in second group = 1) 
+group_committed cfg=(1,2,3,4) cfgj=(3,4,5,6) idx=(101,99,100,102,103,1) gid=(1,_,1,1,_,2)
+----
+1
+
+# cfg 1 = 100
+# cfg 2 = 100
```
NingLin-P, line 38: It will be good to have comment with unified formatted for configurations that have the second group like:
`# minimum index = xx`
or
`# cfg x = min(quorum_commit_index = xx, first index that appears in second group = xx)`

NingLin-P (commented):
So sorry for the delay, the rest LGTM

src/quorum/testdata/joint_group_commit.txt
accelsao: I think the first index is (index: 101, gid: 1), which is different from the quorum (index: 100,  gid: 2)

NingLin-P: Yes, you are right

NingLin-P (approved):

PR #403, opened 2020-10-15
Support asynchronous ready

gengliqi:
## Description
In the past, ready must be used synchronously. It means the application must persist some data in ready and then it can be allowed to get the next ready. 
As https://github.com/etcd-io/etcd/issues/12257 mentioned:

> 1. it can cause high fsync frequency. In the practice of TiKV, we observed the fsync frequency can reach the limit of hardware very easily, which can cause unstable latency and hurt performance. Also high fsync frequency is also expensive in the cloud.
> 2. sync size is unpredictable, which can waste IO when syncing small data size.

This PR aims to support asynchronous ready.

First of all, this PR changes the ready workflow and divides it into two types, i.e. sync and async.

The expected sync ready workflow is

```
1. get ready
2. handle ready
    1. send msg(need not to wait for persisting entries)
    2. apply committed entries(need not to wait for persisting entries) or snapshot
    3. handle read_state, persist hard state and entries
3. if application applies committed entries synchronously, call `advance`, otherwise call `advance_append`
(call `advance_apply_to` after applying). These two functions return LightReady
4. handle LightReady
    1. send msg
    2. apply committed entries
    3. update commit index to hardstate if any
5. call `advance_apply` if applying synchronously
5. get next ready, back to step 1
```

The expected async ready workflow is

```
1. get ready
2. handle ready
    1. send msg(need not to wait for persisting entries)
    2. apply committed_entries(need not to wait for persisting entries) or snapshot
    3. handle read_state, persist hardstate and entries(need not to wait for success)
3. record the number of the ready and call `advance_append_async`
    * if this ready has been persisted synchronously, call `advance_append` instead 
      of `advance_append_async` to get LightReady and then handle it
4. get next ready, back to step 1

If a ready has been persisted, call `on_persist_ready` and pass the number.
```

Actually, not just for the async ready, this PR can also decrease the latency of the sync ready.

```
According to the previous manual, the application must apply committed entries after it has persisted hard state and entries. 
Think of this case:

1. leader gets a new proposal
2. get ready
    1. persist the entries and send append msg to followers
3. leader gets some append response and this proposal can be committed
4. get ready
    1. persist the hard state(for commit index)
    2. apply committed entries

We can find out that leader should wait for two persistences before really applying. 
The latter one is just for persisting a commit index.
Now step 4 can be 
1. apply committed entries
2. persist the hard state(for commit index)
So the latency can be shorter than before.

Note that if this machine crashes before persisting the hard state, the applied index will be greater than the commit index in hard state. 
We will discuss it next.
```

There are some subtle implementation details:
1. The commit index can be smaller than applied index which is impossible in previous implementation. Although commit index persistence has one correctness dependency for conf change(Append Invariant https://github.com/etcd-io/etcd/issues/7625#issuecomment-489232411), the behavior here(i.e. make apply earlier) does not break it. I use the max(hard_state.commit, applied) as the commit index during initialization.
2. The leader forwards `matched` of its progress when a log is proposed. In previous implementation, it's correct because we need to persist entries and then send msg to others. But now it's not correct so the `matched` of leader's progress can be forward after calling `on_persist_ready`.
3. All of the committed entries must be persisted so `persisted` is maintained and it's used for fetching the committed entries(as the upper bound).
4. Remove the `ready_since` and `has_ready_since` and use `commit_since_index` instead. It's also used for fetching the committed entries(as the lower bound).
5. Since `matched` logic is changed and `persisted` is added, some tests are failed.  The method I chose is to add a `persist` function in `Interface` and call it in `Network::send` and some other places.

## Remaining Work
- [x] Update two examples
- [x] Add more tests for async ready
- [ ] Add doc for async ready(I will do this in next PR)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -232,57 +232,7 @@ other nodes:
     }
     ```
 
-2. Check whether `hs` is empty or not. If not empty, it means that the `HardState` of the node has
-changed. For example, the node may vote for a new leader, or the commit index has been increased.
-We must persist the changed `HardState`:
-
-    ```rust
-    # use slog::{Drain, o};
-    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
-    #
-    # let config = Config { id: 1, ..Default::default() };
-    # config.validate().unwrap();
-    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
-    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
-    # let mut node = RawNode::new(&config, store, &logger).unwrap();
-    #
-    # if !node.has_ready() {
-    #   return;
-    # }
-    # let mut ready = node.ready();
-    #
-    if let Some(hs) = ready.hs() {
-        // Raft HardState changed, and we need to persist it.
-        node.mut_store().wl().set_hardstate(hs.clone());
-    }
-    ```
-
-3. Check whether `entries` is empty or not. If not empty, it means that there are newly added
-entries but has not been committed yet, we must append the entries to the Raft log:
-
-    ```rust
-    # use slog::{Drain, o};
-    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
-    #
-    # let config = Config { id: 1, ..Default::default() };
-    # config.validate().unwrap();
-    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
-    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
-    # let mut node = RawNode::new(&config, store, &logger).unwrap();
-    #
-    # if !node.has_ready() {
-    #   return;
-    # }
-    # let mut ready = node.ready();
-    #
-    if !ready.entries().is_empty() {
-        // Append entries to the Raft log
-        node.mut_store().wl().append(ready.entries()).unwrap();
-    }
-
-    ```
-
-4. Check whether `snapshot` is empty or not. If not empty, it means that the Raft node has received
+2. Check whether `snapshot` is empty or not. If not empty, it means that the Raft node has received
 a Raft snapshot from the leader and we must apply the snapshot:
 
     ```rust
@@ -310,7 +260,7 @@ a Raft snapshot from the leader and we must apply the snapshot:
 
     ```
 
-5. Check whether `committed_entires` is empty or not. If not, it means that there are some newly
+3. Check whether `committed_entires` is empty or not. If not, it means that there are some newly
 committed log entries which you must apply to the state machine. Of course, after applying, you
 need to update the applied index and resume `apply` later:
 
@@ -354,8 +304,58 @@ need to update the applied index and resume `apply` later:
     }
     ```
 
+4. Check whether `entries` is empty or not. If not empty, it means that there are newly added
+entries but has not been committed yet, we must append the entries to the Raft log:
+
+    ```rust
+    # use slog::{Drain, o};
+    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
+    #
+    # let config = Config { id: 1, ..Default::default() };
+    # config.validate().unwrap();
+    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
+    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
+    # let mut node = RawNode::new(&config, store, &logger).unwrap();
+    #
+    # if !node.has_ready() {
+    #   return;
+    # }
+    # let mut ready = node.ready();
+    #
+    if !ready.entries().is_empty() {
+        // Append entries to the Raft log
+        node.mut_store().wl().append(ready.entries()).unwrap();
+    }
+
+    ```
+
+5. Check whether `hs` is empty or not. If not empty, it means that the `HardState` of the node has
+changed. For example, the node may vote for a new leader, or the commit index has been increased.
+We must persist the changed `HardState`:
+
+    ```rust
+    # use slog::{Drain, o};
+    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
+    #
+    # let config = Config { id: 1, ..Default::default() };
+    # config.validate().unwrap();
+    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
+    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
+    # let mut node = RawNode::new(&config, store, &logger).unwrap();
+    #
+    # if !node.has_ready() {
+    #   return;
+    # }
+    # let mut ready = node.ready();
+    #
+    if let Some(hs) = ready.hs() {
+        // Raft HardState changed, and we need to persist it.
+        node.mut_store().wl().set_hardstate(hs.clone());
+    }
+    ```
+
 6. Call `advance` to notify that the previous work is completed. Get the return value `PersistLastReadyResult`
-and handle its `messages` and `committed_entries` like step 1 and step 5 does. Then call `advance_apply`
+and handle its `messages` and `committed_entries` like step 1 and step 3 does. Then call `advance_apply`
 to advance the applied index inside.
 
     ```rust
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -251,9 +251,9 @@ entries but has not been committed yet, we must append the entries to the Raft l
     # }
     # let mut ready = node.ready();
     #
-    if !ready.entries().is_empty() {
+    if !ready.entries.is_empty() {
         // Append entries to the Raft log
-        node.mut_store().wl().append(ready.entries()).unwrap();
+        node.mut_store().wl().append(&ready.entries).unwrap();
     }
 
     ```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -251,9 +251,9 @@ entries but has not been committed yet, we must append the entries to the Raft l
     # }
     # let mut ready = node.ready();
     #
-    if !ready.entries.is_empty() {
+    if !ready.entries().is_empty() {
         // Append entries to the Raft log
-        node.mut_store().wl().append(&ready.entries).unwrap();
+        node.mut_store().wl().append(ready.entries()).unwrap();
     }
 
     ```
@@ -307,9 +307,11 @@ messages to the leader after appending the Raft entries:
     if !is_leader {
         // If not leader, the follower needs to reply the messages to
         // the leader after appending Raft entries.
-        let msgs = ready.messages.drain(..);
-        for _msg in msgs {
-            // Send messages to other peers.
+        let msgs = ready.take_messages();
+        for vec_msg in msgs {
+            for _msg in vec_msg {
+                // Send messages to other peers.
+            }
         }
     }
     ```
@@ -339,23 +341,21 @@ need to update the applied index and resume `apply` later:
     # fn handle_normal(e:  raft::eraftpb::Entry) {
     # }
     #
-    if let Some(committed_entries) = ready.committed_entries.take() {
-        let mut _last_apply_index = 0;
-        for entry in committed_entries {
-            // Mostly, you need to save the last apply index to resume applying
-            // after restart. Here we just ignore this because we use a Memory storage.
-            _last_apply_index = entry.index;
+    let mut _last_apply_index = 0;
+    for entry in ready.take_committed_entries() {
+        // Mostly, you need to save the last apply index to resume applying
+        // after restart. Here we just ignore this because we use a Memory storage.
+        _last_apply_index = entry.index;
 
-            if entry.data.is_empty() {
-                // Emtpy entry, when the peer becomes Leader it will send an empty entry.
-                continue;
-            }
+        if entry.data.is_empty() {
+            // Emtpy entry, when the peer becomes Leader it will send an empty entry.
+            continue;
+        }
 
-            match entry.get_entry_type() {
-                EntryType::EntryNormal => handle_normal(entry),
-                EntryType::EntryConfChange => handle_conf_change(entry),
-                EntryType::EntryConfChangeV2 => unimplemented!(),
-            }
+        match entry.get_entry_type() {
+            EntryType::EntryNormal => handle_normal(entry),
+            EntryType::EntryConfChange => handle_conf_change(entry),
+            EntryType::EntryConfChangeV2 => unimplemented!(),
         }
     }
     ```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -354,7 +354,7 @@ We must persist the changed `HardState`:
     }
     ```
 
-6. Call `advance` to notify that the previous work is completed. Get the return value `PersistLastReadyResult`
+6. Call `advance` to notify that the previous work is completed. Get the return value `LightReady`
 and handle its `messages` and `committed_entries` like step 1 and step 3 does. Then call `advance_apply`
 to advance the applied index inside.
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -379,10 +379,10 @@ to advance the applied index inside.
     #
     # fn handle_committed_entries(committed_entries: Vec<Entry>) {
     # }
-    let mut res = node.advance(ready);
+    let mut light_rd = node.advance(ready);
     // Like step 1 and 5, you can use functions to make them behave the same.
-    handle_messages(res.take_messages());
-    handle_committed_entries(res.take_committed_entries());
+    handle_messages(light_rd.take_messages());
+    handle_committed_entries(light_rd.take_committed_entries());
     node.advance_apply();
     ```
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -505,7 +505,7 @@ pub use self::tracker::{Inflights, Progress, ProgressState, ProgressTracker};
 
 #[allow(deprecated)]
 pub use self::raw_node::is_empty_snap;
-pub use self::raw_node::{Peer, RawNode, Ready, SnapshotStatus};
+pub use self::raw_node::{LightReady, Peer, RawNode, Ready, SnapshotStatus};
 pub use self::read_only::{ReadOnlyOption, ReadState};
 pub use self::status::Status;
 pub use self::storage::{RaftState, Storage};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -205,7 +205,84 @@ let mut ready = node.ready();
 The `Ready` state contains quite a bit of information, and you need to check and process them one
 by one:
 
-1. Check whether `snapshot` is empty or not. If not empty, it means that the Raft node has received
+1. Check whether `messages` is empty or not. If not, it means that the node will send messages to
+other nodes:
+
+    ```rust
+    # use slog::{Drain, o};
+    # use raft::{Config, storage::MemStorage, raw_node::RawNode, StateRole};
+    #
+    # let config = Config { id: 1, ..Default::default() };
+    # config.validate().unwrap();
+    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
+    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
+    # let mut node = RawNode::new(&config, store, &logger).unwrap();
+    #
+    # if !node.has_ready() {
+    #   return;
+    # }
+    # let mut ready = node.ready();
+    # let is_leader = node.raft.state == StateRole::Leader;
+    #
+    let msgs = ready.take_messages();
+    for vec_msg in msgs {
+        for _msg in vec_msg {
+            // Send messages to other peers.
+        }
+    }
+    ```
+
+2. Check whether `hs` is empty or not. If not empty, it means that the `HardState` of the node has
+changed. For example, the node may vote for a new leader, or the commit index has been increased.
+We must persist the changed `HardState`:
+
+    ```rust
+    # use slog::{Drain, o};
+    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
+    #
+    # let config = Config { id: 1, ..Default::default() };
+    # config.validate().unwrap();
+    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
+    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
+    # let mut node = RawNode::new(&config, store, &logger).unwrap();
+    #
+    # if !node.has_ready() {
+    #   return;
+    # }
+    # let mut ready = node.ready();
+    #
+    if let Some(hs) = ready.hs() {
+        // Raft HardState changed, and we need to persist it.
+        node.mut_store().wl().set_hardstate(hs.clone());
+    }
+    ```
+
+3. Check whether `entries` is empty or not. If not empty, it means that there are newly added
+entries but has not been committed yet, we must append the entries to the Raft log:
+
+    ```rust
+    # use slog::{Drain, o};
+    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
+    #
+    # let config = Config { id: 1, ..Default::default() };
+    # config.validate().unwrap();
+    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
+    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
+    # let mut node = RawNode::new(&config, store, &logger).unwrap();
+    #
+    # if !node.has_ready() {
+    #   return;
+    # }
+    # let mut ready = node.ready();
+    #
+    if !ready.entries().is_empty() {
+        // Append entries to the Raft log
+        node.mut_store().wl().append(ready.entries()).unwrap();
+    }
+
+    ```
+
+4. Check whether `snapshot` is empty or not. If not empty, it means that the Raft node has received
 a Raft snapshot from the leader and we must apply the snapshot:
 
     ```rust
@@ -231,89 +308,6 @@ a Raft snapshot from the leader and we must apply the snapshot:
             .unwrap();
     }
 
-    ```
-
-2. Check whether `entries` is empty or not. If not empty, it means that there are newly added
-entries but has not been committed yet, we must append the entries to the Raft log:
-
-    ```rust
-    # use slog::{Drain, o};
-    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
-    #
-    # let config = Config { id: 1, ..Default::default() };
-    # config.validate().unwrap();
-    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
-    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
-    # let mut node = RawNode::new(&config, store, &logger).unwrap();
-    #
-    # if !node.has_ready() {
-    #   return;
-    # }
-    # let mut ready = node.ready();
-    #
-    if !ready.entries().is_empty() {
-        // Append entries to the Raft log
-        node.mut_store().wl().append(ready.entries()).unwrap();
-    }
-
-    ```
-
-3. Check whether `hs` is empty or not. If not empty, it means that the `HardState` of the node has
-changed. For example, the node may vote for a new leader, or the commit index has been increased.
-We must persist the changed `HardState`:
-
-    ```rust
-    # use slog::{Drain, o};
-    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
-    #
-    # let config = Config { id: 1, ..Default::default() };
-    # config.validate().unwrap();
-    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
-    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
-    # let mut node = RawNode::new(&config, store, &logger).unwrap();
-    #
-    # if !node.has_ready() {
-    #   return;
-    # }
-    # let mut ready = node.ready();
-    #
-    if let Some(hs) = ready.hs() {
-        // Raft HardState changed, and we need to persist it.
-        node.mut_store().wl().set_hardstate(hs.clone());
-    }
-    ```
-
-4. Check whether `messages` is empty or not. If not, it means that the node will send messages to
-other nodes. There has been an optimization for sending messages: if the node is a leader, this can
-be done together with step 1 in parallel; if the node is not a leader, it needs to reply the
-messages to the leader after appending the Raft entries:
-
-    ```rust
-    # use slog::{Drain, o};
-    # use raft::{Config, storage::MemStorage, raw_node::RawNode, StateRole};
-    #
-    # let config = Config { id: 1, ..Default::default() };
-    # config.validate().unwrap();
-    # let store = MemStorage::new_with_conf_state((vec![1], vec![]));
-    # let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
-    # let mut node = RawNode::new(&config, store, &logger).unwrap();
-    #
-    # if !node.has_ready() {
-    #   return;
-    # }
-    # let mut ready = node.ready();
-    # let is_leader = node.raft.state == StateRole::Leader;
-    #
-    if !is_leader {
-        // If not leader, the follower needs to reply the messages to
-        // the leader after appending Raft entries.
-        let msgs = ready.take_messages();
-        for vec_msg in msgs {
-            for _msg in vec_msg {
-                // Send messages to other peers.
-            }
-        }
-    }
     ```
 
 5. Check whether `committed_entires` is empty or not. If not, it means that there are some newly
@@ -360,11 +354,13 @@ need to update the applied index and resume `apply` later:
     }
     ```
 
-6. Call `advance` to prepare for the next `Ready` state.
+6. Call `advance` to prepare for the next `Ready` state. Get the return value `PersistLastReadyResult`
+and handle its `messages` and `committed_entries` like step 1 and step 5 does.
 
     ```rust
     # use slog::{Drain, o};
-    # use raft::{Config, storage::MemStorage, raw_node::RawNode, eraftpb::EntryType};
+    # use raft::{Config, storage::MemStorage, raw_node::RawNode};
+    # use raft::eraftpb::{EntryType, Entry, Message};
     #
     # let config = Config { id: 1, ..Default::default() };
     # config.validate().unwrap();
@@ -377,7 +373,15 @@ need to update the applied index and resume `apply` later:
     # }
     # let mut ready = node.ready();
     #
-    node.advance(ready);
+    # fn handle_messages(msgs: Vec<Vec<Message>>) {
+    # }
+    #
+    # fn handle_committed_entries(committed_entries: Vec<Entry>) {
+    # }
+    let mut res = node.advance(ready);
+    // Like step 1 and 5, you can use functions to make them behave the same.
+    handle_messages(res.take_messages());
+    handle_committed_entries(res.take_committed_entries());
     ```
 
 For more information, check out an [example](examples/single_mem_node/main.rs#L113-L179).
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -305,7 +305,7 @@ need to update the applied index and resume `apply` later:
     ```
 
 4. Check whether `entries` is empty or not. If not empty, it means that there are newly added
-entries but has not been committed yet, we must append the entries to the Raft log:
+entries but have not been committed yet, we must append the entries to the Raft log:
 
     ```rust
     # use slog::{Drain, o};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -354,8 +354,9 @@ need to update the applied index and resume `apply` later:
     }
     ```
 
-6. Call `advance` to prepare for the next `Ready` state. Get the return value `PersistLastReadyResult`
-and handle its `messages` and `committed_entries` like step 1 and step 5 does.
+6. Call `advance` to notify that the previous work is completed. Get the return value `PersistLastReadyResult`
+and handle its `messages` and `committed_entries` like step 1 and step 5 does. Then call `advance_apply` 
+to advance the applied index inside.
 
     ```rust
     # use slog::{Drain, o};
@@ -382,6 +383,7 @@ and handle its `messages` and `committed_entries` like step 1 and step 5 does.
     // Like step 1 and 5, you can use functions to make them behave the same.
     handle_messages(res.take_messages());
     handle_committed_entries(res.take_committed_entries());
+    node.advance_apply();
     ```
 
 For more information, check out an [example](examples/single_mem_node/main.rs#L113-L179).
--- a/src/log_unstable.rs
+++ b/src/log_unstable.rs
@@ -85,8 +85,9 @@ impl Unstable {
         }
     }
 
-    /// Returns all entries and moves the stable offset up to the last index(if any).
-    pub fn stable_all(&mut self) -> Vec<Entry> {
+    /// Returns the unstable entries and moves the stable offset up to the
+    /// last index, if there is any.
+    pub fn stable_entries(&mut self) -> Vec<Entry> {
         if self.entries.is_empty() {
             return Vec::new();
         }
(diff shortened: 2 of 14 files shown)


src/raft.rs
```diff
@@ -235,7 +235,9 @@ impl<T: Storage> Raft<T> {
     pub fn new(c: &Config, store: T, logger: &Logger) -> Result<Self> {
         c.validate()?;
         let logger = logger.new(o!("raft_id" => c.id));
-        let raft_state = store.initial_state()?;
+        let mut raft_state = store.initial_state()?;
+        // If commit index is smaller than applied index, forward it.
+        raft_state.hard_state.commit = cmp::max(raft_state.hard_state.commit, c.applied);
```
BusyJay, line 240: I don't think it's appropriate to add the assumption. It can break consistency silently if users don't write implementation correctly. Users should guarantee **_only applying committed logs_**. Hence, they should ensure the invariant `applied <= commit`.

src/raft_log.rs
```diff
@@ -354,12 +354,24 @@ impl<T: Storage> RaftLog<T> {
...
-        let offset = cmp::max(since_idx + 1, self.first_index());
-        let committed = self.committed;
-        if committed + 1 > offset {
-            match self.slice(offset, committed + 1, None) {
+    /// Returns any entries between the `since_idx`(if none, use applied)
+    /// and the min(`persisted_idx`(if any), commit index) + 1.
+    pub fn next_entries_between(
+        &self,
+        since_idx: Option<u64>,
+        persisted_idx: Option<u64>,
```
BusyJay, line 362: I suggest to use `u64` instead. So the match is unnecessary.

gengliqi: The function `next_entries` still needs this.

src/raft.rs
BusyJay: > the hard state may be persisted after applying some entries

If users try to do that, then they should be responsible to give a correct commit index. For example, recording the dependent commit index and commit term with the apply index.

> ...raft-rs must depend on some behaviors outside to guarantee correctness.

I don't think raft-rs depends on apply index > commit index.

src/raft_log.rs
BusyJay: `next_entries` can call `self.next_entries_between(self.applied, self.committed + 1)`.

gengliqi: Got it.

src/raft.rs
gengliqi: > If users try to do that, then they should be responsible to give a correct commit index. For example, recording the dependent commit index and commit term with the apply index.

For async ready, commit index may be forwarded after calling `on_persist_last_ready`.  Then the result has committed entries but no new hard state. It's redundant to put hard state in `PersistLastReadyResult` because the only change is `commit` variable.

> I don't think raft-rs depends on apply index > commit index.
Yes. raft-rs does not depend on applied index > commit index.
But it depends on the fact that users must apply the committed entries, which must be satisfied.
If so, it's correct to forward the commit index to applied index during initialization.

BusyJay: > ...It's redundant to put hard state in PersistLastReadyResult because the only change is commit variable.

It's not. Unless you move the commit out of hard state, otherwise it should be updated.

> ...it depends on the fact that users must apply the committed entries, which must be satisfied...

Indeed, so we have to check `commit index >= apply index`.

> ...If so, it's correct to forward the commit index to applied index during initialization.

The problem is only users know if it's so. You can't just assume everyone implement it in this way.

src/raw_node.rs
```diff
@@ -344,92 +380,259 @@ impl<T: Storage> RawNode<T> {
...
+        }
+    }
+
+    /// Notifies that the last ready has been well persisted.
+    ///
+    /// Returns the PersistLastReadyResult that contains committed entries and messages.
+    ///
+    /// Since Ready must be persisted in order, calling this function implicitly means
+    /// all readys collected before have been persisted.
+    pub fn on_persist_last_ready(&mut self) -> PersistLastReadyResult {
```
Connor1996, line 570: As I understand, the async ready convert the process into a pipeline manner with two steps. If there is always getting a new ready(means advancing last ready number), then it hardly can reach the point where the last ready is persisted.

Can we just return `PersistLastReadyResult` for every `on_persist_ready()` and remove `on_persist_last_ready`?

gengliqi: My understanding is that these two functions for different purposes.
The `on_persist_ready` aims at batching the message and committed entries. Users can get a Ready after calling several `on_persist_ready` and `step`.
The `on_persist_last_ready` is a fast path for users. If they have already persisted the last Ready, they can get the `PersistLastReadyResult` without getting `Ready` again.

src/raft.rs
Little-Wallace: We can check `commit-index >= apply-index` in `raft-rs` while change the `commit-index` of `hard-state`  in `initial_state`, which is implemented in TiKV.

src/raft_log.rs
```diff
@@ -354,12 +354,13 @@ impl<T: Storage> RaftLog<T> {
...
-    /// Returns any entries since the a particular index.
-    pub fn next_entries_since(&self, since_idx: u64) -> Option<Vec<Entry>> {
+    /// Returns any entries between max(`since_idx` + 1, first_index)
+    /// and min(`persisted_idx`, committed) + 1.
+    pub fn next_entries_between(&self, since_idx: u64, persisted_idx: u64) -> Option<Vec<Entry>> {
         let offset = cmp::max(since_idx + 1, self.first_index());
-        let committed = self.committed;
-        if committed + 1 > offset {
-            match self.slice(offset, committed + 1, None) {
+        let high = cmp::min(persisted_idx, self.committed) + 1;
```
Little-Wallace, line 361: Could `persisted_idx ` be smaller than `self.committed`?

src/raw_node.rs
```diff
@@ -196,6 +192,63 @@ impl Ready {
...
+    last_entry: Option<(u64, u64)>,
+    snapshot: Snapshot,
+    messages: Vec<Message>,
+}
+
+/// PersistLastReadyResult encapsulates the commit index, committed entries and
+/// messages that are ready to be applied or be sent to other peers.
+#[derive(Default, Debug, PartialEq)]
+pub struct PersistLastReadyResult {
+    commit_index: Option<u64>,
```
Little-Wallace, line 209: Why it could be None?

src/raw_node.rs
```diff
@@ -196,6 +192,63 @@ impl Ready {
...
+    committed_entries: Vec<Entry>,
+    messages: Vec<Vec<Message>>,
+}
+
+impl PersistLastReadyResult {
+    /// The current commit index.
+    /// It will be None state if there is no update.
+    /// It is not required to save it to stable storage.
+    #[inline]
+    pub fn commit_index(&self) -> Option<u64> {
```
Little-Wallace, line 219: I do not think that we shall check whether there are committed entries by `Option`

src/raft_log.rs
gengliqi: Yes. If it is a follower, it may receive some committed entries and they are not persisted in local.

src/raw_node.rs
gengliqi: You can see the `fn commit_index` annotation above.

src/raw_node.rs
gengliqi: Hmmm, it is not used to check whether there are committed entries, it is used to inform the application the `commit index` is changed. 
Actually, it implicitly means that this peer is a leader if the `commit index` is changed after persisting the entries.

gengliqi:
PTAL

gengliqi:
@NingLin-P PTAL too 🤣

src/raw_node.rs
```diff
@@ -349,94 +384,290 @@ impl<T: Storage> RawNode<T> {
...
+        // So it should be called after `next_entries_between`.
+        rd.entries = raft.raft_log.unstable.stable_all();
+        if let Some(e) = rd.entries.last() {
+            // If the last entry exists, the entries must not empty, vice versa.
+            rd.must_sync = true;
+            rd_record.last_entry = Some((e.get_index(), e.get_term()));
+        }
+
+        if !self.messages.is_empty() {
+            rd.messages = mem::take(&mut self.messages);
```
NingLin-P, line 476: `mem::take` will create a default value,  maybe [mem::swap](https://doc.rust-lang.org/std/mem/fn.swap.html) is better

src/raw_node.rs
NingLin-P: `PersistLastReadyResult` seem like the same `Ready` do, can we make `on_persist_last_ready` to be replaced with:
```
self.on_persist_ready(self.max_number);
self.ready()
```

src/raw_node.rs
```diff
@@ -349,94 +384,290 @@ impl<T: Storage> RawNode<T> {
...
+        }
+        if !raft.msgs.is_empty() {
+            if raft.state == StateRole::Leader {
+                // Leader can send messages immediately to make replication concurrently.
+                // For more details, check raft thesis 10.2.1.
+                rd.messages.push(mem::take(&mut raft.msgs));
+            } else {
+                rd_record.messages = mem::take(&mut raft.msgs);
+            }
+        }
```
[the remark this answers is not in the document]
gengliqi: Yes, it's an optimization. I think we can do it further. For example, if this ready has no update, the messages can be added to the last `ReadyRecord` which should be persisted.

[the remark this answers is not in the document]
gengliqi: But I think maybe we can do this optimization later and keep the behavior the same as before.

src/raw_node.rs
```diff
@@ -349,94 +384,290 @@ impl<T: Storage> RawNode<T> {
...
 
-    /// HasReady called when RawNode user need to check if any Ready pending.
-    /// Checking logic in this method should be consistent with Ready.containsUpdates().
-    #[inline]
-    pub fn has_ready(&self) -> bool {
-        self.has_ready_since(None)
+    fn commit_ready(&mut self, rd: Ready) {
+        if rd.ss.is_some() {
+            self.prev_ss = rd.ss.unwrap();
+        }
```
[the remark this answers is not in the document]
gengliqi: It's the original code. 🤣

src/raw_node.rs
gengliqi: 👍

src/raw_node.rs
```diff
@@ -349,94 +384,290 @@ impl<T: Storage> RawNode<T> {
...
+        if !raft.msgs.is_empty() {
+            if raft.state == StateRole::Leader {
+                // Leader can send messages immediately to make replication concurrently.
+                // For more details, check raft thesis 10.2.1.
+                rd.messages.push(mem::take(&mut raft.msgs));
+            } else {
+                rd_record.messages = mem::take(&mut raft.msgs);
+            }
+        }
+        self.records.push_back(rd_record);
```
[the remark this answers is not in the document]
gengliqi: `commit_ready` has a number check for ensuring the order of function call, i.e. should call the `advance`-like function after calling `ready`. I think the cost is little because, in general, users should call `has_ready` before calling `ready`.

src/lib.rs
```diff
@@ -251,16 +251,14 @@ entries but has not been committed yet, we must append the entries to the Raft l
...
+        // Raft HardState changed, and we need to persist it.
+        node.mut_store().wl().set_hardstate(hs.clone());
     }
-
     ```
 
-3. Check whether `hs` is empty or not. If not empty, it means that the `HardState` of the node has
-changed. For example, the node may vote for a new leader, or the commit index has been increased.
-We must persist the changed `HardState`:
+3. Check whether `entries` is empty or not. If not empty, it means that there are newly added
```
BusyJay, line 260: Why handling hardstate first? If they are not written atomically, it's possible commit index in hard state pointing to an missing entry.

src/raft_log.rs
```diff
@@ -354,12 +364,13 @@ impl<T: Storage> RaftLog<T> {
         term > self.last_term() || (term == self.last_term() && last_index >= self.last_index())
     }
 
-    /// Returns any entries since the a particular index.
+    /// Returns any entries between max(`since_idx` + 1, first_index)
```
BusyJay, line 367:
```suggestion
    /// Returns committed and persisted entries since max(`since_idx` + 1, first_index).
```

BusyJay (commented):
In the past, we use `last_index` as `persisted_index`, but the assumption changes in this PR. You need to review every usage of `last_index` and see if it should be replaced with `persisted_index`. For example, in `reset` function, `last_index` is used for updating `matched` directly.

src/raw_node.rs
gengliqi: Indeed, but I think using `PersistLastReadyResult` instead of `ready`  is better because I want to tell users that **only** committed entries and messages need be handled.

src/raw_node.rs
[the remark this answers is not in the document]
NingLin-P: > commit_ready has a number check for ensuring the order of function call, i.e. should call the advance-like function after calling ready

We can do the number check by replacing `Ready::number: u64` with `Ready::record_number: Option<u64>` and only set the `Ready::record_number` to `Some` when there are corresponding `ReadyRecord`, and do the checking:
```
if let Some(number) = rd.record_number {
        let rd_record = self.records.back().unwrap();
        assert!(rd_record.number == number);
}
```

src/raw_node.rs
NingLin-P: >  to tell users that only committed entries and messages are needed to be handled.

I not sure is it better because it forces the user to write some new logic to handle `PersistLastReadyResult` instead of reusing the exiting logic of handling `Ready`

gengliqi:
> In the past, we use `last_index` as `persisted_index`, but the assumption changes in this PR. You need to review every usage of `last_index` and see if it should be replaced with `persisted_index`. For example, in `reset` function, `last_index` is used for updating `matched` directly.

Hmmm, I think the key is `matched` not `last_index`.
Actually, we always use some "future"(i.e. not persisted) information to make decisions(e.g. generating response).
This method can work correctly depends on the fact that the message must be sent after persisting the data.

So for non-leader peers, the behavior is exactly the same as before.
But for the leader, the behavior has some differences because leader can send messages directly.

As you mentioned, the leader's `matched`, it must be the truly persisted index because the leader uses it to calculate commit index and broadcast the message without persisting any data.

> For example, in `reset` function, `last_index` is used for updating `matched` directly.
Actually, it's ok because if this peer becomes leader, it must be candidate before thus it should persist the `term`, `vote` and entries(if any) before sending the vote msg.

Besides leader's `matched`, is another variable used on leader has a similar problem? We can think it twice.

src/lib.rs
gengliqi: I slipped up there. 
Is the order below ok?
```
1. messages
2. committed entries
3. snapshot
4. entries
5. hs
```

BusyJay: Perhaps step 2 and 3 is conflict with each other. Or we should apply snapshot first and then committed entries to get a consistent apply order.

gengliqi: Oh, I think it's ok because now `committed entries` and snapshot is exclusive now. 
https://github.com/tikv/raft-rs/blob/7aa651b9d83bc73221ff094da888753f6bdb5d92/src/raw_node.rs#L438-L454

BusyJay:
> Hmmm, I think the key is `matched` not `last_index`.

The key is all those informations that have not been persisted in disk in more than 1 round of handling. `last_index` is one of them.

> Actually, it's ok because if this peer becomes leader, it must be candidate before thus it should persist the term, vote and entries(if any) before sending the vote msg.

Please add it as comment in the function.

gengliqi:
I find a bug that `persisted` is not updated when getting a snapshot.
I have changed & added more unit tests.
PTAL, thanks! @BusyJay @Connor1996 @Little-Wallace

src/lib.rs
BusyJay: Actually they are not exclusive. For example, when leader sends a snapshot to a follower and then receives a heartbeat response from the same follower, leader will update the progress to probe and will send one entry to follower next time. So a follower is applying snapshot or handling snapshot doesn't mean it has no committed logs.

src/lib.rs
```diff
@@ -277,20 +279,37 @@ We must persist the changed `HardState`:
...
+        }
     }
     ```
 
-4. Check whether `messages` is empty or not. If not, it means that the node will send messages to
-other nodes. There has been an optimization for sending messages: if the node is a leader, this can
-be done together with step 1 in parallel; if the node is not a leader, it needs to reply the
-messages to the leader after appending the Raft entries:
+4. Check whether `entries` is empty or not. If not empty, it means that there are newly added
+entries but has not been committed yet, we must append the entries to the Raft log:
```
BusyJay, line 308:
```suggestion
entries but have not been committed yet, we must append the entries to the Raft log:
```

src/log_unstable.rs
```diff
@@ -85,29 +85,19 @@ impl Unstable {
...
-            return;
-        }
-
-        if t.unwrap() == term && idx >= self.offset {
-            let start = idx + 1 - self.offset;
-            self.entries.drain(..start as usize);
-            self.offset = idx + 1;
+    /// Returns the unstable entries and moves the stable offset up to the
+    /// last index, if there is any.
+    pub fn stable_entries(&mut self) -> Vec<Entry> {
```
BusyJay, line 90: Supposing entries in storage are [(3, 6), (3, 11)], and entries in unstable are [(3, 12), (3, 15)]. After creating a ready, before ready is written to storage, unstable entries become ((3, 16), ∞). At that time a new leader sends conflict entries [(4, 13), (4, 15)], the storage will be queried the term at 13, which doesn't exist yet and then cause panic.

src/log_unstable.rs
```diff
@@ -85,29 +85,19 @@ impl Unstable {
...
-    /// Removes the snapshot from self if the index of the snapshot matches
-    pub fn stable_snap_to(&mut self, idx: u64) {
-        if self.snapshot.is_none() {
-            return;
-        }
-        if idx == self.snapshot.as_ref().unwrap().get_metadata().index {
-            self.snapshot = None;
-        }
+    /// Returns the snapshot and removes it from self.
+    pub fn stable_snap(&mut self) -> Option<Snapshot> {
```
BusyJay, line 99: This can cause the `first_index` go backward in the similar case as `stable_entries`.

src/raft_log.rs
```diff
@@ -272,14 +282,15 @@ impl<T: Storage> RaftLog<T> {
         self.applied
     }
 
-    /// Attempts to set the stable up to a given index.
-    pub fn stable_to(&mut self, idx: u64, term: u64) {
-        self.unstable.stable_to(idx, term)
+    /// Returns the unstable entries and moves the stable offset up to the
+    /// last index, if there is any.
+    pub fn stable_entries(&mut self) -> Vec<Entry> {
```
NingLin-P, line 287: I think we should reuse the previous `stable_to` and `stable_snap_to` functions, these two functions called after entries and/or snapshot did stable, and which means we can get them from the `Storage` interface, so we can remove them from `unstable`. Otherwise, if we call `stable_entries` and/or `stable_snap` and before we actually stable them to `Storage`, any calls (outside of `raft-rs` by the applications) of functions like `last_index`,  `last_term`, etc. will return a false result.

src/raw_node.rs
```diff
@@ -349,94 +384,287 @@ impl<T: Storage> RawNode<T> {
...
+            // left are messages and they can be sent without persisting because no key data changes
+            // (term, vote, entry). These messages should be added before raft.msgs to avoid out of order.
+            for record in self.records.drain(..) {
+                assert_eq!(record.last_entry, None);
+                assert!(!record.has_snapshot);
+                if !record.messages.is_empty() {
+                    self.messages.push(record.messages);
+                }
+            }
+        }
```
NingLin-P, line 420: This may imply that we must call `commit_ready` (by functions like `advance_append` ) to update `self.prev_ss` before getting the next `Ready`, maybe we should put this constraint as a comment on `fn ready`

gengliqi:
I change the `stable_to` to `stable_entries` mainly because the copy of unstable entries is wasteful in the previous implementation.

I thought about the problems you mentioned before.

Users should handle ready and then call `advance` or `advance_append` or `advance_append_async`.
Rule 1: between the `ready` and `advance`-like functions, users should handle the ready only and should not call some functions in `raw_node` like `step`, `propose`. 
Only after calling `advance`-like functions, users are allowed to call some other functions in `raw_node`.
Actually, I think the original design idea for users should be "Only the functions of raw_node can be used". (Rule 2)
So if the users obey this rule like the document and examples, this optimization works well.
 
But for TiKV, we have a lot of special logic and use many variables&functions in `raft` in addition to `raw_node`.
TiKV still obeys rule 1 so the problems of @jay mentioned do not exist.

However, TiKV does not obey Rule 2 during handling the ready.
Like @NingLin-P said, the `last_index` and `last_term` function in `raft_log` may get a wrong answer, it's because the unstable entries are on the road, i.e. do not save to storage. Some data about `raft_log` should not be gotten from raft-rs because it thinks the entries and snapshot in ready have been saved to storage.
I think it's the cost of this optimization.
How do you think about it? @BusyJay @NingLin-P

src/raw_node.rs
gengliqi: Yes. The users must obey this rule. It's mentioned in the document. Let me think about the comments 🤣.

BusyJay:
> TiKV still obeys rule 1 so the problems of @jay mentioned do not exist.

That's not true. The reason why the problem doesn't exit in TiKV is because TiKV has entry caches. Even the logs are not written into rocksdb, raft-rs can still fetch logs from caches.

I would suggest to focus on what we really want to do. What this PR wants to do is to make it possible to handle ready asynchronously, performance improvement is a bonus but not goal. I suggest to keep the old behavior for now and optimize the performance later.

I think keeping changes as small as possible is the key to merge PRs quickly.

gengliqi:
I remove the optimization of copying unstable entries. So the problem mentioned by @NingLin-P will not exist.
But I keep the behavior of `stable_entries` that it just clears the entries because the users should obey Rule 1.
BTW, I think maybe we should shrink this vector of unstable entries otherwise it may occupy too many memory. In the original code, we use `drain` which has the same problem. But which size is better? Maybe 64 or 128? @BusyJay @NingLin-P

src/raft_log.rs
```diff
@@ -1402,4 +1489,7 @@ mod test {
             }
         }
     }
+
+    #[test]
+    fn test_maybe_persist() {}
```
BusyJay, line 1494: What's the test content?

src/raft.rs
```diff
@@ -959,21 +960,33 @@ impl<T: Storage> Raft<T> {
...
             e.term = self.term;
             e.index = li + 1 + i as u64;
         }
-        // use latest "last" index after truncate/append
-        li = self.raft_log.append(es);
+        self.raft_log.append(es);
 
-        let self_id = self.id;
-        self.mut_prs().get_mut(self_id).unwrap().maybe_update(li);
+        // Not move on self's pr.matched until on_persist_entries
```
BusyJay, line 970:
```suggestion
        // Not update self's pr.matched until on_persist_entries
```

src/raw_node.rs
```diff
@@ -204,6 +258,15 @@ pub struct RawNode<T: Storage> {
     pub raft: Raft<T>,
     prev_ss: SoftState,
     prev_hs: HardState,
+    // Current max number of Record and ReadyRecord.
+    max_number: u64,
+    records: VecDeque<ReadyRecord>,
+    // If there is a pending snapshot.
+    pending_snapshot: bool,
```
NingLin-P, line 265: Maybe this flag is unnecessary, IMO, pending snapshot just like pending entries that both waiting for persisting, 
we can handle pending snapshot like how we handle pending entries (stash messages and  `on_persist`-like function)

src/raw_node.rs
```diff
@@ -349,94 +385,299 @@ impl<T: Storage> RawNode<T> {
...
+            number: self.max_number,
+            ..Default::default()
+        };
+        let mut rd_record = ReadyRecord {
+            number: self.max_number,
+            ..Default::default()
+        };
+
+        // If there is a pending snapshot, do not get the whole Ready.
+        if self.pending_snapshot {
```
BusyJay, line 403: Why not check it at the beginning of the function?

src/raw_node.rs
```diff
@@ -349,94 +385,299 @@ impl<T: Storage> RawNode<T> {
...
+    /// Notifies that the ready of this number has been well persisted.
+    ///
+    /// Since Ready must be persisted in order, calling this function implicitly means
+    /// all readys with numbers smaller than this have been persisted.
+    pub fn on_persist_ready(&mut self, number: u64) {
+        while let Some(record) = self.records.front() {
+            if record.number > number {
+                break;
+            }
+            let mut record = self.records.pop_front().unwrap();
```
BusyJay, line 569: Probably should assert `record.number == number`.

src/raw_node.rs
```diff
@@ -349,94 +385,299 @@ impl<T: Storage> RawNode<T> {
...
+    /// Notifies that the last ready has been well persisted.
+    ///
+    /// Returns the PersistLastReadyResult that contains commit index, committed entries
+    /// and messages.
+    ///
+    /// This function must be called before entering the next round(e.g. propose, step).
+    /// Since Ready must be persisted in order, calling this function implicitly means
+    /// all readys collected before have been persisted.
+    pub fn on_persist_last_ready(&mut self) -> PersistLastReadyResult {
+        self.on_persist_ready(self.max_number);
```
NingLin-P, line 595: I think this may also assume that all previous `Ready` has been `commit_ready()` meaning that they all write to the file system (but not `fsync`). It may possible that the user gets a new `Ready` and calling `on_persist_last_ready`  before `commit_ready()` it

src/raft_log.rs
gengliqi: I remove these lines. The test is `test_maybe_persist_with_snap` above.

src/raw_node.rs
gengliqi: Because the `Ready` and its corresponding  `ReadyRecord` should be created.

BusyJay:
> .. because the users should obey Rule 1.

Actually in etcd, `ready` is split into `readyWithoutAccept` and `acceptReady`. So the rule 1 may not be required at all.

src/raw_node.rs
gengliqi: This function means all readys with numbers smaller than this have been persisted. You can see the annotation above this function.

src/raw_node.rs
gengliqi: Yes. But I think we should not worry about it. (I haven't added the doc for async ready yet)
I think maybe we can add some checking code for flow control in order to find users' wrong behavior.

src/raw_node.rs
BusyJay: Why not return `None` when `pending_snapshot` is true?

src/raw_node.rs
BusyJay: I agree with @Connor1996 and @NingLin-P that this extra method make logic complicated and not helpful as it seems.

src/raft_log.rs
```diff
@@ -495,14 +532,23 @@ impl<T: Storage> RaftLog<T> {
             snapshot_index = snapshot.get_metadata().index,
             snapshot_term = snapshot.get_metadata().term,
         );
-        self.committed = snapshot.get_metadata().index;
+        let index = snapshot.get_metadata().index;
+        assert!(index >= self.committed, "{} < {}", index, self.committed);
```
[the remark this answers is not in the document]
gengliqi: Actually, the `Raft::Restore` has the same check `snap.get_metadata().index < self.raft_log.committed`, this code here just for robustness.

src/raw_node.rs
gengliqi: Do you mean `ready` return `Option<Ready>`? I think it's not necessary because users will call `has_ready` in general cases.

src/raft_log.rs
```diff
@@ -495,14 +532,23 @@ impl<T: Storage> RaftLog<T> {
...
+        self.committed = index;
+        // persisted must be less than unstable.offset at any time.
+        if self.persisted > index {
+            self.persisted = index;
```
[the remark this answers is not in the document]
gengliqi: The invariant of `persisted` is `persisted < unstable.offset && applied <= persisted`. So here it should be `index`.(The `unstable.offset` is `index + 1` and the `applied` will be `index` after applying this snapshot.
By the way, I think it twice and remove this `if` statement.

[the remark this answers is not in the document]
BusyJay: I suggested the changes based on its meanings instead of invariant. It's should indicate that any logs before the index have been persisted. Snapshot at `index` is apparently not the case.

src/raw_node.rs
gengliqi: Imagine this case:
1. The application calls several ready and `advance_append_sync`
2. After calling the last ready's `advance_append_sync`, it finds it's time to call fsync
3. If there is no `on_persist_last_ready`, it should call `on_persist_ready` and then call `ready` to get the next ready and the only meaningful things in `Ready` are `committed_entries` and `messages`.

In TiKV, we have lots of logic during handling `Ready` which is very wasteful to get a `ready` again and do the same logic like normal `ready` does just for applying committed entries and sending messages.

src/raw_node.rs
gengliqi: If this peer needs to apply snapshot, it must not belong to quorum members of this raft group.
And the leader will not send the subsequent entries unless there is a new leader and the snapshot has exactly the last entries which the new leader knows, obviously, this case is very rare. So I don't see any advantage to get a `Ready` before applying snapshot.

gengliqi:
PTAL again @BusyJay @NingLin-P

src/raft_log.rs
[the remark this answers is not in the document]
gengliqi: I think about it before. Maybe we should have a better name?
We need `persisted` because the entries need to be persisted first and then they can be applied.
But for snapshot, the behavior of persisting and applying is only one step. That's the difference between them.

src/raw_node.rs
NingLin-P: > If this peer needs to apply snapshot, it must not belong to quorum members of this raft group.

I don't think this is true, for example, a raft group with three voters and one follower is down and another is applying snapshot

> So I don't see any advantage to get a Ready before applying snapshot.

The biggest benefit to remove `pending_snapshot` is to reuse the logic of handling pending entries

src/raft_log.rs
```diff
@@ -354,12 +373,12 @@ impl<T: Storage> RaftLog<T> {
...
     }
 
-    /// Returns any entries since the a particular index.
+    /// Returns committed and persisted entries since max(`since_idx` + 1, first_index).
     pub fn next_entries_since(&self, since_idx: u64) -> Option<Vec<Entry>> {
         let offset = cmp::max(since_idx + 1, self.first_index());
-        let committed = self.committed;
-        if committed + 1 > offset {
-            match self.slice(offset, committed + 1, None) {
+        let high = cmp::min(self.committed, self.persisted) + 1;
```
BusyJay, line 379: Case `self.committed > self.persisted` is not covered in this file.

src/raft_log.rs
```diff
@@ -776,100 +824,108 @@ mod test {
...
-            raft_log.stable_to(stablei, stablet);
-            if raft_log.unstable.offset != wunstable {
-                panic!(
-                    "#{}: unstable = {}, want {}",
-                    i, raft_log.unstable.offset, wunstable
-                );
-            }
-        }
-    }
+    fn test_restore_() {}
```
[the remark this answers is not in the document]
gengliqi: I remove it. The new test is `test_restore_snap` at the end of this file.

src/raft_log.rs
gengliqi: I merge the `test_has_next_ents` and `test_next_ents` and add more cases

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+    /// Advance append the ready value asynchronously.
     #[inline]
-    pub fn advance_append(&mut self, rd: Ready) {
+    pub fn advance_append_async(&mut self, rd: Ready) {
         self.commit_ready(rd);
     }
 
+    /// Advance apply to the index of the last committed entries given before.
+    #[inline]
+    pub fn advance_apply(&mut self) {
```
BusyJay, line 619: Why not just keeping the original interface and implementations?

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+    ///
+    /// Returns the LightReady that contains commit index, committed entries and messages.
+    pub fn advance(&mut self, rd: Ready) -> LightReady {
+        let applied = self.commit_since_index;
+        let light_rd = self.advance_append(rd);
+        self.advance_apply_to(applied);
+        light_rd
+    }
+
+    /// Advance append the ready value synchronously.
```
BusyJay, line 583:
```suggestion
    /// Advances the ready without applying committed entries. `advance_apply` should be used later to update applying progress.
```

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+        assert_eq!(
+            hard_state, self.prev_hs,
+            "hard state {:?} != prev_hs {:?}",
+            hard_state, self.prev_hs
+        );
+        light_rd
     }
 
-    /// Appends and commits the ready value.
+    /// Advance append the ready value asynchronously.
```
BusyJay, line 611:
```suggestion
    /// Same as `advance_append` except that it allows to only store the updates in cache. `on_persist_ready` 
    /// should be used later to update the persisting progress.
    ///
    /// Raft works on an assumption persisted updates should not be lost, which usually requires expensive
    /// operations like `fsync`. `advance_append_async` allows you to control the rate of such operations and
    /// get a reasonable batch size. However, it's still required that the updates can be read by raft from the
    /// `Storage` trait before calling `advance_append_async`.
```

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
         }
+        if term != 0 {
+            self.raft.on_persist_entries(index, term);
+        }
+    }
+
+    /// Advance notifies the RawNode that the application has applied and saved progress
+    /// in the last Ready results.
+    ///
+    /// Returns the LightReady that contains commit index, committed entries and messages.
```
BusyJay, line 575:
```suggestion
    /// Advances the ready after fully processing it.
    ///
    /// Fully processing a ready requires to persist snapshot, entries and hard states, apply all committed entries,
    /// send all messages.
    ///
    /// Returns the LightReady that contains commit index, committed entries and messages. `LightReady`
    /// contains updates that only valid after persisting last ready. It should also be fully processed.
```

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+        if raft.state == StateRole::Leader && !raft.msgs.is_empty() {
+            // Leader can send messages immediately to make replication concurrently.
+            // For more details, check raft thesis 10.2.1.
+            rd.messages.push(mem::take(&mut raft.msgs));
+        }
+
+        rd
     }
 
     /// Ready returns the current point-in-time state of this RawNode.
```
BusyJay, line 410:
```suggestion
    /// Returns the outstanding work that the application needs to handle.
    ///
    /// This includes appending and applying entries or a snapshot, updating the HardState,
    /// and sending messages. The returned `Ready` *must* be handled and subsequently
    /// passed back via advance() or its families.
    ///
    /// `has_ready` should be called first to check if it's necessary to handle the ready.
```

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
-            // until something changes, but we do send any un-applied but
-            // committed entries (and previously-committed entries may be
-            // incorporated into the snapshot, even if rd.CommittedEntries is
-            // empty). Therefore we mark all committed entries as applied
-            // whether they were included in rd.HardState or not.
-            self.advance_apply(commit_idx);
+    /// Notifies that the ready of this number has been well persisted.
+    ///
+    /// Since Ready must be persisted in order, calling this function implicitly means
+    /// all readys with numbers smaller than this one have been persisted.
```
BusyJay, line 549:
```suggestion
    /// Notifies that the ready of this number has been persisted.
    ///
    /// Since Ready must be persisted in order, calling this function implicitly means
    /// all readies with numbers smaller than this one have been persisted.
    ///
    /// `has_ready` and `ready` should be called later to handle further updates that become
    /// valid after ready being persisted.
```

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+        rd
     }
 
-    /// Given an index, can determine if there is a ready state from that time.
-    pub fn has_ready_since(&self, applied_idx: Option<u64>) -> bool {
+    /// HasReady called when RawNode user need to check if any Ready pending.
+    pub fn has_ready(&self) -> bool {
         let raft = &self.raft;
-        if !raft.msgs.is_empty() || raft.raft_log.unstable_entries().is_some() {
+        if raft.soft_state() != self.prev_ss {
```
BusyJay, line 492: `msgs` is a probably a true short path.

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
-        self.has_ready_since(None)
+    fn commit_ready(&mut self, rd: Ready) {
+        if let Some(ss) = rd.ss {
+            self.prev_ss = ss;
+        }
+        if let Some(hs) = rd.hs {
+            self.prev_hs = hs;
+        }
+        let rd_record = self.records.back().unwrap();
+        assert!(rd_record.number == rd.number);
```
BusyJay, line 532: `assert_eq` is better.

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+
+        if !self.messages.is_empty() {
+            mem::swap(&mut rd.messages, &mut self.messages);
+        }
+
+        if raft.state == StateRole::Leader && !raft.msgs.is_empty() {
+            // Leader can send messages immediately to make replication concurrently.
+            // For more details, check raft thesis 10.2.1.
+            rd.messages.push(mem::take(&mut raft.msgs));
+        }
```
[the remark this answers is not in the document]
BusyJay: No, it can also be true after persisting some logs, which can trigger log commitment.

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+            rd.hs = Some(hs);
+        }
+
+        if !raft.read_states.is_empty() {
+            mem::swap(&mut rd.read_states, &mut raft.read_states);
+        }
+
+        if let Some(snapshot) = &raft.raft_log.unstable_snapshot() {
+            rd.snapshot = snapshot.clone();
+            assert!(self.commit_since_index <= rd.snapshot.get_metadata().index);
```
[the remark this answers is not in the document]
BusyJay: If request snapshot is used, `commit_since_index` can be larger than snapshot index.

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+        if hard_state.commit > self.prev_hs.commit {
+            light_rd.commit_index = Some(hard_state.commit);
+            self.prev_hs.commit = hard_state.commit;
+        } else {
+            assert!(hard_state.commit == self.prev_hs.commit);
+            light_rd.commit_index = None;
+        }
+        assert_eq!(
+            hard_state, self.prev_hs,
+            "hard state {:?} != prev_hs {:?}",
```
BusyJay, line 605: `assert_eq` will print the values, so the customized format is redundant.

src/raw_node.rs
gengliqi: If so, what's another function's name which uses `commit_since_index` for sync applying?

src/raw_node.rs
gengliqi: Appending synchronously does not mean applying synchronously. 
Maybe the "`advance_apply` should be used later to update applying progress." can be removed?

src/raw_node.rs
```diff
@@ -349,94 +379,259 @@ impl<T: Storage> RawNode<T> {
...
+    }
+
+    /// Advance notifies the RawNode that the application has applied and saved progress
+    /// in the last Ready results.
+    ///
+    /// Returns the LightReady that contains commit index, committed entries and messages.
+    pub fn advance(&mut self, rd: Ready) -> LightReady {
+        let applied = self.commit_since_index;
+        let light_rd = self.advance_append(rd);
+        self.advance_apply_to(applied);
```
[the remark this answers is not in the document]
gengliqi: They are not the same because the `commit_since_index` maybe updated in `advance_append`.

src/raw_node.rs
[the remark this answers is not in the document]
gengliqi: Maybe this design is a little bit strange? 🤔

src/raw_node.rs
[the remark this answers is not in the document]
gengliqi: Jay is right. You can see the implementation of `Raft::on_persist_entries` for details.

src/raw_node.rs
BusyJay: The comment doesn't mention whether to apply synchronously, it just suggests the advance is split into two step, and apply is excluded from the first step.

src/raw_node.rs
BusyJay: What I mean is just keep one function that accepts explicit index is enough. Application always knows what index has been applied.

src/raw_node.rs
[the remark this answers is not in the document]
BusyJay: The design is for online recovery. Supposing some regions are corrupted while others in the same node are not, using request snapshot allows TiKV to fix those corrupted regions without bringing unavailability to others.

src/raw_node.rs
gengliqi: The function which uses `commit_since_index` is convenient for users. For the previous implementation, users don't need to record the applied index if the logs are applied synchronously. As our examples, this function is used after handling the `LightReady`.

src/raw_node.rs
gengliqi: Got it.

BusyJay (approved):

NingLin-P (approved):

overvenus (approved):

Little-Wallace (approved):
LGTM

PR #409, opened 2020-12-07
Fix examples and snapshot code of MemStorageCore

gengliqi:

Issue: close https://github.com/tikv/raft-rs/issues/408

The commit index of `LightReady` is not handled in examples.
But the test can not pass is mainly due to the bug of generating snapshot in `MemStorageCore`.
The term in snapshot meta should not be the `hard_state.term`. It should be the log term of `meta.index`.

By the way, the snapshot has an assumption that all entries whose index is less than `hard_state.commit` has been applied, which is not true when using async ready. So I leave a TODO and I will resolve it soon.

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -193,7 +193,6 @@ impl MemStorageCore {
     /// Panics if the snapshot index is less than the storage's first index.
     pub fn apply_snapshot(&mut self, mut snapshot: Snapshot) -> Result<()> {
         let mut meta = snapshot.take_metadata();
-        let term = meta.term;
         let index = meta.index;
 
         if self.first_index() > index {
@@ -202,7 +201,6 @@ impl MemStorageCore {
 
         self.snapshot_metadata = meta.clone();
 
-        self.raft_state.hard_state.term = term;
         self.raft_state.hard_state.commit = index;
         self.entries.clear();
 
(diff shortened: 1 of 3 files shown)


gengliqi:
@BusyJay @NingLin-P  PTAL, thanks.

NingLin-P (approved):
LGTM

src/storage.rs
```diff
@@ -214,12 +214,25 @@ impl MemStorageCore {
     fn snapshot(&self) -> Snapshot {
         let mut snapshot = Snapshot::default();
 
-        // Use the latest applied_idx to construct the snapshot.
-        let applied_idx = self.raft_state.hard_state.commit;
-        let term = self.raft_state.hard_state.term;
+        // We assume all entries whose index is less than hard_state.commit
+        // has been applied yet.
```
BusyJay, line 218:
```suggestion
        // has been applied.
```

src/storage.rs
```diff
@@ -214,12 +214,25 @@ impl MemStorageCore {
...
+        // So use the latest commit_idx to construct the snapshot.
```
BusyJay, line 219:
```suggestion
        // We assume all entries whose indexes are less than `hard_state.commit`
        // have been applied, so use the latest `commit_idx` to construct the snapshot.
```

BusyJay (approved):

NingLin-P (approved):

PR #410, opened 2020-12-10
Change the behavior of persisted index for snapshot

gengliqi:

In previous async ready PR(https://github.com/tikv/raft-rs/pull/403), After getting a snapshot and calling the `restore` function, the `persisted` is changed to the index of this snapshot. This makes the `persisted` index does not mean the data before this index is _really_ persisted, which is not intuitive.
The reason for this behavior is to maintain a invariant that `applied <= min(committed, persisted)`. However, I find it's not needed if the application calls the `advance_append` or `on_persist_ready` then calls `advance_apply_to`. This is intuitive because the snapshot is persisted does not mean it is applied so it should be persisted first, then be applied.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -971,7 +971,7 @@ impl<T: Storage> Raft<T> {
         true
     }
 
-    /// Notifies that these raft logs have been well persisted.
+    /// Notifies that these raft logs or snapshot have been persisted.
     pub fn on_persist_entries(&mut self, index: u64, term: u64) {
         let update = self.raft_log.maybe_persist(index, term);
         if update && self.state == StateRole::Leader {
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -974,26 +974,24 @@ impl<T: Storage> Raft<T> {
     /// Notifies that these raft logs or snapshot have been persisted.
     pub fn on_persist_entries(&mut self, index: u64, term: u64) {
         let update = self.raft_log.maybe_persist(index, term);
-        // Actually, if it is a leader and persisted index is updated, this term
-        // must be equal to self.term because the persisted index must be equal to
-        // the last index of entries from previous leader when it becomes leader
-        // (see the comments in become_leader), namely, the new persisted entries
-        // must come from this leader. Here checking the term just for robustness.
         if update && self.state == StateRole::Leader {
-            if term == self.term {
-                let self_id = self.id;
-                let pr = self.mut_prs().get_mut(self_id).unwrap();
-                pr.maybe_update(index);
-                if self.maybe_commit() && self.should_bcast_commit() {
-                    self.bcast_append();
-                }
-            } else {
+            // Actually, if it is a leader and persisted index is updated, this term
+            // must be equal to self.term because the persisted index must be equal to
+            // the last index of entries from previous leader when it becomes leader
+            // (see the comments in become_leader), namely, the new persisted entries
+            // must come from this leader.
+            if term != self.term {
                 error!(
                     self.logger,
-                    "persisted index changed but the term {} is not the same as {}",
+                    "leader's persisted index changed but the term {} is not the same as {}",
                     term,
                     self.term
                 );
+            }
+            let self_id = self.id;
+            let pr = self.mut_prs().get_mut(self_id).unwrap();
+            if pr.maybe_update(index) && self.maybe_commit() && self.should_bcast_commit() {
+                self.bcast_append();
             }
         }
     }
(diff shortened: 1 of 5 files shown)


gengliqi:
@BusyJay @NingLin-P PTAL, thanks.

BusyJay:
So if the `persisted` is 2048 and snapshot reset first index back to 2040, then following new committed and not saved entries will be fetched to ready directly?

gengliqi:
> So if the `persisted` is 2048 and snapshot resets last index back to 2040, then following new committed and not saved entries will be fetched to ready directly?

It won't happen because the `persisted` should be changed when appending new entries.

It seems it's hard to make the `persisted` to be *really persisted index*. Maybe it's no need to change?

BusyJay:
> It won't happen because the persisted should be changed when appending new entries.

It's not true. Entries after snapshot won't cause conflict, so `persisted` won't be changed.

gengliqi:
> > It won't happen because the persisted should be changed when appending new entries.
> 
> It's not true. Entries after snapshot won't cause conflict, so `persisted` won't be changed.

If there is no conflict, the persisted is also be changed.
https://github.com/tikv/raft-rs/blob/fc1ef2f2d9711f7b9e88fe4399445b135d5f8023/src/raft_log.rs#L227-L232

BusyJay:
If there is no conflict, `conflict_idx` will be zero, so the else branch will be skipped.

https://github.com/tikv/raft-rs/blob/fc1ef2f2d9711f7b9e88fe4399445b135d5f8023/src/raft_log.rs#L217-L236

gengliqi:
> If there is no conflict, `conflict_idx` will be zero, so the else branch will be skipped.
> 
> https://github.com/tikv/raft-rs/blob/fc1ef2f2d9711f7b9e88fe4399445b135d5f8023/src/raft_log.rs#L217-L236

Hmmm, if it's zero, there is no new entries.

gengliqi:
By the way, I think we should set `persisted` to make the logic more intuitive, but set it to `index - 1` seems not good enough, I have no better idea.

BusyJay:
How about setting it to zero, which means unknown? Or use `Option<NonZeroU64>` and make it `None`.

gengliqi:
> How about setting it to zero, which means unknown? Or use `Option<NonZeroU64>` and make it `None`.

It has problems because there is an invariant that `applied <= min(committed, persisted)`.

After thinking for a while, I believe the most suitable method is `if self.persisted > self.committed { self.persisted = self.committed; }` so we can say the `persisted` is the **really** persisted index all the time. You can see the comments in code for details.

Also, I add test for the case mentioned before, which is 
1. Append some entries(index 20) and persist them
2. Receive a snapshot(index 10) then some new entries(index to 13)
3. Get a ready then check it(if there is a snapshot and some new entries can be applied, the code should panic in `RawNode::ready` 
```
assert!(
    !raft
        .raft_log
        .has_next_entries_since(self.commit_since_index),
     "has snapshot but also has committed entries since {}",
     self.commit_since_index
);
```

gengliqi:
@BusyJay @NingLin-P PTAL again, thanks.

src/raft.rs
```diff
@@ -971,10 +971,15 @@ impl<T: Storage> Raft<T> {
...
+    /// Notifies that these raft logs or snapshot have been persisted.
     pub fn on_persist_entries(&mut self, index: u64, term: u64) {
         let update = self.raft_log.maybe_persist(index, term);
-        if update && self.state == StateRole::Leader {
+        // Actually, if it is a leader and persisted index is updated, this term
+        // must be equal to self.term because the persisted index must be equal to
+        // the last index of entries from previous leader when it becomes leader
+        // (see the comments in become_leader), namely, the new persisted entries
+        // must come from this leader. Here checking the term just for robustness.
+        if update && self.state == StateRole::Leader && term == self.term {
```
BusyJay, line 982: How about using a log? Even if the term doesn't match `self.term` in the future adaption, for example introducing paging, it's still safe to enter the if branch.

src/raw_node.rs
```diff
@@ -565,6 +569,11 @@ impl<T: Storage> RawNode<T> {
             }
             let mut record = self.records.pop_front().unwrap();
 
+            if let Some((i, t)) = record.snapshot {
+                index = i;
```
BusyJay, line 573: Any case to cover?

BusyJay (commented):
Rest LGTM

NingLin-P (dismissed):
LGTM

src/raft.rs
gengliqi: Done.

src/raw_node.rs
gengliqi: Added by changing `test_async_ready_follower`

BusyJay (approved):

NingLin-P (approved):

PR #411, opened 2020-12-22
*: allow vote to inform log commit

BusyJay:
This is a port of etcd-io/etcd#12435. I made some adaption, notable
changes are:
- Candidate steps down as follower when new commit logs contains conf
change.
- Allow log commit only when a node has no leader.

Those changes are not necessary in theory, but I add them for best
safety.

### DIFF
--- a/harness/tests/test_util/mod.rs
+++ b/harness/tests/test_util/mod.rs
@@ -212,6 +212,10 @@ pub fn conf_state_v2(
     cs
 }
 
+pub fn conf_change(t: ConfChangeType, node_id: u64) -> ConfChange {
+    new_conf_change(t, node_id)
+}
+
 pub fn conf_change_v2(steps: Vec<ConfChangeSingle>) -> ConfChangeV2 {
     let mut cc = ConfChangeV2::default();
     cc.set_changes(steps.into());
--- a/harness/tests/test_util/mod.rs
+++ b/harness/tests/test_util/mod.rs
@@ -172,7 +172,7 @@ pub fn new_snapshot(index: u64, term: u64, voters: Vec<u64>) -> Snapshot {
     s
 }
 
-fn new_conf_change(ty: ConfChangeType, node_id: u64) -> ConfChange {
+pub fn conf_change(ty: ConfChangeType, node_id: u64) -> ConfChange {
     let mut cc = ConfChange::default();
     cc.node_id = node_id;
     cc.set_change_type(ty);
@@ -180,15 +180,15 @@ fn new_conf_change(ty: ConfChangeType, node_id: u64) -> ConfChange {
 }
 
 pub fn remove_node(node_id: u64) -> ConfChangeV2 {
-    new_conf_change(ConfChangeType::RemoveNode, node_id).into_v2()
+    conf_change(ConfChangeType::RemoveNode, node_id).into_v2()
 }
 
 pub fn add_node(node_id: u64) -> ConfChangeV2 {
-    new_conf_change(ConfChangeType::AddNode, node_id).into_v2()
+    conf_change(ConfChangeType::AddNode, node_id).into_v2()
 }
 
 pub fn add_learner(node_id: u64) -> ConfChangeV2 {
-    new_conf_change(ConfChangeType::AddLearnerNode, node_id).into_v2()
+    conf_change(ConfChangeType::AddLearnerNode, node_id).into_v2()
 }
 
 pub fn conf_state(voters: Vec<u64>, learners: Vec<u64>) -> ConfState {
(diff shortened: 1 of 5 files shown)


src/raft.rs
```diff
@@ -1425,7 +1428,15 @@ impl<T: Storage> Raft<T> {
                         new_message(m.from, vote_resp_msg_type(m.get_msg_type()), None);
                     to_send.reject = true;
                     to_send.term = self.term;
+                    let (commit, commit_term) = self.raft_log.commit_info();
+                    to_send.commit = commit;
+                    to_send.commit_term = commit_term;
                     self.r.send(to_send, &mut self.msgs);
+                    // It should be safe to commit log by vote in all conditions as long as it's
+                    // not leader. Here use a stricter condition for best safety.
+                    if self.leader_id == INVALID_ID {
```
NingLin-P, line 1437: Looks like only `(pre-)candidate`'s `leader_id` will be `INVALID_ID`, so we can't forward a follower's commit index by vote?

src/raft.rs
```diff
@@ -2043,6 +2075,27 @@ impl<T: Storage> Raft<T> {
                 }
 
                 self.poll(m.from, m.get_msg_type(), !m.reject);
+                let last_commit = self.raft_log.committed;
+                if self.maybe_commit_by_vote(&m) && self.state != StateRole::Follower {
```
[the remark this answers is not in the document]
BusyJay: It can become follower after `poll`.

src/raft.rs
BusyJay: When a peer's term is increased, its `leader_id` will be reset. So at the beginning of every term, `leader_id` of all followers and candidates should be `INVALID_ID`.

NingLin-P: Follower won't increase its term for `MsgRequestPreVote`

BusyJay: It doesn't have to. It will increase its term when ticks over election timeout; otherwise there is already a leader, no need to advance commit index by messages from others.

NingLin-P: If the follower thinks itself is a learner, it won't increase its term and start campaign, and the issue still exists.

harness/tests/test_util/mod.rs
```diff
@@ -212,6 +212,10 @@ pub fn conf_state_v2(
     cs
 }
 
+pub fn conf_change(t: ConfChangeType, node_id: u64) -> ConfChange {
+    new_conf_change(t, node_id)
+}
```
NingLin-P, line 217: Is this function needed?

NingLin-P (commented):
Reat LGTM

NingLin-P (approved):

src/raft.rs
```diff
@@ -1425,7 +1428,15 @@ impl<T: Storage> Raft<T> {
...
                     to_send.reject = true;
                     to_send.term = self.term;
+                    let (commit, commit_term) = self.raft_log.commit_info();
+                    to_send.commit = commit;
+                    to_send.commit_term = commit_term;
                     self.r.send(to_send, &mut self.msgs);
+                    // It should be safe to commit log by vote in all conditions. Here
+                    // use a stricter condition for best safety.
+                    if self.leader_id == INVALID_ID || !self.promotable {
+                        self.maybe_commit_by_vote(&m);
```
gengliqi, line 1438: If we assume quorum won't change during election, the candidate may get a `RequestVote` and forward the commit index. I suggest the check of new conf change can be written under `maybe_commit_by_vote`.
Also, I don't think the `if self.leader_id == INVALID_ID || !self.promotable` can improve the safety. Instead, it adds some dependency that is not necessary, such that the `leader_id` must be set to `INVALID_ID` after some time. How about just using `self.maybe_commit_by_vote` or, more simply, calling this function whatever this peer rejects the vote request or not because it has no side effect in all conditions.

BusyJay: The condition can prevent follower in lease from updating its commit index by non-leader's message. What I try to do here is only updating commit index when there is no valid leader.

gengliqi: You don't need to worry about this case. The code here will ignore the RequestVote msg when it still in lease.
https://github.com/tikv/raft-rs/blob/a45c4a33229090a0984638bfa4bc163ef8b284ea/src/raft.rs#L1256-L1287

src/raft.rs
```diff
@@ -1966,6 +1973,47 @@ impl<T: Storage> Raft<T> {
...
+                    e
+                );
+            });
+        if self.num_pending_conf(&ents) != 0 {
+            // The candidate doesn't have to step down in theory, here just for best
+            // safety as we assume quorum won't change during election.
+            let term = self.term;
+            self.become_follower(term, INVALID_ID);
+        }
+    }
```
gengliqi, line 2015: How about adding a test for this logic?

gengliqi (commented):
Rest LGTM

src/raft.rs
BusyJay: Following assert is exactly for that purpose:

https://github.com/tikv/raft-rs/pull/411/files#diff-fbf583b45a1333793ec034c2c5f932b3d13543758165cf96878a110f56204b84R4732-R4737

gengliqi (approved):
LGTM

NingLin-P (approved):

PR #417, opened 2021-01-22
fix persisted index bug of snapshot

gengliqi:

This PR fixes a bug introduced by https://github.com/tikv/raft-rs/pull/410.
Consider the case below
1. A receives a snapshot with index 10
2. A gets a ready and handles it asynchronously
3. A receives a new snapshot with index 20
4. A calls on_persist_ready for ready 1
In step 4, the persisted index can not be updated to 10 because the first_index has changed to 21 so the term check can not be passed. (details in `RaftLog::term`)
I add a `maybe_persist_snap` function to fix this problem. The snapshot does not need to check the term because its data must be committed before and can not be changed in future.

### DIFF
--- a/src/log_unstable.rs
+++ b/src/log_unstable.rs
@@ -87,16 +87,23 @@ impl Unstable {
 
     /// Clears the unstable entries and moves the stable offset up to the
     /// last index, if there is any.
-    pub fn stable_entries(&mut self) {
+    pub fn stable_entries(&mut self, index: u64, term: u64) {
+        // The snapshot must be stabled before entries
+        assert!(self.snapshot.is_none());
         if let Some(entry) = self.entries.last() {
+            assert_eq!(entry.get_index(), index);
+            assert_eq!(entry.get_term(), term);
             self.offset = entry.get_index() + 1;
             self.entries.clear();
         }
     }
 
     /// Clears the unstable snapshot.
-    pub fn stable_snap(&mut self) {
-        self.snapshot = None;
+    pub fn stable_snap(&mut self, index: u64) {
+        if let Some(snap) = &self.snapshot {
+            assert_eq!(snap.get_metadata().index, index);
+            self.snapshot = None;
+        }
     }
 
     /// From a given snapshot, restores the snapshot to self, but doesn't unpack.
@@ -319,7 +326,7 @@ mod test {
     }
 
     #[test]
-    fn test_stable_entries() {
+    fn test_stable_snapshot_and_entries() {
         let ents = vec![new_entry(5, 1), new_entry(5, 2), new_entry(6, 3)];
         let mut u = Unstable {
             entries: ents.clone(),
@@ -328,7 +335,8 @@ mod test {
             logger: crate::default_logger(),
         };
         assert_eq!(ents, u.entries);
-        u.stable_entries();
+        u.stable_snap(4);
+        u.stable_entries(6, 3);
         assert!(u.entries.is_empty());
         assert_eq!(u.offset, 7);
     }
--- a/src/log_unstable.rs
+++ b/src/log_unstable.rs
@@ -91,18 +91,46 @@ impl Unstable {
         // The snapshot must be stabled before entries
         assert!(self.snapshot.is_none());
         if let Some(entry) = self.entries.last() {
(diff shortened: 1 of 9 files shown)


src/raft.rs
```diff
@@ -996,6 +996,11 @@ impl<T: Storage> Raft<T> {
         }
     }
 
+    /// Notifies that the snapshot have been persisted.
+    pub fn on_persist_snap(&mut self, index: u64) {
+        let _ = self.raft_log.maybe_persist_snap(index);
```
BusyJay, line 1001: `let _ =` is unnecessary.

src/raft_log.rs
```diff
@@ -477,6 +487,35 @@ impl<T: Storage> RaftLog<T> {
...
+            // commit index should not be less than snapshot's index
+            if index > self.committed {
+                fatal!(
+                    self.unstable.logger,
+                    "snapshot's index {} > committed {}",
+                    index,
+                    self.committed,
+                )
+            }
+            // All of the index of latter entries must be greater than snapshot's index
```
BusyJay, line 502:
```suggestion
            // All of the indexes of later entries must be greater than snapshot's index
```

src/raw_node.rs
```diff
@@ -563,15 +563,17 @@ impl<T: Storage> RawNode<T> {
...
         while let Some(record) = self.records.front() {
             if record.number > number {
                 break;
             }
             let mut record = self.records.pop_front().unwrap();
 
-            if let Some((i, t)) = record.snapshot {
-                index = i;
-                term = t;
+            if let Some((i, _)) = record.snapshot {
```
hicqu, line 573: Is it possible that some entries follow one snapshot in one ready?

gengliqi: Yes, it happens when the snapshot has up-to-date data and the leader is changed.
If a ready has a snapshot, the entries of the readies before this one must have smaller indexes so they don't need to call `on_persist_entries`.

src/raft_log.rs
```diff
@@ -477,6 +487,35 @@ impl<T: Storage> RaftLog<T> {
         }
     }
 
+    /// Attempts to persist the snapshot and returns whether it did.
+    pub fn maybe_persist_snap(&mut self, index: u64) -> bool {
```
BusyJay, line 491: How about also checking term and index matches from store for safety?

BusyJay (commented):
Rest LGTM

src/raft_log.rs
```diff
@@ -461,13 +461,32 @@ impl<T: Storage> RaftLog<T> {
...
+        // 2. A crashes and C becomes leader after getting the vote from D and E.
+        // 3. C proposes some raft logs and B receives these logs.
+        // 4. C crashes and A restarts and becomes leader again after getting the vote from D and E.
+        // 5. B receives the logs from A which are the same to the ones from step 1.
+        // 6. The logs from Ready has been persisted on B so it calls on_persist_ready and comes to here.
+        //
+        // We solve this problem by not forwarding the persisted index. It's pretty intuitive
+        // because the first_update_index means there are snapshot or some entries whose indexes
+        // are greater than or equal to the first_update_index have not been persisted yet.
+        let first_update_index = if let Some(index) = self
```
BusyJay, line 477: Following seems more readable.
```
let first_update_index = match &self.unstable.snapshot {
    Some(s) => s.get_metadata().index,
    None => self.unstable.offset,
};
```

src/log_unstable.rs
```diff
@@ -87,16 +87,23 @@ impl Unstable {
 
     /// Clears the unstable entries and moves the stable offset up to the
     /// last index, if there is any.
-    pub fn stable_entries(&mut self) {
+    pub fn stable_entries(&mut self, index: u64, term: u64) {
+        // The snapshot must be stabled before entries
+        assert!(self.snapshot.is_none());
         if let Some(entry) = self.entries.last() {
```
BusyJay, line 93: Now `index` and `term` are required, it can't be `None`.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

hicqu:
LGTM.

hicqu (approved):

PR #420, opened 2021-03-06
*: introduce judge_split_prevote

BusyJay:
During split prevote, a campaign will fail because all nodes think
it will collect enough votes, so after they actually start campaign,
no one votes for the other, the campaign has to fail.

`judge_split_prevote` solves the problem by adding extra constraint
to split prevote: only vote for nodes that have greater IDs. It's easy
to conclude that it works for peer numbers not greater 5. For 7 nodes,
it's still possible to split again. But it should be enough for most
cases. Because the constraint is only added for split prevote, so even
failure won't lead to worse result.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp;
+use std::cmp::{self, Ordering};
 use std::ops::{Deref, DerefMut};
 
 use crate::eraftpb::{
@@ -228,6 +228,7 @@ pub struct RaftCore<T: Storage> {
 
     skip_bcast_commit: bool,
     batch_append: bool,
+    judge_split_prevote: bool,
 
     heartbeat_timeout: usize,
     election_timeout: usize,
@@ -329,6 +330,7 @@ impl<T: Storage> Raft<T> {
                 promotable: false,
                 check_quorum: c.check_quorum,
                 pre_vote: c.pre_vote,
+                judge_split_prevote: c.judge_split_prevote,
                 read_only: ReadOnly::new(c.read_only_option),
                 heartbeat_timeout: c.heartbeat_tick,
                 election_timeout: c.election_tick,
@@ -1398,10 +1400,7 @@ impl<T: Storage> Raft<T> {
                     // ...or this is a PreVote for a future term...
                     (m.get_msg_type() == MessageType::MsgRequestPreVote && m.term > self.term);
                 // ...and we believe the candidate is up to date.
-                if can_vote
-                    && self.raft_log.is_up_to_date(m.index, m.log_term)
-                    && (m.index > self.raft_log.last_index() || self.priority <= m.priority)
-                {
+                if can_vote && self.may_vote(&m) {
                     // When responding to Msg{Pre,}Vote messages we include the term
                     // from the message, not the local term. To see why consider the
                     // case where a single node was previously partitioned away and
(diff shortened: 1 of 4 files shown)


BusyJay:
I have tested it with both configuration capacity 3 and 5, both can finish in one election timeout.

src/raft.rs
```diff
@@ -1444,6 +1459,33 @@ impl<T: Storage> Raft<T> {
...
+    fn may_vote(&self, m: &Message) -> bool {
+        match self.raft_log.is_up_to_date(m.index, m.log_term) {
+            Ordering::Greater => true,
+            Ordering::Equal => {
+                match self.priority.cmp(&m.priority) {
+                    Ordering::Greater => false,
+                    Ordering::Equal => {
+                        // judge split vote can break symmetry of campaign, but as
+                        // it only happens during split vote, the impact should not
+                        // be significant.
```
NingLin-P, line 1476: How can we tell if a campaign will end up split vote?

NingLin-P: This could cause some raft nodes with lower ID impossible to become the leader even we want to transfer leadership to it.

NingLin-P:
To avoid split prevote, maybe we can record the prevote like the read vote (but not need to persist it) and reject incoming vote request in the same term like what real campaign does.

src/raft.rs
BusyJay: If the configuration size is an odd number and leader is down, split vote can probably happen If two nodes are in PreCandidate state. `judge_split_prevote` only works on prevote, transfering leader skips prevote, so they won't have impact on the other.

BusyJay:
> ...record the prevote like the read vote...

Recording votes won't solve split vote. If pre-campaign works like actual campaign and split vote can happen in actual campaign, then it can also happen in pre-campaign.

NingLin-P:
> > ...record the prevote like the read vote...
> 
> Recording votes won't solve split vote. If pre-campaign works like actual campaign and split vote can happen in actual campaign, then it can also happen in pre-campaign.

We can't prevent split vote completely, but with random election timeout and vote recording, split vote should happen rarely.

BusyJay:
> ...split vote should happen rarely.

How can recording vote reduce the probability of split vote? It just make the split happen in early stage, from actual vote to pre vote. The strategy here can solve split completely in configuration size 3, and make it hardly happen in configuration size 5.

src/raft.rs
NingLin-P: > judge_split_prevote only works on prevote

Besides transfer leader, a node needs to pass pre-campaign before start the actual campaign, so `judge_split_prevote` will impact the whole election process (pre-vote should consider enabled as this is when `judge_split_prevote` work).

BusyJay: > so `judge_split_prevote` will impact the whole election process...

Indeed, it depends on whether nodes are working slowly. But in my tests, when one node is down, and after elections are finished, the leader count on each nodes don't have much differences (I remove balance leader scheduler before shutdown a node). And even it leads to more leaders on some node, it should not be a problem with the help of PD to reach a eventually balance.

NingLin-P:
> How can recording vote reduce the probability of split vote?

After recording a prevote at future term, a node should not start pre-compaing as it already prevote at that future term.

BusyJay:
A common situation of split vote in 3 voters situation is that leader is down, and two followers start campaign at the same time. How does the strategy you describe make the campaign succeed in one round of election?

NingLin-P:
> How does the strategy you describe make the campaign succeed in one round of election?

It can't, forget about it, I have some misunderstanding about the problem previously.

But `judge_split_prevote` may conflict with the random election timeout mechanism.  Random election timeout tends to let voters with shorter election timeout become leader, but `judge_split_prevote` prefer voters with large ID.

Together, they may cause election to take longer to finish: it is possible that voters with lower ID and shorter election timeout could win the election without split vote, but we had to wait for the voters with larger ID and longer election timeout to start campaign and become the leader.

NingLin-P:
> A common situation of split vote in 3 voters situation is that leader is down, and two followers start campaign at the same time. How does the strategy you describe make the campaign succeed in one round of election?

It can't, forget about it, I have some misunderstanding about the problem previously.

src/raft.rs
```diff
@@ -1469,6 +1484,37 @@ impl<T: Storage> Raft<T> {
...
+                        !self.judge_split_prevote
+                            || self.state != StateRole::PreCandidate
```
gengliqi, line 1503: Add some comments for transfer leader?

src/raft.rs
```diff
@@ -1447,6 +1458,10 @@ impl<T: Storage> Raft<T> {
                         self.election_elapsed = 0;
                         self.vote = m.from;
                     }
+                    // This means it's in split vote, give up election.
+                    if self.judge_split_prevote && self.state == StateRole::PreCandidate {
```
gengliqi, line 1462: It seems if this peer grants its (pre)vote to another one, it should reset `election_elapsed` or become follower.

BusyJay: This can lead to an awkward situation that A and B all grant prevotes to each other and all become follower.

gengliqi: Got it.

src/raft.rs
```diff
@@ -1312,6 +1314,18 @@ impl<T: Storage> Raft<T> {
                         "msg type" => ?m.get_msg_type(),
                     );
 
+                    // When judge_split_prevote, reject explicitly to make candidate exit PreCandiate early
+                    // so it will vote for other peer later.
+                    if self.judge_split_prevote
+                        && m.get_msg_type() == MessageType::MsgRequestPreVote
+                    {
```
gengliqi, line 1321: Seems `Follower` and `PreCandidate` are not different, both of them can (pre)vote to other peers.

BusyJay: `Follower` will ignore all `PreVoteResponse`. For example, A and B split pre-votes, and C can vote for both A and B. If B is chosen, and A doesn't step down to follower, it will still start campaign when C's prevote is received.

BusyJay:
PTAL

PR #423, opened 2021-03-18
Add free() for raft node

### LINKED ISSUE #419: Optimize memory footprint of raft state machine
After introducing [memory tracing](https://github.com/tikv/rfcs/pull/59) in TiKV, we found that the memory footprint could be very large if there were too many raft state machine. Most of them came from the inflight vector. Supposing `max_inflight_messages` was set to 4096, and there were 3 peers, leader would need to allocate 3 * 4096 * 8 = 96K memory. If there were 100k regions, then the virt memory became 9.1 GiB. And most of the memory were idle and perhaps never visited again.

There are many possible optimizations for this situation. On the one hand, if we can build a global flow control for all raft state machines, then the memory usage can be reduced; on the other hand, if we can reduce the total number of regions, the impact of flow control can also be omitted. But for the short term, I think it's easy to allocate lazily for inflight vector. We can use a linked list to allocate block by block and free them when it's idle. Because most raft state machines should be idle when it comes to 100k, so there will be only one block for all inflight vector. If each block is allocated as 512, then the memory footprint can be reduced to 572MiB, which is reasonably small enough.

jayzhan211:
Close #419 and https://github.com/tikv/raft-rs/pull/448 .

Add two API 
1. free_resources(id: u64)
2. free_resources_for_all()

Signed-off-by: jayzhan <jayzhan211@gmail.com>

### DIFF
--- a/harness/tests/integration_cases/test_raft_flow_control.rs
+++ b/harness/tests/integration_cases/test_raft_flow_control.rs
@@ -239,6 +239,7 @@ fn test_msg_app_flow_control_with_freeing_resources() {
 
     assert_eq!(r.prs().get(2).unwrap().ins.count(), 0);
     assert_eq!(r.prs().get(3).unwrap().ins.count(), 2);
+    assert_eq!(r.inflight_buffers_size(), 512);
 
     /*
     1: cap=0/start=0/count=0/buffer=[]
@@ -250,6 +251,7 @@ fn test_msg_app_flow_control_with_freeing_resources() {
 
     assert!(!r.prs().get(2).unwrap().ins.buffer_is_allocated());
     assert_eq!(r.prs().get(2).unwrap().ins.count(), 0);
+    assert_eq!(r.inflight_buffers_size(), 256);
 
     /*
     1: cap=0/start=0/count=0/buffer=[]
--- a/harness/tests/integration_cases/test_raft_flow_control.rs
+++ b/harness/tests/integration_cases/test_raft_flow_control.rs
@@ -175,3 +175,85 @@ fn test_msg_app_flow_control_recv_heartbeat() {
         r.read_messages();
     }
 }
+
+#[test]
+fn test_msg_app_flow_control_with_freeing_resources() {
+    let l = default_logger();
+    let mut r = new_test_raft(1, vec![1, 2, 3], 5, 1, new_storage(), &l);
+
+    r.become_candidate();
+    r.become_leader();
+
+    for (_, pr) in r.prs().iter() {
+        assert!(!pr.ins.buffer_is_allocated());
+    }
+
+    for i in 1..=3 {
+        // Force the progress to be in replicate state.
+        r.mut_prs().get_mut(i).unwrap().become_replicate();
+    }
+
+    r.step(new_message(1, 1, MessageType::MsgPropose, 1))
+        .unwrap();
+
+    for (&id, pr) in r.prs().iter() {
+        if id != 1 {
+            assert!(pr.ins.buffer_is_allocated());
+            assert_eq!(pr.ins.count(), 1);
+        }
+    }
+
+    /*
+    1: cap=0/start=0/count=0/buffer=[]
+    2: cap=256/start=0/count=1/buffer=[2]
+    3: cap=256/start=0/count=1/buffer=[2]
+    */
+
+    let mut resp = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+    resp.index = r.raft_log.last_index();
+    r.step(resp).unwrap();
+
+    assert_eq!(r.prs().get(2).unwrap().ins.count(), 0);
+
+    /*
+    1: cap=0/start=0/count=0/buffer=[]
+    2: cap=256/start=1/count=0/buffer=[2]
+    3: cap=256/start=0/count=1/buffer=[2]
+    */
+
+    r.step(new_message(1, 1, MessageType::MsgPropose, 1))
+        .unwrap();
+
+    assert_eq!(r.prs().get(2).unwrap().ins.count(), 1);
+    assert_eq!(r.prs().get(3).unwrap().ins.count(), 2);
+
+    /*
+    1: cap=0/start=0/count=0/buffer=[]
+    2: cap=256/start=1/count=1/buffer=[2,3]
+    3: cap=256/start=0/count=2/buffer=[2,3]
+    */
+
+    let mut resp = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+    resp.index = r.raft_log.last_index();
+    r.step(resp).unwrap();
+
+    assert_eq!(r.prs().get(2).unwrap().ins.count(), 0);
+    assert_eq!(r.prs().get(3).unwrap().ins.count(), 2);
+
+    /*
+    1: cap=0/start=0/count=0/buffer=[]
+    2: cap=256/start=2/count=0/buffer=[2,3]
+    3: cap=256/start=0/count=2/buffer=[2,3]
+    */
+
+    r.maybe_free_inflight_buffers();
+
+    assert!(!r.prs().get(2).unwrap().ins.buffer_is_allocated());
+    assert_eq!(r.prs().get(2).unwrap().ins.count(), 0);
+
+    /*
+    1: cap=0/start=0/count=0/buffer=[]
+    2: cap=0/start=0/count=0/buffer=[]
+    3: cap=256/start=0/count=2/buffer=[2,3]
+    */
+}
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -2835,4 +2835,13 @@ impl<T: Storage> Raft<T> {
     pub fn uncommitted_size(&self) -> usize {
         self.uncommitted_state.uncommitted_size
     }
+
+    /// A Raft leader allocates a vector with capacity `max_inflight_msgs` for every peer.
+    /// It takes a lot of memory if there are too many Raft groups. `maybe_free_inflight_buffers`
+    /// is used to free memory if necessary.
+    pub fn maybe_free_inflight_buffers(&mut self) {
+        for (_, pr) in self.mut_prs().iter_mut() {
+            pr.ins.maybe_free_buffer();
+        }
+    }
 }
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -845,6 +845,15 @@ impl<T: Storage> RaftCore<T> {
 }
 
 impl<T: Storage> Raft<T> {
+    /// Get the inflight buffer size.
+    pub fn inflight_buffers_size(&self) -> usize {
+        let mut total_size = 0;
+        for (_, pr) in self.prs().iter() {
+            total_size += pr.ins.cap();
+        }
+        total_size
+    }
+
     /// Sends an append RPC with new entries (if any) and the current commit index to the given
     /// peer.
     pub fn send_append(&mut self, to: u64) {
--- a/src/tracker/inflights.rs
+++ b/src/tracker/inflights.rs
@@ -15,7 +15,7 @@
 // limitations under the License.
 
 /// A buffer of inflight messages.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct Inflights {
     // the starting index in the buffer
     start: usize,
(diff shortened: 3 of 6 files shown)


accelsao:
how to add test? or not

src/tracker/inflights.rs
```diff
@@ -122,6 +122,12 @@ impl Inflights {
         self.count = 0;
         self.start = 0;
     }
+
+    /// free buffer memory
+    #[inline]
+    pub fn shrink(&mut self) {
+        self.buffer.shrink_to_fit();
```
accelsao, line 129: After shrink the capacity will change,  the size for determine full() will be effected,  is it fine?

src/raft.rs
```diff
@@ -2678,4 +2678,27 @@ impl<T: Storage> Raft<T> {
     pub fn uncommitted_size(&self) -> usize {
         self.uncommitted_state.uncommitted_size
     }
+
+    /// Free resources for all the raft node
+    pub fn free_resources_for_all(&mut self) {
```
BusyJay, line 2683: Just this method is enough. Choosing target id is too verbose.

src/tracker/inflights.rs
BusyJay: It's wrong. It should be able to reallocate when new messages need to be sent.

BusyJay:
Tests should be added to check if flow control still work as expected before and after calling shrink.

src/tracker/inflights.rs
```diff
@@ -122,6 +127,23 @@ impl Inflights {
         self.count = 0;
         self.start = 0;
     }
+
+    /// free buffer memory by reassignment
+    #[inline]
+    pub fn free_mem(&mut self) {
+        let mut v = Vec::with_capacity(self.count);
+        for i in 0..self.count {
+            v.push(self.buffer[self.start + i])
```
BusyJay, line 136: I don't get it. `v` doesn't seem to be used, and why append the element from buffer to `v`? On the other hand, `self.buffer` should be dropped to free memory.

accelsao: I am trying to re-create an `buffer` with `length` & `capacity` equal to `count`, in order to decrease the allocated memory, 
and remain only from `self.start` to `self.start + self.count - 1`

> On the other hand, `self.buffer` should be dropped to free memory

If  `self.count` is zero, `self.buffer` will be `Vec` with capacity 0, I am not sure what is the implementation inside, but I guess there is no memory allocated?
If  `self.count` is positive, we should at least remain `self.count` sized Vec (?

accelsao: `shrink` that mentioned before, I think it will do nothing if the `Vec` is full once

jayzhan211: What is the issue left related to the memory problem? Is #10156?

BusyJay: What is 10156?

jayzhan211: Sorry, is this https://github.com/tikv/tikv/issues/10156, found it under https://github.com/tikv/rfcs/pull/59

BusyJay: /cc @hicqu

hicqu: The left issue is unchanged for raft-rs. Memory usage of `Inflight` is too large if there are lots of raft peers.

hicqu:
How about lazily allocate memory instead of adding `free_*` functions? @jayzhan211

hicqu:
How about this solution:
for an given `Inflight`,
* allocate memory only if necessary
* reallocate memory if more items are pushed
* free and malloc again after `free_to` is called, and shrink the buffer if necessary

Or this:
* Allocate link-list nodes on heap, every node carries 16 u64 numbers

Any suggestions @BusyJay

src/tracker/inflights.rs
hicqu: @BusyJay do you mean `free_mem` should only free empty inflights, but keep no-empty inflights unchanged?

hicqu: I have discussed with Jay. He thinks the current implemenatation is OK. BTW how about using `vec.extend_from_slice` instead?

src/tracker/inflights.rs
```diff
@@ -122,6 +127,23 @@ impl Inflights {
...
+        }
+        self.buffer = v;
+        self.start = 0;
+    }
+
+    /// reallocate buffer memory
+    #[inline]
+    pub fn reallocate(&mut self) {
```
BusyJay, line 144: Progress should reallocate when necessary.

jayzhan211: Currently, buffer will be reallocated when you call this function, what is other situation when we need reallocation?

hicqu:
@jayzhan211 we can try this solution:
* `free_mem` **shouldn't** re-allocate memory. So for an un-empty `Inflights`, we can
  1. leave it unchanged, or
  2. Call `shrink_to_fit` for it
  
  Both of them are OK.
* `Inflights::reallocate` is not necessary. Memory can be allocated when `Inflights::add` is called.

@BusyJay please help to confirm whether it fits with your design or not.

hicqu:
@jayzhan211 any updates?

jayzhan211:
> @jayzhan211 any updates?

Does it fit @BusyJay 's design

BusyJay:
> @jayzhan211 we can try this solution:
> 
> * `free_mem` **shouldn't** re-allocate memory. So for an un-empty `Inflights`, we can
>   
>   1. leave it unchanged, or
>   2. Call `shrink_to_fit` for it
>   
>   Both of them are OK.
> * `Inflights::reallocate` is not necessary. Memory can be allocated when `Inflights::add` is called.
> * Call `free_mem` when a peer become follower.
> 
> @BusyJay please help to confirm whether it fits with your design or not.

LGTM, thanks!

jayzhan211:
I have difficult for adding test (not famailiar with APIs).
Maybe add it afterall.

src/tracker/inflights.rs
```diff
@@ -122,6 +127,19 @@ impl Inflights {
...
+
+    /// Free unused memory
+    #[inline]
+    pub fn free_mem(&mut self) {
+        // remove unused buffer, from start + count to the end of buffer
+        // ignore first index to self.start - 1
+        let len = self.buffer.len();
+        for _ in self.start + self.count..len {
+            self.buffer.pop();
+        }
```
jayzhan211, line 139: We initialize the buffer with fixed capacity, so we need to remove element for reducing allocated memory, any better ways?

harness/tests/integration_cases/test_raft_flow_control.rs
```diff
@@ -175,3 +175,24 @@ fn test_msg_app_flow_control_recv_heartbeat() {
...
+//     // force the progress to be in replicate state
+//     r.mut_prs().get_mut(2).unwrap().become_replicate();
+//
+//     // send message from 2 to 1 for adding inflights
+//     for _ in 0..10 {
+//         r.step(new_message(1, 1, MessageType::MsgPropose, 1))
+//             .expect("");
+//         r.read_messages();
+//     }
+//
```
jayzhan211, line 195: After this line, inflights 2 (follower) is filled with 10 element, inflights 1 (leader) is empty, so it seems that we need to transfer leader from 2 to 1, then transfer 1 to 2 again, so peer 2 will call `free_mem` ?

src/tracker/inflights.rs
hicqu: How about don't touch it if it's not empty? So the resource can be freed later, and no allocations are introduced.

hicqu:
@jayzhan211 I have send a patch PR to your code. Could you take a look?

src/tracker/inflights.rs
jayzhan211: > How about don't touch it if it's not empty? 

I am not sure how frequently the buffer is empty. If it happens a lot, this approach is good enough.

hicqu (approved):

harness/tests/integration_cases/test_raft_flow_control.rs
```diff
@@ -175,3 +175,44 @@ fn test_msg_app_flow_control_recv_heartbeat() {
...
+            assert!(pr.ins.buffer_is_allocated());
+        }
+    }
+
+    let mut resp = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+    resp.index = r.raft_log.last_index();
+    r.step(resp).unwrap();
+
+    // Test `maybe_free_inflight_buffers` works as expected.
+    r.maybe_free_inflight_buffers();
```
BusyJay, line 208: It seems the case can pass even the function does nothing.

src/tracker/inflights.rs
```diff
@@ -35,6 +38,7 @@ impl Clone for Inflights {
             start: self.start,
             count: self.count,
             buffer,
+            cap: self.cap,
```
BusyJay, line 41: L36 can be also removed.

jayzhan211: Then, I think we can use derive(Clone) for inflight

BusyJay: Now that capacity is maintained explicitly, so the statement is not valid anymore. `derive(Clone)` is a good idea!.

harness/tests/integration_cases/test_raft_flow_control.rs
```diff
@@ -175,3 +175,103 @@ fn test_msg_app_flow_control_recv_heartbeat() {
...
+
+    for (&id, pr) in r.prs().iter() {
+        if id != 1 {
+            assert!(pr.ins.buffer_is_allocated());
+        }
+    }
+
+    /*
+    1: cap=0/start=0/count=0/buffer=[]
+    2: cap=256/start=0/count=1/buffer=[2]
```
hicqu, line 207: count should be 1?

harness/tests/integration_cases/test_raft_flow_control.rs
```diff
@@ -175,3 +175,103 @@ fn test_msg_app_flow_control_recv_heartbeat() {
...
+    3: cap=256/start=0/count=1/buffer=[2]
+    */
+
+    let mut resp = new_message(2, 1, MessageType::MsgAppendResponse, 0);
+    resp.index = r.raft_log.last_index();
+    r.step(resp).unwrap();
+
+    for (&id, pr) in r.prs().iter() {
```
hicqu, line 215: The loop is not necessary if only tests `id == 2`.

jayzhan211:
I will use buffer.capacity, not cap(), because I found that we initialize buffer with `cap`

harness/tests/integration_cases/test_raft_flow_control.rs
jayzhan211: I am not sure why `MsgPropose` sends `2` to others, but it did.

jayzhan211:
@hicqu PTAL

hicqu (dismissed):

hicqu:
@jayzhan211 seems some commits break the DCO check. Could you rebase and squash the PR like this:
```sh
git checkout free -b backup-free
git branch -D free
git checkout master -b free
git merge backup-free --no-commits --no-logs --squash
git commit -s     # then cleanup squashed commit logs
```

jayzhan211:
> @jayzhan211 seems some commits break the DCO check. Could you rebase and squash the PR like this:
> 
> ```shell
> git checkout free -b backup-free
> git branch -D free
> git checkout master -b free
> git merge backup-free --no-commits --no-logs --squash
> git commit -s     # then cleanup squashed commit logs
> ```

Done

src/tracker/inflights.rs
```diff
@@ -121,6 +112,29 @@ impl Inflights {
     pub fn reset(&mut self) {
         self.count = 0;
         self.start = 0;
+        self.buffer = vec![];
+    }
+
+    /// Number of inflight messages.
+    #[inline]
+    pub fn count(&self) -> usize {
```
BusyJay, line 120: How about making it doc hiddern as it's only used in tests?

BusyJay (commented):
@hicqu you may need a method to monitor the memory correctly.

hicqu:
> @hicqu you may need a method to monitor the memory correctly.

Do you mean we need `fn Raft::inflight_buffers_size`?

jayzhan211:
> > you may need a method to monitor the memory correctly.
> 
> @BusyJay Do you mean we need `fn Raft::inflight_buffers_size`?

Add a function that return `buffer.capacity()`?

BusyJay:
> @BusyJay Do you mean we need `fn Raft::inflight_buffers_size`?

Yes, something like that.

jayzhan211:
I think the buffer size will be either `0` or `cap`

hicqu:
@jayzhan211 we need to add a new method for `Raft`:
```rust
+
+    /// Get the inflight buffer size.
+    pub fn inflight_buffers_size(&self) -> usize {
+        let mut total_size = 0;
+        for (_, pr) in self.prs().iter() {
+            total_size += pr.ins.buffer.capacity();
+        }
+        total_size
+    }
```

to get the allocated buffer size for a raft peer.

src/tracker/inflights.rs
jayzhan211: I got an error: missing documentation for an associated function

hicqu (approved):

BusyJay (approved):
Thanks!

jayzhan211:
Thanks! @hicqu @BusyJay

PR #433, opened 2021-05-07
Separate message and persisted message in Ready

gengliqi:
When using async ready, the io process may belong to another thread. The messages of the follower can only be achieved after it knows the ready is persisted, namely, after receiving the io completion msg from the io thread, which unnecessarily increases the msg latency.
The problem above can be fixed by packaging messages and io tasks and sending them to the io thread.

This PR separates message and persisted message in Ready.

### DIFF
--- a/src/raw_node.rs
+++ b/src/raw_node.rs
@@ -24,13 +24,15 @@ use std::{collections::VecDeque, mem};
 
 use protobuf::Message as PbMessage;
 use raft_proto::ConfChangeI;
+use slog::Logger;
 
 use crate::eraftpb::{ConfState, Entry, EntryType, HardState, Message, MessageType, Snapshot};
 use crate::errors::{Error, Result};
 use crate::read_only::ReadState;
 use crate::{config::Config, StateRole};
 use crate::{Raft, SoftState, Status, Storage};
-use slog::Logger;
+
+use slog::info;
 
 /// Represents a Peer node in the cluster.
 #[derive(Debug, Default)]
@@ -640,7 +642,7 @@ impl<T: Storage> RawNode<T> {
     /// Returns the LightReady that contains commit index, committed entries and messages.
     ///
     /// Since Ready must be persisted in order, calling this function implicitly means
-    /// all readys collected before have been persisted.
+    /// all ready collected before have been persisted.
     #[inline]
     pub fn advance_append(&mut self, rd: Ready) -> LightReady {
         self.commit_ready(rd);
--- a/src/raw_node.rs
+++ b/src/raw_node.rs
@@ -542,11 +542,11 @@ impl<T: Storage> RawNode<T> {
         let rd_record = self.records.back().unwrap();
         assert!(rd_record.number == rd.number);
         let raft = &mut self.raft;
-        if rd_record.snapshot.is_some() {
-            raft.raft_log.stable_snap();
+        if let Some((index, _)) = rd_record.snapshot {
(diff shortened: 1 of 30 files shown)


src/raw_node.rs
```diff
@@ -209,7 +224,8 @@ struct ReadyRecord {
 pub struct LightReady {
     commit_index: Option<u64>,
     committed_entries: Vec<Entry>,
-    messages: Vec<Vec<Message>>,
+    messages: Vec<Message>,
+    is_persisted_msg: bool,
```
BusyJay, line 228: Probably should be stored and checked in `Ready`. `take_persisted_messages` is meaningless for `LightReady`.

src/raw_node.rs
```diff
@@ -397,14 +439,11 @@ impl<T: Storage> RawNode<T> {
...
-            mem::swap(&mut rd.messages, &mut self.messages);
-        }
-
-        if raft.state == StateRole::Leader && !raft.msgs.is_empty() {
+        if !raft.msgs.is_empty() {
+            rd.messages = mem::take(&mut raft.msgs);
             // Leader can send messages immediately to make replication concurrently.
             // For more details, check raft thesis 10.2.1.
-            rd.messages.push(mem::take(&mut raft.msgs));
+            rd.is_persisted_msg = raft.state != StateRole::Leader;
```
BusyJay, line 446: This implicitly requires followers to not step any messages before last `Ready` is advanced. This requirement should be documented in `ready` function.

BusyJay (commented):
Rest LGTM

src/raw_node.rs
gengliqi: Done

BusyJay (approved):

src/raw_node.rs
```diff
@@ -174,18 +176,50 @@ impl Ready {
...
+        } else {
+            &[]
+        }
     }
 
     /// Take the Messages.
     #[inline]
-    pub fn take_messages(&mut self) -> Vec<Vec<Message>> {
-        self.light.take_messages()
+    pub fn take_messages(&mut self) -> Vec<Message> {
```
hicqu, line 189: How about `fn take_messages(&mut self, persisted: bool)`? So that `take_persisted_messages` can be removed.

src/raw_node.rs
```diff
@@ -174,18 +176,50 @@ impl Ready {
...
+        } else {
+            Vec::new()
+        }
     }
 
-    /// MustSync indicates whether the HardState and Entries must be synchronously
-    /// written to disk or if an asynchronous write is permissible.
+    /// MustSync is true if and only if
+    /// 1. no HardState or only its commit is different from before
+    /// 2. no Entries and Snapshot
```
hicqu, line 220: Seems it's different from the implementation?

hicqu:
Rest LGTM.

src/raw_node.rs
gengliqi: My mistake. The `true` should be `false`.

gengliqi: My mistake. The `true` should be `false`.

src/raw_node.rs
gengliqi: I am ok with these two ways. What's your opinion? @BusyJay

BusyJay: Separating them seems more readable. It's strange to call `take_xxx` twice.

hicqu (approved):

BusyJay (approved):

PR #434, opened 2021-05-11
fix docs about joint consensus

BusyJay:


NingLin-P (dismissed):

src/lib.rs
```diff
@@ -424,42 +423,37 @@ It permits resilient arbitrary dynamic membership changes. A membership change c
...
-use slog::{Drain, o};
-
-let mut config = Config { id: 1, ..Default::default() };
-let store = MemStorage::new_with_conf_state((vec![1, 2], vec![]));
-let logger = slog::Logger::root(slog_stdlog::StdLog.fuse(), o!());
-let mut node = RawNode::new(&mut config, store, &logger).unwrap();
-node.raft.become_candidate();
-node.raft.become_leader();
-
+For example to promote a learner 4 and demote an existing voter 3:
```
gengliqi, line 431: Seems 3 and 4 is not in config?

BusyJay: It's just a code showing how to construct joint changes, it doesn't matter whether nodes are in configurations as long as it can compile.

BusyJay: Codes after `#` will not be decorated.

gengliqi (approved):

PR #435, opened 2021-05-16
*: bump deps, replace `quick-error` with `thiserror`

PsiACE:
In particular, getset 0.1.1 and rand 0.8.

Signed-off-by: Chojan Shang <psiace@outlook.com>


Cargo.toml
```diff
@@ -26,27 +26,27 @@ default-logger = ["slog-stdlog", "slog-envlogger", "slog-term"]
 # Make sure to synchronize updates with Harness.
 [dependencies]
 fxhash = "0.2.1"
-fail = { version = "0.3", optional = true }
-getset = "0.0.9"
+fail = { version = "0.4", optional = true }
```
BusyJay, line 29: Perhaps replace it with `thiserror` is better.

datadriven/Cargo.toml
```diff
@@ -7,10 +7,10 @@ edition = "2018"
...
 [dependencies]
-anyhow = "1.0.31"
-regex = "1.3.9"
+anyhow = "1.0.40"
+regex = "1"
 lazy_static = "1.4.0"
-slog = "2.5.2"
-slog-term = "2.6.0"
-slog-async = "2.5.0"
+slog = "2.7.0"
```
BusyJay, line 13: Better keep it as it is. Cargo will try to use latest version when possible.

Cargo.toml
PsiACE: quick-error -> thiserror, done!

But, do we need to remove the dependency of fail and the feature of failpoints? @BusyJay

datadriven/Cargo.toml
PsiACE: Just to unify the dependency version, please ignore this

BusyJay: Then we can keep using 2.2 as the library does. A smaller version leaves more flexibility for downstream to choose a suitable version.

BusyJay (approved):
Thanks!

hicqu (approved):

PR #436, opened 2021-05-16
feat: use parking_lot

PsiACE:
use `parking_lot::{Mutex, Once, RwLock}` can bring better performance 

Signed-off-by: Chojan Shang <psiace@outlook.com>


BusyJay:
These places are not in performance critical path, I prefer not to introduce another dependency.

PsiACE:
> These places are not in performance critical path, I prefer not to introduce another dependency.

OK, I will close this PR

PR #438, opened 2021-05-17
use bytes for protobuf

hicqu:

Waiting for https://github.com/tikv/protobuf-build/pull/50.

If `protobuf-codec` is enabled, `bytes` will be used to replace `Vec<u8>` for protobuf `bytes` type.

However if `prost-codec` is enabled,  `Vec<u8>` will be still used.


Cargo.toml
```diff
@@ -25,10 +25,11 @@ default-logger = ["slog-stdlog", "slog-envlogger", "slog-term"]
 
 # Make sure to synchronize updates with Harness.
 [dependencies]
+bytes = "1.0"
 fxhash = "0.2.1"
 fail = { version = "0.3", optional = true }
 getset = "0.0.9"
-protobuf = "2"
+protobuf = "2.23"
```
BusyJay, line 32: We are using a fork of protobuf, which may not be at the same version.

hicqu: protobuf 2.8 uses a stale version for `bytes`. I think we can upgrade our protobuf.

.travis.yml
```diff
@@ -26,8 +26,19 @@ cache:
...
 
+before_install:
+  - if [[ $TRAVIS_OS_NAME == "linux" ]]; then export PATH=$HOME/protobuf/bin:$PATH; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" ]]; then export C_INCLUDE_PATH=$C_INCLUDE_PATH:$HOME/protobuf/include; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" ]]; then export CPLUS_INCLUDE_PATH=$CPLUS_INCLUDE_PATH:$HOME/protobuf/include; fi
 
 install:
+  - >
+    if [[ $TRAVIS_OS_NAME == "linux" ]]; then
+        mkdir -p $HOME/protobuf && pushd $HOME/protobuf &&
```
BusyJay, line 37: How can test pass in the past?

src/raft.rs
```diff
@@ -1238,7 +1238,7 @@ impl<T: Storage> Raft<T> {
             m.commit = commit;
             m.commit_term = commit_term;
             if campaign_type == CAMPAIGN_TRANSFER {
-                m.context = campaign_type.to_vec();
+                m.context = campaign_type.to_vec().into();
```
BusyJay, line 1241: Perhaps we can add a new method without allocation. `Bytes::from_static_bytes` will just store the static pointer.

.travis.yml
hicqu: This PR adds `rustproto.proto` in `proto/include`. The compiler needs to compile it with `protoc`.

BusyJay (approved):

gengliqi (approved):

BusyJay (approved):

NingLin-P (approved):

PR #440, opened 2021-05-31
implement committed entries pagination

hicqu:

This is a re-implement for #356, except such things:
* the new configuration `max_committed_size_per_ready` isn't pushed into `RaftLog`
* new test cases are simplified into one, and it covers `Ready` and `LightReady`

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -259,7 +259,7 @@ pub struct RaftCore<T: Storage> {
     uncommitted_state: UncommittedState,
 
     /// Max size per committed entries in a `Read`.
-    pub max_committed_size_per_ready: u64,
+    pub(crate) max_committed_size_per_ready: u64,
 }
 
 /// A struct that represents the raft consensus itself. Stores details concerning the current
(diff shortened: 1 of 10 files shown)


BusyJay:
Why not set it by `Config`?

hicqu:
> Why not set it by `Config`?

Using an `Options` makes it configurable. For example, we can set it to `8M` generally, however if there is no available memory, we can set it to `0` to fetch nothing temporarely.

BusyJay:
How about adding an API to mutate the configuration instead of adding a new parameter?

BusyJay:
I recall there is already a PR implementing the feature #356, can you update the PR and pick to master?

src/raft.rs
```diff
@@ -257,6 +257,9 @@ pub struct RaftCore<T: Storage> {
 
     /// Track uncommitted log entry on this node
     uncommitted_state: UncommittedState,
+
+    /// Max size per committed entries in a `Read`.
+    pub max_committed_size_per_ready: u64,
```
BusyJay, line 262: Should be private.

BusyJay (commented):
test_commit_pagination_after_restart should also be ported.

BusyJay (approved):

nolouch (approved):

5kbpers (approved):

PR #441, opened 2021-06-03
add unstable entries size

gengliqi:
Add unstable entries size in order that the application can decide when to get ready if the entries size is large.


gengliqi:
@BusyJay @hicqu @NingLin-P  A small PR, PTAL.

BusyJay:
Can you provide more detail explanation? The standard way is use `has_ready`.

gengliqi:
The purpose is to avoid too large io size. Too large io size is inefficient for network and disk. If the application knows the pending entries size is larger than a typical value, it can get ready and then sending msg and io request to disk.

src/log_unstable.rs
```diff
@@ -31,6 +32,9 @@ pub struct Unstable {
     /// All entries that have not yet been written to storage.
     pub entries: Vec<Entry>,
 
+    /// The size of entries
+    pub entries_size: usize,
```
BusyJay, line 36: Should be private.

src/log_unstable.rs
```diff
@@ -150,20 +157,26 @@ impl Unstable {
...
+            for e in &self.entries[(after - off) as usize..] {
+                self.entries_size -= e.compute_size() as usize;
+            }
             self.entries.truncate((after - off) as usize);
-            self.entries.extend_from_slice(ents);
         }
+        self.entries.extend_from_slice(ents);
+        self.entries_size += ents
+            .iter()
+            .map(|ent| ent.compute_size() as usize)
```
BusyJay, line 178: How about counting the data size and context size directly?

gengliqi: Done

src/util.rs
```diff
@@ -156,3 +156,17 @@ impl<'a> Union<'a> {
...
+pub fn entry_approximate_size(e: &Entry) -> usize {
+    //  message Entry {
+    //      EntryType entry_type = 1;
+    //      uint64 term = 2;
+    //      uint64 index = 3;
+    //      bytes data = 4;
+    //      bytes context = 6;
+    //      bool sync_log = 5;(Deprecated)
+    // }
+    e.data.len() + e.context.len() + 20
```
BusyJay, line 171: Probably 12 is enough. Each field has 2 bytes for tag and values and extra 1 byte for data length, 1 byte for large index.

src/log_unstable.rs
gengliqi: It's the same as other variables.

BusyJay: We probably should also make other fields private. Prevent another pub field to not make access control worse.

gengliqi: I agree. I think most of the public fields can be private. We should clean them up afterward.

zhangjinpeng1987 (commented):
Is it possible push this change to ectd? Does etcd also need such change?

gengliqi:
> Is it possible push this change to ectd? Does etcd also need such change?

Maybe etcd/raft needs this change too. Since it's a very small change, I think we can leave it alone first.

BusyJay (approved):

src/util.rs
```diff
@@ -155,3 +155,17 @@ impl<'a> Union<'a> {
...
+pub fn entry_approximate_size(e: &Entry) -> usize {
+    //  message Entry {
+    //      EntryType entry_type = 1;
+    //      uint64 term = 2;
+    //      uint64 index = 3;
+    //      bytes data = 4;
+    //      bytes context = 6;
+    //      bool sync_log = 5;(Deprecated)
+    // }
+    e.data.len() + e.context.len() + 12
```
NingLin-P, line 170: Plz add more comments about the magic number `12` here.

NingLin-P (commented):
Rest LGTM

NingLin-P:
> The purpose is to avoid too large io size. Too large io size is inefficient for network and disk due to bad pipeline. If the application knows the pending entries size is larger than a typical value, it can get ready and then sending msg and io request to disk.

It seems the purpose is more like to prevent too small io size, as the application can prevent getting and handling a `Ready` if the `entries_size` is smaller than a typical value, but the application can do nothing if `entries_size` is grown and larger than a typical value between tow check. Or do you mean to check the `entries_size` every time after handle a `MsgPropose`/`MsgAppend` message?

src/util.rs
gengliqi: Updated

gengliqi:
> > The purpose is to avoid too large io size. Too large io size is inefficient for network and disk due to bad pipeline. If the application knows the pending entries size is larger than a typical value, it can get ready and then sending msg and io request to disk.
> 
> It seems the purpose is more like to prevent too small io size, as the application can prevent getting and handling a `Ready` if the `entries_size` is smaller than a typical value, but the application can do nothing if `entries_size` is grown and larger than a typical value between tow check. Or do you mean to check the `entries_size` every time after handle a `MsgPropose`/`MsgAppend` message?

Yes, I mean checking the `entries_size` every time to prevent too large io size.

gengliqi:
@BusyJay @NingLin-P PTAL again, tks.

src/util.rs
NingLin-P: In a long run system, `index` and `term` may larger than 16383 (2 bytes), but it is okay to me since it is an approximate size.

NingLin-P (approved):

BusyJay (approved):

PR #442, opened 2021-06-07
update docs

BusyJay:
- mention async ready
- fix doc warnings

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -301,6 +301,14 @@ need to update the applied index and resume `apply` later:
     }
     ```
 
+    Note, although Raft guarentees only persisted entries will be applied, but it
+    doesn't guarentee commit index is persisted before being applied. For example,
+    if application is restarted after applying committed entries before persisting
+    commit index, apply index can be larger than commit index and cause panic. To
+    solve the problem, persisting commit index with or before applying entries.
+    You can also always assign commit index to the `max(commit_index, applied_index)`
+    after restarting, *it may work but potential log loss may also be ignored silently*.
+
 4. Check whether `entries` is empty or not. If not empty, it means that there are newly added
 entries but have not been committed yet, we must append the entries to the Raft log:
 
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -412,6 +418,15 @@ to advance the applied index inside.
 
 For more information, check out an [example](examples/single_mem_node/main.rs#L113-L179).
 
+Sometimes it's better not to block the raft machine in IO operation, so that latency of
+read/write can be more predictable and the fsync frequencey can be controlled. The crate
+supports async ready to offload the IO operation to other thread. The usage is the same as
+above except:
+1. all writes are not required to be persisted, they can be written into memory caches;
```
NingLin-P, line 425:
```suggestion
1. Before handling the next write, all previous writes are not required to be persisted, they can be written into memory caches;
```

src/lib.rs
```diff
@@ -412,6 +418,15 @@ to advance the applied index inside.
...
+2. persisted messages should be sent after all coresponding writes are persisted;
+3. [`advance_append_async`](RawNode::advance_append_async) is used when all writes are finished
+    instead of `advance/advance_append`.
```
NingLin-P, line 428: Should also mention what function to call when the writes are persisted.

NingLin-P (commented):
Rest LGTM

src/lib.rs
BusyJay: What's the difference?

src/lib.rs
BusyJay: It's mentioned in `advance_append_async`.

src/lib.rs
NingLin-P: `all writes are not required to be persisted` looks like no need to persist the write at all.

BusyJay: How about "all writes are not required to be persisted immediately"?

NingLin-P: lgtm

NingLin-P (approved):

gengliqi (approved):
LGTM

PR #448, opened 2021-07-01
introduce elastic inflight to save memory

### LINKED ISSUE #419: Optimize memory footprint of raft state machine

hicqu:

Like #423 this PR also want to resolve #419 , but use another solution that lazy allocating inflight buffers and shrink it if necessary. No new APIs are introduced.

Would you mind to take a look? @jayzhan211

### DIFF
--- a/src/tracker/inflights.rs
+++ b/src/tracker/inflights.rs
@@ -141,137 +141,86 @@ impl Inflights {
         self.offset = 0;
         self.buffer = VecDeque::with_capacity(0);
     }
+
+    #[cfg(test)]
+    fn inflights(&self) -> Vec<u64> {
+        let mut ret = Vec::with_capacity(self.count);
+        let (mut chunk, mut offset) = (0, self.offset);
+        for _ in 0..self.count {
+            ret.push(self.buffer[chunk][offset]);
+            offset += 1;
+            if offset == Self::BUF_CHUNK_LEN {
+                offset = 0;
+                chunk += 1;
+            }
+        }
+        ret
+    }
 }
 
-/************************
 #[cfg(test)]
 mod tests {
     use super::Inflights;
 
     #[test]
     fn test_inflight_add() {
-        let mut inflight = Inflights::new(10);
-        for i in 0..5 {
-            inflight.add(i);
+        let chunk_len = Inflights::BUF_CHUNK_LEN;
+
+        for &offset in &[0, 8, chunk_len - 1] {
+            let mut inflights = Inflights::new(100);
+            inflights.offset = offset;
+            assert_eq!(inflights.capacity, 100);
+
+            (0..100).for_each(|i| inflights.add(i as u64));
+            assert_eq!(inflights.count, 100);
+            assert_eq!(inflights.offset, offset);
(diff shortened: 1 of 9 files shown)


jayzhan211:
LGTM

hicqu:
@jayzhan211 thank you! I have added unit test cases for the new implementation. Could you take a look again and approve it?

hicqu:
@jayzhan211 or could you push more commits to #423 ? Your implementation involves less allocation and free. I can compare them later.

jayzhan211:
> @jayzhan211 or could you push more commits to #423 ? 

What commits do you mean?

src/tracker/inflights.rs
```diff
@@ -130,127 +164,63 @@ mod tests {
...
+            assert_eq!(inflights.offset, (offset + 1) % chunk_len);
+            assert_eq!(inflights.inflights(), (1..100).collect::<Vec<_>>());
+
+            // Free to the middle item in buffer[1].
+            let to = inflights.buffer[1][3];
+            inflights.free_to(to);
+            let count = match offset {
+                0 => 80,
+                8 => 88,
+                _ => 79,
```
jayzhan211, line 202: if `_` is chunk_len(15), why count is not 100 - ((16 - 15) + 4) = 95?

jayzhan211 (commented):
Rest LGTM

src/tracker/inflights.rs
jayzhan211: > > @jayzhan211 or could you push more commits to #423 ?
> 
> What commits do you mean?


If you mean adding test, I am not sure how to test `free_*` or `reallocate_*`

hicqu:
Here is a benchmark result:
for master branch:
```
Inflights(256)          time:   [2.1513 us 2.2001 us 2.2609 us]
Inflights(4096)         time:   [35.550 us 36.232 us 36.976 us]
```
for the patch:
```
Inflights(256)          time:   [6.8643 us 7.1716 us 7.5381 us]
                        change: [+205.14% +217.31% +228.40%] (p = 0.00 < 0.05)
                        Performance has regressed.
Inflights(4096)         time:   [103.92 us 108.21 us 113.07 us]
                        change: [+192.46% +201.84% +211.64%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

PR #450, opened 2021-07-30
support to adjust max inflight msgs

hicqu:

Sometimes we want to custom `max-inflight-msgs` for a given peer. For example, the target peer's memory or disk usage is almost full. This PR allows this.

BTW a bug in `inflight_buffers_size` is fixed. With the patch its return value is real **size** instead of **length**.

### DIFF
--- a/src/tracker/inflights.rs
+++ b/src/tracker/inflights.rs
@@ -56,22 +56,7 @@ impl Inflights {
                 self.cap = incoming_cap;
                 self.incoming_cap = None;
             }
-            Ordering::Greater => {
-                if self.count <= incoming_cap && self.start + self.count < incoming_cap {
-                    self.cap = incoming_cap;
-                    let (cur_cap, cur_len) = (self.buffer.capacity(), self.buffer.len());
-                    if cur_cap > incoming_cap && cur_len <= incoming_cap {
-                        // TODO: Simplify it after `shrink_to` is stable.
-                        unsafe {
-                            self.buffer.set_len(incoming_cap);
-                            self.buffer.shrink_to_fit();
-                            self.buffer.set_len(cur_len);
-                        }
-                    }
-                } else {
-                    self.incoming_cap = Some(incoming_cap);
-                }
-            }
+            Ordering::Greater => self.incoming_cap = Some(incoming_cap),
         }
     }
 
@@ -383,18 +368,9 @@ mod tests {
         inflight.set_cap(1024);
         assert_eq!(inflight.incoming_cap, None);
 
-        // 1024 -> 512. The internal buffer should be shrinked.
+        // 1024 -> 512. The internal buffer shouldn't be shrinked.
         inflight.set_cap(512);
-        assert_eq!(inflight.cap, 512);
-        assert_eq!(inflight.incoming_cap, None);
-        assert_eq!(inflight.buffer_capacity(), 512);
-
-        // 1024 -> 512. The buffer shouldn't be shrinked as the tail part is in used.
-        let mut inflight = Inflights::new(1024);
-        inflight.buffer = vec![0; 1024];
-        inflight.start = 800;
-        (0..16).for_each(|i| inflight.add(i));
-        inflight.set_cap(512);
+        assert_eq!(inflight.cap, 1024);
         assert_eq!(inflight.incoming_cap, Some(512));
         assert_eq!(inflight.buffer_capacity(), 1024);
     }
--- a/src/tracker/inflights.rs
+++ b/src/tracker/inflights.rs
@@ -52,11 +52,24 @@ impl Inflights {
         match self.cap.cmp(&incoming_cap) {
             Ordering::Equal => self.incoming_cap = None,
             Ordering::Less => {
-                self.buffer.reserve(incoming_cap - self.count);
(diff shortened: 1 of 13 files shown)


src/tracker/inflights.rs
```diff
@@ -37,13 +42,38 @@ impl Inflights {
...
+            Ordering::Greater => {
+                if self.count == 0 {
+                    debug_assert!(self.incoming_cap.is_none());
+                    self.cap = incoming_cap;
+                    if self.buffer.capacity() > 0 {
+                        self.buffer = Vec::with_capacity(incoming_cap);
+                    }
+                } else {
+                    // TODO: is it better to shrink if count is less than incoming_cap?
+                    self.incoming_cap = Some(incoming_cap);
```
jayzhan211, line 67: How about enabling both of them, at least for future testing?

src/tracker/inflights.rs
```diff
@@ -37,13 +42,38 @@ impl Inflights {
...
+    pub fn set_cap(&mut self, incoming_cap: usize) {
+        match self.cap.cmp(&incoming_cap) {
+            Ordering::Equal => self.incoming_cap = None,
+            Ordering::Less => {
+                self.buffer.reserve(incoming_cap - self.count);
+                self.cap = incoming_cap;
+                self.incoming_cap = None;
+            }
+            Ordering::Greater => {
+                if self.count == 0 {
```
BusyJay, line 59: It seems it has to allocate in the end. So it's simpler to just do allocation now instead of adding a new field.

src/tracker/inflights.rs
jayzhan211: In my opinion, we should shrink the buffer in this case, so we can enforce reallocating memory in case we are running out of space.

src/tracker/inflights.rs
hicqu: The new field is `incoming_cap`? I think it's required. Please consider calling `set_cap(x)` when `self.count > x`.

src/tracker/inflights.rs
hicqu: It's updated. PTAL again, thank you!

src/tracker/inflights.rs
```diff
@@ -37,13 +42,43 @@ impl Inflights {
...
+                self.incoming_cap = None;
+            }
+            Ordering::Greater => {
+                if self.count <= incoming_cap && self.start + self.count < incoming_cap {
+                    self.cap = incoming_cap;
+                    let (cur_cap, cur_len) = (self.buffer.capacity(), self.buffer.len());
+                    if cur_cap > incoming_cap && cur_len <= incoming_cap {
+                        // TODO: Simplify it after `shrink_to` is stable.
+                        unsafe {
+                            self.buffer.set_len(incoming_cap);
```
BusyJay, line 66: I don't think it's a good idea. It will introduce read uninitialized undefined behavior.

src/tracker/inflights.rs
```diff
@@ -37,13 +42,28 @@ impl Inflights {
...
+        }
+    }
+
+    /// Adjust inflight buffer capacity. Set it to `0` will disable the progress.
+    // Calling it between `self.full()` and `self.add()` can cause a panic.
+    pub fn set_cap(&mut self, incoming_cap: usize) {
+        match self.cap.cmp(&incoming_cap) {
+            Ordering::Equal => self.incoming_cap = None,
+            Ordering::Less => {
+                self.buffer.reserve(incoming_cap - self.count);
```
BusyJay, line 55: I think it should be `self.buffer.len()`.

src/tracker/inflights.rs
```diff
@@ -37,13 +42,28 @@ impl Inflights {
...
+                self.cap = incoming_cap;
+                self.incoming_cap = None;
+            }
+            Ordering::Greater => self.incoming_cap = Some(incoming_cap),
         }
```
BusyJay, line 60: If it's empty already, shrinking may never happen.

BusyJay (dismissed):

hicqu:
@jayzhan211 could you take a look again?

jayzhan211 (commented):
LGTM

jayzhan211 (approved):
LGTM

src/tracker/inflights.rs
```diff
@@ -37,13 +42,41 @@ impl Inflights {
...
+        }
+    }
+
+    /// Adjust inflight buffer capacity. Set it to `0` will disable the progress.
+    // Calling it between `self.full()` and `self.add()` can cause a panic.
+    pub fn set_cap(&mut self, incoming_cap: usize) {
+        match self.cap.cmp(&incoming_cap) {
+            Ordering::Equal => self.incoming_cap = None,
+            Ordering::Less => {
+                self.cap = incoming_cap;
```
NingLin-P, line 55: We use a ring buffer mechanism, and calculate the next index base on `self.cap` in `fn add`, direct changing `self.cap` may cause us to get a wrong next index.

NingLin-P (approved):
LGTM

NingLin-P (approved):

BusyJay (approved):

NingLin-P (approved):

PR #451, opened 2021-08-02
Fix protobuf dependency invalid semver for cargo 1.54 for v0.5.x

jeanphilippeD:
With this, https://github.com/rust-lang/cargo/pull/9508, the dependency can no longer be used.
What was picked with 1.53 cargo was ~2.0, so use that instead.

**This needs to be a new branch 0.5.1 that I branched from tag v0.5.0**

Test:
Verify dependency can now be used.
protobuf in cargo.lock is unchanged in my dependent project.


BusyJay (approved):
Thanks!

BusyJay:
You need to sign off your commits.

jeanphilippeD:
Thanks, Added the sign off line in the commit now.

PR #452, opened 2021-08-05
adjust_max_inflight_msgs shouldn't make program panic

hicqu:


NingLin-P (approved):

BusyJay (approved):
Hope such complicated feature can be really utilized one day.

PR #454, opened 2021-08-18
减少一点括号

egmkang:
如题


hicqu:
@egmkang can you fix DCO check? It's required to use `git commit -s` option.

disksing (approved):

PR #459, opened 2021-12-15
Support fetch raft entry in async way

Connor1996:
Support fetch entry in async way, here are some changes:
- Add an error `LogTemporarilyUnavailable`
- Add a field `context` to `entries()`
   - Some process still need to fetch the entries in sync way, add a constraint to limit the condition when async fetch is permitted.
   - when entries are fetched asynchronously , it should return error `LogTemporarilyUnavailable`
- Add a public function `on_entries_fetched` for triggering refetch the logs after async log fetch is finished.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "raft"
-version = "0.6.1"
+version = "0.7.0"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
 keywords = ["raft", "distributed-systems", "ha"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "raft"
-version = "0.6.0"
+version = "0.6.1"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
 keywords = ["raft", "distributed-systems", "ha"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ fail = { version = "0.4", optional = true }
 getset = "0.1.1"
 protobuf = "2"
 thiserror = "1.0"
-raft-proto = { path = "proto", version = "0.6.0", default-features = false }
+raft-proto = { path = "proto", version = "0.7.0", default-features = false }
 rand = "0.8"
 slog = "2.2"
 slog-envlogger = { version = "2.1.0", optional = true }
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -760,6 +760,12 @@ impl<T: Storage> RaftCore<T> {
         self.maybe_send_append(to, pr, true, msgs);
     }
 
+    fn send_append_aggressively(&mut self, to: u64, pr: &mut Progress, msgs: &mut Vec<Message>) {
+        // If we have more entries to send, send as many messages as we
+        // can (without sending empty messages for the commit index)
+        while self.maybe_send_append(to, pr, false, msgs) {}
+    }
+
     /// Sends an append RPC with new entries to the given peer,
     /// if necessary. Returns true if a message was sent. The allow_empty
     /// argument controls whether messages with no entries will be sent
@@ -792,7 +798,10 @@ impl<T: Storage> RaftCore<T> {
             let ents = self.raft_log.entries(
                 pr.next_idx,
                 self.max_msg_size,
-                GetEntriesContext(GetEntriesFor::SendAppend { to }),
+                GetEntriesContext(GetEntriesFor::SendAppend {
+                    to,
+                    aggressively: !allow_empty,
+                }),
             );
             if !allow_empty && ents.as_ref().ok().map_or(true, |e| e.is_empty()) {
                 return false;
@@ -859,15 +868,14 @@ impl<T: Storage> Raft<T> {
 
     /// Sends an append RPC with new entries (if any) and the current commit index to the given
     /// peer.
-    pub fn send_append(&mut self, to: u64, exhaust: bool) {
+    pub fn send_append(&mut self, to: u64) {
         let pr = self.prs.get_mut(to).unwrap();
-        if exhaust {
-            // If we have more entries to send, send as many messages as we
-            // can (without sending empty messages for the commit index)
-            while self.r.maybe_send_append(to, pr, false, &mut self.msgs) {}
-        } else {
-            self.r.send_append(to, pr, &mut self.msgs)
-        }
+        self.r.send_append(to, pr, &mut self.msgs)
+    }
+
+    pub(super) fn send_append_aggressively(&mut self, to: u64) {
+        let pr = self.prs.get_mut(to).unwrap();
+        self.r.send_append_aggressively(to, pr, &mut self.msgs)
     }
 
     /// Sends RPC, with entries to all peers that are not up-to-date
@@ -1739,7 +1747,7 @@ impl<T: Storage> Raft<T> {
                 if pr.state == ProgressState::Replicate {
                     pr.become_probe();
                 }
-                self.send_append(m.from, false);
+                self.send_append(m.from);
             }
             return;
         }
@@ -1770,22 +1778,21 @@ impl<T: Storage> Raft<T> {
                 self.bcast_append()
             }
         } else if old_paused {
-            self.send_append(m.from, false);
+            self.send_append(m.from);
         }
-        // Hack to get around borrow check. It may be possible to move L1448 above L1433 to
-        // get around the problem. But here choose to keep consistent with Etcd.
-        let pr = self.prs.get_mut(m.from).unwrap();
+
         // We've updated flow control information above, which may
         // allow us to send multiple (size-limited) in-flight messages
         // at once (such as when transitioning from probe to
         // replicate, or when freeTo() covers multiple messages). If
         // we have more entries to send, send as many messages as we
         // can (without sending empty messages for the commit index)
-        while self.r.maybe_send_append(m.from, pr, false, &mut self.msgs) {}
+        self.send_append_aggressively(m.from);
 
         // Transfer leadership is in progress.
         if Some(m.from) == self.r.lead_transferee {
             let last_index = self.r.raft_log.last_index();
+            let pr = self.prs.get_mut(m.from).unwrap();
             if pr.matched == last_index {
                 info!(
                     self.logger,
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -756,8 +756,13 @@ impl<T: Storage> RaftCore<T> {
 
     /// Sends an append RPC with new entries (if any) and the current commit index to the given
     /// peer.
-    fn send_append(&mut self, to: u64, pr: &mut Progress, msgs: &mut Vec<Message>) -> bool {
-        self.maybe_send_append(to, pr, true, msgs)
+    fn send_append(&mut self, to: u64, pr: &mut Progress, msgs: &mut Vec<Message>) {
+        // Sending multiple (size-limited) in-flight messages at once
+        // are allowed (such as when transitioning from probe to
+        // replicate, or when freeTo() covers multiple messages). If
+        // we have more entries to send, send as many messages as we
+        // can (without sending empty messages for the commit index)
+        while self.maybe_send_append(to, pr, false, msgs) {}
     }
 
     /// Sends an append RPC with new entries to the given peer,
@@ -857,9 +862,9 @@ impl<T: Storage> Raft<T> {
 
     /// Sends an append RPC with new entries (if any) and the current commit index to the given
     /// peer.
-    pub fn send_append(&mut self, to: u64) -> bool {
+    pub fn send_append(&mut self, to: u64) {
         let pr = self.prs.get_mut(to).unwrap();
-        self.r.send_append(to, pr, &mut self.msgs)
+        self.r.send_append(to, pr, &mut self.msgs);
     }
 
     /// Sends RPC, with entries to all peers that are not up-to-date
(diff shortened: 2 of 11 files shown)


Connor1996:
This PR is ready now, PTAL @BusyJay @gengliqi @NingLin-P

tonyxuqqi:
Could you add some detail about the issue this PR tries to solve?

Connor1996:
@tonyxuqqi https://github.com/tikv/tikv/issues/11320

src/storage.rs
```diff
@@ -56,6 +56,23 @@ impl RaftState {
     }
 }
 
+/// Records the context of the caller who calls entries() of Storage trait.
+pub enum GetEntriesContext {
```
BusyJay, line 60: How about:
```
enum GetEntriesFor {
    ...
}
pub struct GetEntriesContext(GetEntriesFor);
```
? So user won't be able to rely on any details of the context.

src/storage.rs
```diff
@@ -75,10 +92,24 @@ pub trait Storage {
     /// the slice of entries returned will always have length at least 1 if entries are
     /// found in the range.
     ///
+    /// If the entries are fetched asynchorously, it would return LogTemporarilyUnavailable,
+    /// and application needs to call `on_entries_fetched(context)` to trigger re-fetch of the entries
+    /// after the storage finishes fetching the entries.   
+    ///
+    /// Constraint: only [low, high] covers unapplied entries are permitted to fetch entries
+    /// asynchorously, otherwise some callers may panic when meeting LogTemporarilyUnavailable.
```
BusyJay, line 100:
```suggestion
    /// Constraint: due to implementation limits, currently only when [low, high] covers any applied
    /// entries are permitted to fetch entries asynchorously, otherwise it may panic.
```

src/storage.rs
```diff
@@ -102,7 +133,8 @@ pub trait Storage {
     /// so raft state machine could know that Storage needs some time to prepare
     /// snapshot and call snapshot later.
     /// A snapshot's index must not less than the `request_index`.
-    fn snapshot(&self, request_index: u64) -> Result<Snapshot>;
+    /// `to` indicates which peer is requesting the snapshot, and it's only used for printing log.
```
BusyJay, line 136:
```suggestion
    /// `to` indicates which peer is requesting the snapshot.
```

Connor1996:
one question: why the clippy check is failing. It passes on my local env, and seems what clippy complains doesn't related to this PR's changes. @BusyJay

BusyJay:
You may need to `rustup update` first.

src/raft_log.rs
```diff
@@ -580,9 +586,10 @@ impl<T: Storage> RaftLog<T> {
 
         if low < self.unstable.offset {
             let unstable_high = cmp::min(high, self.unstable.offset);
-            match self.store.entries(low, unstable_high, max_size) {
+            match self.store.entries(low, unstable_high, max_size, context) {
                 Err(e) => match e {
                     Error::Store(StorageError::Compacted) => return Err(e),
+                    Error::Store(StorageError::LogTemporarilyUnavailable) => return Err(e),
```
NingLin-P, line 592:
```suggestion
                    Error::Store(StorageError::Compacted) | Error::Store(StorageError::LogTemporarilyUnavailable) => return Err(e),
```

src/raft_log.rs
```diff
@@ -379,19 +379,24 @@ impl<T: Storage> RaftLog<T> {
...
         }
-        self.slice(idx, last + 1, max_size)
+        self.slice(idx, last + 1, max_size, context)
     }
 
     /// Returns all the entries.
     pub fn all_entries(&self) -> Vec<Entry> {
         let first_index = self.first_index();
-        match self.entries(first_index, None) {
+        match self.entries(first_index, None, GetEntriesContext::Test) {
```
NingLin-P, line 399: This may imply `all_entries` must only use in test, how about passing `GetEntriesContext::Test` as a parameter to `all_entries` instead.

Connor1996: But seems `all_entries` is indeed only used by tests.

NingLin-P: Yes, but nothing can stop people use it outside of tests.

NingLin-P: Or if this function is dedicated to test, it may be better to update the comment or change the name to something like `all_entries_test`

src/storage.rs
```diff
@@ -75,10 +92,24 @@ pub trait Storage {
     /// the slice of entries returned will always have length at least 1 if entries are
     /// found in the range.
     ///
+    /// If the entries are fetched asynchorously, it would return LogTemporarilyUnavailable,
+    /// and application needs to call `on_entries_fetched(context)` to trigger re-fetch of the entries
+    /// after the storage finishes fetching the entries.   
+    ///
+    /// Constraint: only [low, high] covers unapplied entries are permitted to fetch entries
```
NingLin-P, line 99: It may be a burden for the user to determine `[low, high] covers unapplied entries`, and there may also have use cases that required fetching unapplied logs synchronously in the future.

How about providing a helper function like `GetEntriesContext::can_async` instead? /cc @BusyJay

BusyJay: It's discussed before that changing the same function from asynchronous to synchronous depends on the parameters can be weird. Unapplied logs are usually stored in memory, so it's unlikely need to be fetched asynchronously. Note synchronous version is always compatible with the asynchronous version, so even it's supported in the future, no breaking change is introduced.

src/raft_log.rs
Connor1996: will update the comment

src/storage.rs
NingLin-P: > It's discussed before that changing the same function from asynchronous to synchronous depends on the parameters can be weird. 

Using sync or async never depends on the parameter but the user, namely how the user implements `Storage::entries`. The user can always ignore the `GetEntriesContext` parameter and fetch logs in sync way, if the user wants the performance gain by fetch logs asynchronously, `GetEntriesContext::can_async` can tell whether `raft-rs` support doing so.  /cc @tonyxuqqi 

> Unapplied logs are usually stored in memory, so it's unlikely need to be fetched asynchronously.

Again, it is very up to the user's implementation of the `Storage` trait.

> Note synchronous version is always compatible with the asynchronous version, so even it's supported in the future, no breaking change is introduced.

What I mean is the usage inside `raft-rs`.  For example, `self.raft_log.entries(l, h, ctx).unwrap()` where `[l. h]` only covers unapplied entries so the user may think it is okay to fetch logs asynchronously, and thus get an unexpected panic. Whether fetching log asynchronously is allowed is up to the usage inside `raft-rs` not other conditions.

BusyJay: > ...where [l. h] only covers unapplied entries...

However such usage doesn't exist. Raft query unapplied logs only when sending logs to followers.

NingLin-P: Even the correct constraint can still meet the abovementioned problem.

Connor1996: ping

Connor1996: `can_async` is quite like the previous implementation which has the issue that changing the same function from asynchronous to synchronous depends on the parameters, though I'm still not sure if it's really a problem. And the constraint is not a good way either in my point of view. Both are acceptable for me now. 😑

BusyJay: > It could easily be broken by future use cases or future use cases may have to subject to this unreasonable constraint.

My suggestion is to remove the constraint in the end.

Connor1996: Encounter a problem that some other places TiKV utilize `entries()` as well. They need to pass a context to make sure not to use async way. So `can_async` is needed. I think we waste too much time on it and it makes everything complicated to avoid the so-called style problem.

BusyJay: I checked the code, and it's indeed hard to solve. Perhaps we do need to expose `can_async` and context definition, at least constructor. I'm now OK with the changes.

/cc @tonyxuqqi what's your suggestion?

NingLin-P: IMO, there are two ways to implement this feature, either integrated into `Storage::entries` or a new interface like `Storage::entries_async`.

With `Storage::entries`, as there are cases that allow using the async way and cases that **must** use the sync way, we must pass some information to let `Storage::entries` know what the caller expected.  And I think helper function like `can_async` is a more clear, straightforward and robust way to do so. (actually, the abovementioned constraint also rely on the parameter, namely the `[low, high]` range)

With `Storage::entries_async`, although make the `Storage` trait look more redundant, both `entries` and `entries_async` keep in simple and much clear than any parameter. With a default implement like:
```
fn entries_async(...) {
    self.entries(...)
}
```
there are only very small changes to the codebase, with no breaking change in the interface and backward compatibility.

Both ways are okay to me.

tonyxuqqi:
@NingLin-P @BusyJay @Connor1996 I still keep my original thought that having two API separately is better. To me the context parameter seems more complex than can_sync,  though it's being used almost the same way of "can_async". For the 3rd party developer they might not understand how to handle this context parameter correctly.  (I bet it's easier to implement async_entries() for them). And also as NingLin mentioned, "it will have very small changes to code base, no breaking changes and backward compatibility"

But I agree with BoKang we spend too much time on this. As long as it's correct, I'm fine.

src/storage.rs
Connor1996: After all the callers support async way, the `entries_async` is verbose. So I choose to provide a `can_async`.

Connor1996:
`can_async` is added for `GetEntriesContext`, and bump the version to `0.6.1`. PTAL @BusyJay @NingLin-P @tonyxuqqi

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "raft"
-version = "0.6.0"
+version = "0.6.1"
```
BusyJay, line 3: It should be "0.7".

src/raft_log.rs
```diff
@@ -379,19 +379,24 @@ impl<T: Storage> RaftLog<T> {
...
         if idx > last {
             return Ok(Vec::new());
         }
-        self.slice(idx, last + 1, max_size)
+        self.slice(idx, last + 1, max_size, context)
     }
 
-    /// Returns all the entries.
+    /// Returns all the entries. Only used by tests.
     pub fn all_entries(&self) -> Vec<Entry> {
```
BusyJay, line 397: How about marking it as private?

src/raft.rs
```diff
@@ -851,9 +859,15 @@ impl<T: Storage> Raft<T> {
 
     /// Sends an append RPC with new entries (if any) and the current commit index to the given
     /// peer.
-    pub fn send_append(&mut self, to: u64) {
+    pub fn send_append(&mut self, to: u64, exhaust: bool) {
```
BusyJay, line 862: How about using a new method? So you don't need to change so many codes and the new method can be marked as private.

src/raw_node.rs
```diff
@@ -410,6 +410,14 @@ impl<T: Storage> RawNode<T> {
         Err(Error::StepPeerNotFound)
     }
 
+    /// A callback when entries are fetched asynchronously .
+    pub fn on_entries_fetched(&mut self, context: GetEntriesContext) {
+        match context.0 {
+            GetEntriesFor::SendAppend { to } => self.raft.send_append(to, true),
```
BusyJay, line 416: Better use the context to decide whether send all pending entries.

src/raft_log.rs
Connor1996: it's used in `harness/test` which is out of the crate. So can't make it private

BusyJay: Then maybe can make it doc hidden.

BusyJay (approved):

src/raft.rs
```diff
@@ -789,7 +795,14 @@ impl<T: Storage> RaftCore<T> {
                 return false;
             }
         } else {
-            let ents = self.raft_log.entries(pr.next_idx, self.max_msg_size);
+            let ents = self.raft_log.entries(
+                pr.next_idx,
+                self.max_msg_size,
+                GetEntriesContext(GetEntriesFor::SendAppend {
+                    to,
+                    aggressively: !allow_empty,
```
NingLin-P, line 803: Why let `allow_empty` decide whether send append log aggressively?

src/raw_node.rs
```diff
@@ -410,6 +410,20 @@ impl<T: Storage> RawNode<T> {
         Err(Error::StepPeerNotFound)
     }
 
+    /// A callback when entries are fetched asynchronously .
```
NingLin-P, line 413: Should mention the `context` parameter is the one pass to `Storage::entries` previously, and how this function may panic.

NingLin-P (commented):
Rest LGTM

src/raft.rs
Connor1996: `allow_empty` only set to `false` when sending log aggressively, you can check the comment of `maybe_send_append`

src/storage.rs
```diff
@@ -56,6 +56,45 @@ impl RaftState {
...
+        aggressively: bool,
+    },
+    // for getting committed entries in a ready
+    GenReady,
+    // for getting entries to check pending conf when transferring leader
+    TransferLeader,
+    // for getting entries to check pending conf when forwarding commit index by vote messages
+    CommitByVote,
+    // It's not called by the raft itself
+    Empty(bool),
```
NingLin-P, line 95: Do we need `Empty(true)`? it is a noop for `on_entries_fetched`

Connor1996: Yes, TiKV needs it to test the process of async fetch

BusyJay (approved):

NingLin-P (approved):

PR #462, opened 2022-02-10
Fix async entries fetched callback panic on removed node

Connor1996:
If there has conf change been applied before the async fetched callback is called, it would be panic due to missing progress for the peer. 
 
image

So this PR checks the validation of the peer id in the context

### DIFF
--- a/harness/tests/integration_cases/test_raw_node.rs
+++ b/harness/tests/integration_cases/test_raw_node.rs
@@ -932,7 +932,7 @@ fn test_raw_node_with_async_entries() {
     let _ = raw_node.advance_append(rd);
 }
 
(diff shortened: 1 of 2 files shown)


harness/tests/integration_cases/test_raw_node.rs
```diff
@@ -932,6 +932,60 @@ fn test_raw_node_with_async_entries() {
     let _ = raw_node.advance_append(rd);
 }
 
+// Test entries are handled properly when they are fetched asynchronously
```
BusyJay, line 935: Please be more specific.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

NingLin-P (approved):

PR #465, opened 2022-02-28
Ignore async entries when leadership or term is changed

Connor1996:
To make raft-rs robust, when calling `on_entries_fetched`, ignore the entries when leadership or term is changed. Otherwise , it may cause potential panics in `send_append`.

### DIFF
--- a/src/raw_node.rs
+++ b/src/raw_node.rs
@@ -419,11 +419,20 @@ impl<T: Storage> RawNode<T> {
     /// Panics if passed with the context of context.can_async() == false
     pub fn on_entries_fetched(&mut self, context: GetEntriesContext) {
         match context.0 {
-            GetEntriesFor::SendAppend { to, aggressively } => {
+            GetEntriesFor::SendAppend {
+                to,
+                term,
+                aggressively,
+            } => {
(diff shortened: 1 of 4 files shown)


Connor1996:
PTAL @BusyJay @NingLin-P

src/raw_node.rs
```diff
@@ -419,11 +419,20 @@ impl<T: Storage> RawNode<T> {
...
+            GetEntriesFor::SendAppend {
+                to,
+                term,
+                aggressively,
+            } => {
                 if self.raft.prs().get(to).is_none() {
                     // the peer has been removed, do nothing
                     return;
                 }
+                if self.raft.term != term || self.raft.state != StateRole::Leader {
```
BusyJay, line 431: Should be checked before L427. Actually, role changes implies term changes.

BusyJay: Oh, seems quorum active check will make it step down without changing its term. In this case, it should also not send any entries to other peers. Should add a case to cover that.

BusyJay (approved):

BusyJay:
clippy error should be fixed by #466.

NingLin-P (approved):

PR #466, opened 2022-03-02
proto: update prost to 0.9

### LINKED ISSUE #449: RUSTSEC-2021-0073 in prost 0.7
[RUSTSEC-2021-0073](https://rustsec.org/advisories/RUSTSEC-2021-0073.html) was issued recently against prost 0.7, which is used by the `prost-codec` feature flag. It's fixed in prost 0.8, which has some backwards-incompatible changes but not much.

I believe the version of prost can be updated without any code changes. Unfortunately it is technically a breaking change as public raft types implement `prost::Message`, and any clients who use this impl will therefore be broken by the prost update. This could be mitigated going forward by re-exporting `prost` itself and documenting that backward compatibility of the prost-codec feature requires clients who wish to use the `prost::Message` impls to use the re-exported prost. Or duplicate the types so the ones that implement `prost::Message` are private. But for this particular update, the technically-correct thing to do is update the version to 0.7.

The rationale for going ahead and doing this now is anyone using [`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny) will by default get an error about this RUSTSEC.

BusyJay:
Close #449.

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -109,7 +109,7 @@ fn next_ents(r: &mut Raft<MemStorage>, s: &MemStorage) -> Vec<Entry> {
     }
     let ents = r.raft_log.next_entries(None);
     r.commit_apply(r.raft_log.committed);
-    ents.unwrap_or_else(Vec::new)
+    ents.unwrap_or_default()
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


Connor1996 (approved):
LGTM

NingLin-P (approved):

PR #469, opened 2022-03-07
add prev_lead_transferee into Ready

hicqu:

In TiKV when a leader peer is in a leadership transferring, it reports a `NotLeader` error to all CDC downstreams. Currently it doesn't know which peer will be the next leader, so CDC clients will resubscribe the region from any peers blindly.

With the patch we can return the potential next leader of the region so that clients can retry requests with a better strategy. Here is how the patch is used in TiKV: https://github.com/tikv/tikv/pull/12124.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -201,6 +201,9 @@ pub struct RaftCore<T: Storage> {
     /// If this is Some(id), we follow the procedure defined in raft thesis 3.10.
     pub lead_transferee: Option<u64>,
 
+    /// Just like `lead_transferee` but only used to construct a `Ready`.
+    pub(crate) prev_lead_transferee: Option<u64>,
+
     /// Only one conf change may be pending (in the log, but not yet
     /// applied) at a time. This is enforced via `pending_conf_index`, which
     /// is set to a value >= the log index of the latest pending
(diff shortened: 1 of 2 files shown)


BusyJay:
Can you explain why the change is needed?

hicqu:
> Can you explain why the change is needed?

Yes I have updated the PR description.

BusyJay:
How about implementing it on the TiKV side? It should be easy to store the value in `Peer`.

overvenus:
> How about implementing it on the TiKV side? It should be easy to store the value in `Peer`.

Do you mean something like that https://github.com/tikv/tikv/pull/12110?

hicqu:
> It should be easy to store the value in Peer

We met some difficulties for this. In TiKV when we get the role change event from `Ready`, `raft.lead_transferee` has already been reset. And not only role change can change `raft.lead_transferee`. Maintain this in TiKV need to cover all those cases.

BusyJay:
> Do you mean something like that https://github.com/tikv/tikv/pull/12110? It adds a hook on_transfer_leader in RoleObserver.

Yes.

> And not only role change can change raft.lead_transferee. Maintain this in TiKV need to cover all those cases.

Is it necessary to get correct `transferee`? Isn't it just a hint? I suggest to store transferee in `Peer`, and whenever the leader is step down, pass the transferee to role changed event and clear the field.

BusyJay:
Or perhaps the existing hard state can be used directly. The next leader is likely to be either `hs.vote` or `ss.leader_id`.

hicqu:
I moved the state into TiKV raftstore peer. You can take a look at https://github.com/tikv/tikv/pull/12124 if you are free @BusyJay

PR #475, opened 2022-04-21
raft:Fix one log append reject bug about MsgUnreachable msg

tier-cap:
Problem Description：
close tikv/tikv#11371
1. When follower reject the appendlog requests, it will send a MsgUnreachable msg to leader
2. But the MsgUnreachable is a local type which means only be used in inner raftcore, cannot transfered by net, which leads to some logic check failed.

Solution:
1. Let the MsgUnreachable be a response message, and when leader receives, call the raft core step directly.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -265,6 +265,7 @@ pub struct RaftCore<T: Storage> {
 /// A struct that represents the raft consensus itself. Stores details concerning the current
 /// and possible state the system can take.
 pub struct Raft<T: Storage> {
+    /// the progress tracker of each peer.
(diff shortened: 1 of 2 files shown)


BusyJay:
I don't think it's a good idea. Internal messages mean we are subject to change without breaking any compatibility. You should define a new message on the TiKV side.

tier-cap:
> I don't think it's a good idea. Internal messages mean we are subject to change without breaking any compatibility. You should define a new message on the TiKV side.

I have two fix, one is this pr, the other is add a new msg type. but the change is more tthan this.

PR #479, opened 2022-07-03
Fix a broken anchor link

IDJack:
The origin link(https://docs.rs/raft/latest/raft/examples/single_mem_node/main.rs#L113-L179) is broken. 

The path of the corresponding code in docs.rs is (https://docs.rs/crate/raft/0.6.0/source/examples/single_mem_node/main.rs), but the line number range cannot be specified.

I changed the link to point to the corresponding code on Github.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -416,7 +416,8 @@ to advance the applied index inside.
     node.advance_apply();
     ```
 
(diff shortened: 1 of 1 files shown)


tisonkun:
@IDJack This patch looks good to me. Can you fix the DCO requirement? I can approve and schedule a merge once it's donw.

PR #482, opened 2022-07-25
fix clippy error

BusyJay:


jayzhan211 (approved):
I forgot why I don't implement with `write!`

LGTM

PR #484, opened 2022-08-10
Add new message type and corresponding handler for group broadcast

LintianShi:
Modify raft-rs to support follower replication in TiKV.
Main change:
* Add new message type: MsgGroupBroadcast, which contains information of forwarding.
* Add handler for MsgGroupBroadcast, which appends entries to local log and forward MsgAppend to other peers.


LintianShi:
Split this PR into several PRs.

PR #485, opened 2022-08-29
Add new message type MsgGroupBroadcast and corresponding handler

LintianShi:
Signed-off-by: LintianShi <lintian.shi@pingcap.com>

Modify raft-rs to support follower replication in TiKV.
Main change:
- Add option of follower replication for `RawNode` and `Config`
- Add new message type: `MsgGroupBroadcast` and an extra field `forwards` in `Message`, which contains information that needs to be forwarded.
- Add handler for `MsgGroupBroadcast`, which appends entries to local log and forward `MsgAppend` to other peers.


LintianShi:
@Connor1996 PTAL

src/raft.rs
```diff
@@ -236,6 +236,13 @@ pub struct RaftCore<T: Storage> {
     /// Enable this if greater cluster stability is preferred over faster elections.
     pub pre_vote: bool,
 
+    /// Enable follower replication.
+    ///
+    /// This enables data replication from a follower to other servers in the same available zone.
+    ///
+    /// Enable this for reducing across-AZ traffic of cloud deployment.
+    pub follower_repl: bool,
```
BusyJay, line 244: Why `pub`?

src/raft.rs
```diff
@@ -2893,4 +3036,36 @@ impl<T: Storage> Raft<T> {
             pr.ins.set_cap(cap);
         }
     }
+
+    /// Whether this RawNode is active recently.
+    pub fn is_recent_active(&self, id: u64) -> bool {
+        if let Some(pr) = self.prs().get(id) {
+            if pr.recent_active {
```
BusyJay, line 3043: You mean `self.prs().get(id).map_or(false, |pr| pr.recent_active)`?

BusyJay (commented):
When will `MsgGroupBroadcast` be set?

src/raft.rs
LintianShi: Yes, I will modify it later.

src/raft.rs
LintianShi: `raftstore` will check whether follower replication is enabled in order to merge `MsgAppend` in `fn build_raft_messages`.

src/raft.rs
```diff
@@ -2483,9 +2522,90 @@ impl<T: Storage> Raft<T> {
...
+                            m_append.from = m.get_from();
+                            m_append.set_msg_type(MessageType::MsgAppend);
+                            m_append.index = forward.get_index();
+                            m_append.log_term = forward.get_log_term();
+                            m_append.set_entries(ents.into());
+                            m_append.commit = m.get_commit();
+                            m_append.commit_term = m.get_commit_term();
+                            self.r.send(m_append, &mut self.msgs);
+                        }
+                        Err(_) => {
```
Connor1996, line 2572: consider temporary unavailable

src/raft.rs
```diff
@@ -2458,7 +2497,7 @@ impl<T: Storage> Raft<T> {
         } else {
             debug!(
                 self.logger,
-                "rejected msgApp [logterm: {msg_log_term}, index: {msg_index}] \
+                "Reject append [logterm: {msg_log_term}, index: {msg_index}] \
```
Connor1996, line 2500: use lowercase for the initial letter, same for other places to fit the log style

src/raft.rs
LintianShi: Is the processing like the way in `fn send_append_aggressively`, looping for fetching until it success?

src/raft.rs
```diff
@@ -2483,9 +2522,90 @@ impl<T: Storage> Raft<T> {
...
+    }
+
+    // TODO: revoke pub when there is a better way to test.
+    /// For a given message, append the entries to the log.
+    pub fn handle_append_entries(&mut self, m: &Message) {
+        self.try_append_entries(m);
+    }
+
+    // For a broadcast, append entries to onw log and forward MsgAppend to other dest.
+    fn handle_group_broadcast(&mut self, m: &Message) {
```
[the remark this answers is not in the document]
LintianShi: In raftstore, several MsgAppend will be merged into a MsgGroupBroadcast in `fn build_raft_messages`

src/raft.rs
Connor1996: nope, call it again in `on_entries_fetched`

src/raft.rs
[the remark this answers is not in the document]
LintianShi: It is used in raftstore.

src/raft.rs
LintianShi: Async fetch for forwarding is committed. PTAL.

BusyJay:
Test case is missing.

src/config.rs
```diff
@@ -67,6 +67,10 @@ pub struct Config {
     /// rejoins the cluster.
     pub pre_vote: bool,
 
+    /// Enables follower replication.
+    /// This reduces the across-AZ traffic of cloud deployment.
+    pub follower_repl: bool,
```
BusyJay, line 72: Hard to understand `repl`.

src/storage.rs
```diff
@@ -56,9 +56,44 @@ impl RaftState {
     }
 }
 
+/// Purpose of getting entries.
+#[derive(Debug)]
+pub enum GetEntriesFor {
```
BusyJay, line 61: Why need to make it public?

src/tracker/progress.rs
```diff
@@ -203,6 +203,12 @@ impl Progress {
         true
     }
 
+    /// Determine whether progress is in the Replicate state;
+    #[inline]
+    pub fn is_replicate_state(&self) -> bool {
```
BusyJay, line 208:
```suggestion
    pub fn is_replicating(&self) -> bool {
```

src/raft.rs
BusyJay, line 244: Should be private.

src/config.rs
LintianShi: What about `follower_replication`?

src/storage.rs
LintianShi: In `fn on_raft_log_fetched`, we need to do different things according to the different type of `GetEntriesFor`. If `GetEntriesFor` is `SendAppend`, the peer must be the leader. But if `GetEntriesFor` is `SendForward`, the peer should not be the leader.

src/raft.rs
LintianShi: I see.

src/storage.rs
BusyJay: > If GetEntriesFor is SendAppend, the peer must be the leader. But if GetEntriesFor is SendForward, the peer should not be the leader.

Why? What if the role is changed when the logs are being fetched?

LintianShi: The agent is the follower. So we need to check whether it is still the follower but not the leader in `SendForward` situation. In addition, we check the term in `fn on_raft_log_fetched`. If the role is changed, the term will be changed too.

BusyJay: Does it matter? What could go wrong if always update the fetched result and clear it afterwards?

LintianShi: The peer selected as agent must be a follower. In the original `fn on_raft_log_fetched`, it first checks whether this peer is the leader. If this peer is not the leader now, it will return. Hence, if the async fetch result is temporarily unavailable, the callback function will never be called  successfully for the agent.
However, it does not affect the correctness of follower replication. It is just like that the agent always fetches log entries in sync way.

src/raft.rs
LintianShi: Resolved.

src/raft.rs
LintianShi: Resolved.

src/tracker/progress.rs
LintianShi: Resolved.

src/raft.rs
LintianShi: Resolved.

src/raft.rs
```diff
@@ -2483,9 +2599,51 @@ impl<T: Storage> Raft<T> {
...
+    pub fn handle_group_broadcast(&mut self, m: &Message) {
+        if self.try_append_entries(m) {
+            // If the agent fails to append entries from the leader,
+            // the agent cannot forward MsgAppend.
+            for forward in m.get_forwards() {
+                self.r
+                    .send_forward(m.from, m.commit, m.commit_term, forward, &mut self.msgs);
+            }
+        } else {
+            for forward in m.get_forwards() {
```
BusyJay, line 2625: What if the agent has conflict log sequence?

BusyJay:
Can you add some docs or examples to show how to use group broadcast?

src/raft.rs
LintianShi: According to Log Matching Property, the agent must have identical log entries up through the last index if the agent appends log entries from leader successfully.

BusyJay: This is `else` branch, which means append fail.

LintianShi: If append fails, the agent just sends an empty MsgAppend with no entry. Hence, conflict log sequence in the agent will not affect other followers.
If the follower that receives MsgAppend from the agent also has conflict log, this empty MsgAppend will trigger append reject due to mismatch of previous log, and leader can decrement next index. If the follower does not have conflict log, this empty MsgAppend truncates some uncommitted log entries in the worst case.

BusyJay: > If append fails, the agent just sends an empty MsgAppend with no entry.

Why? For example, the leader has (2, 3), (3, 4), (4, 4). The agent has (2, 3), (3, 3), (4, 3), (5, 3). Leader wants to send logs [2, 4) to a follower via this agent. Agent may fail to append the log, but it can send its conflict entries to follower.

LintianShi: If the agent has conflict log, it will fail to append. Then it goes to `else` branch. In `else` branch, the agent only sends an empty MsgAppend. 
In this example, if leader wants to send logs [4,5], the previous log in MsgGroupBroadcast is (5, 4) which is a check for the agent, and previous log in Forward is (3, 4) which is the beginning of log entries we want to send via agent. In other words, MsgGroupBroadcast is a MsgAppend with several Forward information. The part of MsgAppend guarantees the log consistency of agent, and Forward information records the range we want to send via agent.
When the agent receives MsgGroupBroadcast, it will first check whether the previous log (5, 4) in MsgGroupBroadcast and (5, 4) mismatches with agent's log. So the agent fail to append the log and goes to `else` branch which only sends empty MsgAppend.

BusyJay: If the agent is requesting snapshot, it will always reject the logs. Because the message append is forwarded to follower, so response will be sent back to leader. Leader will retry forwarding again. Hence dead loop.

LintianShi: If the agent is requesting snapshot, the progress state is `Snapshot`. And only peers in state `Replicate` will be involved with follower replication. If a peer is in other progress state, messages will be sent in leader replication. https://github.com/LintianShi/tikv/blob/80968da659f04bc172db8fbf497dbe4f034387ce/components/raftstore/src/store/peer.rs#L1953

BusyJay: The progress state is not `Snapshot`. When request snapshot explicitly, the state will be `Prob`. How can the library user possibly know the constraint?

LintianShi: Maybe we should add more comment on `fn handle_groupbroadcast` We can also write all things in RFC and give the link in comment. Since the whole processing of follower replication is separated in TiKV and raft-rs, it is a bit obscure.

BusyJay: The implementation should be more defensive.

LintianShi: What does 'defensive' mean? We need double-check in case of the library user possibly break the constraint?

BusyJay: Yes. We need to expose usable API and validate the input.

LintianShi:
> Can you add some docs or examples to show how to use group broadcast?

Add an example that shows how upper layer uses group broadcast.

src/raft.rs
```diff
@@ -836,6 +872,83 @@ impl<T: Storage> RaftCore<T> {
...
+                m.index = forward.get_index();
+                m.log_term = forward.get_log_term();
+                self.send(m, msgs);
+                warn!(
+                    self.logger,
+                    "The agent fails to fetch entries, index {} log term {} in forward message to peer {}.",
+                    forward.get_index(),
+                    forward.get_log_term(),
+                    forward.get_to()
+                );
```
tonyxuqqi, line 947: I think warn! could lead too much logs.

src/raft.rs
```diff
@@ -836,6 +872,83 @@ impl<T: Storage> RaftCore<T> {
...
+        );
+
+        match ents {
+            Ok(ents) => {
+                m.index = forward.get_index();
+                m.log_term = forward.get_log_term();
+                m.set_entries(ents.into());
+                self.send(m, msgs);
+            }
+            // TODO: Consider a better processing for temporary unavailable in async fetch,
```
tonyxuqqi, line 932: What about not using async fetch in this case? Since it's the follower replication, it's unnecessary to use async fetch which primarily is for reduce IO latency in raftstore thread.

src/raft.rs
```diff
@@ -836,6 +872,83 @@ impl<T: Storage> RaftCore<T> {
         true
     }
 
+    // Pack MsgAppend according to forward info, and send it to target peer.
+    fn send_forward(
```
tonyxuqqi, line 876: Questions: is it used for leader to resend information that follower fails to replicated? If that maybe resend_forward may be better?

src/raft.rs
```diff
@@ -2483,9 +2633,64 @@ impl<T: Storage> Raft<T> {
...
+    pub fn handle_append_entries(&mut self, m: &Message) {
+        self.try_append_entries(m);
+    }
+
+    /// For a group broadcast, append entries to local log and forward MsgAppend to other dest.
+    /// The usage of group broadcast is in examples/follower_replication/main.rs.
+    pub fn handle_group_broadcast(&mut self, m: &Message) {
+        // The agent should handle appending log entries in MsgGroupBroadcast firstly, in order to
+        // guarantee that agent's local raft log is identical to leader's log up through agent's last index.
+        // If the agent fails to append log entries, agent's log cannot be used for frowarding.
```
tonyxuqqi, line 2654: typo: frowarding

src/raft.rs
tonyxuqqi, line 932: Since it's sending logs by follower, I think async fetch can be skipped. Just use sync fetch is fine.

src/storage.rs
```diff
@@ -70,6 +70,7 @@ impl GetEntriesContext {
     pub fn can_async(&self) -> bool {
         match self.0 {
             GetEntriesFor::SendAppend { .. } => true,
+            GetEntriesFor::SendForward { .. } => true,
```
tonyxuqqi, line 73: I think it can be false to avoid the async fetch's potential latency in follower replication.

src/raft.rs
```diff
@@ -884,16 +998,137 @@ impl<T: Storage> Raft<T> {
...
-        let core = &mut self.r;
-        let msgs = &mut self.msgs;
-        self.prs
-            .iter_mut()
-            .filter(|&(id, _)| *id != self_id)
-            .for_each(|(id, pr)| core.send_append(*id, pr, msgs));
+        let mut msgs: Vec<Message> = Vec::default();
+        {
+            // Messages are stored in a vector temporarily.
+            // They will be pushed to message queue later.
```
Connor1996, line 1060: Seems `send_append` still pushes the messages to `self.msgs`

src/raft.rs
```diff
@@ -884,16 +998,137 @@ impl<T: Storage> Raft<T> {
...
+            // Get the broadcast group id of target peer.
+            let group_id = self.prs().get(msg.to).map_or(0, |pr| pr.broadcast_group_id);
+            // Do not need merge if broadcast group id is unknown or in the same
+            // group with leader. Valid broadcast group id should be greater than 0.
+            if group_id == 0 || group_id != leader_group_id {
+                continue;
+            }
+
+            // Group messages
+            if let Some(group) = msg_group.get_mut(&group_id) {
```
Connor1996, line 1106: `msg_group.entry().or_default().and_modify()`

src/raft.rs
```diff
@@ -884,16 +998,137 @@ impl<T: Storage> Raft<T> {
...
+        }
+
+        // Merge MsgAppend in broadcast groups.
+        for (_, group) in msg_group {
+            self.merge_append_group(&group, &mut msgs, &mut skip);
+        }
+
+        let mut idx: usize = 0;
+        for msg in msgs {
+            if !skip[idx] {
```
Connor1996, line 1120: How about grouping the messages instead of the pos. And `merge_append_group` modifies or filters the messages directly instead of by passing `skip`

src/raft.rs
Connor1996: Nope, sync-fetch blocks the raftstore which affects other leaders on that store.

src/raft.rs
```diff
@@ -836,6 +873,83 @@ impl<T: Storage> RaftCore<T> {
...
+        forward: &Forward,
+        msgs: &mut Vec<Message>,
+    ) {
+        // initialize MsgAppend
+        let mut m = Message::default();
+        m.to = forward.to;
+        m.from = from;
+        m.commit = commit;
+        m.commit_term = commit_term;
+        m.set_msg_type(MessageType::MsgAppend);
```
Connor1996, line 891: Seems we can set here
```
m.index = forward.get_index();
m.log_term = forward.get_log_term();
```

LintianShi: Resolved in new commit

src/raft.rs
LintianShi: In new commit, I have modified it. PTAL.

src/raft.rs
```diff
@@ -517,6 +554,31 @@ impl<T: Storage> Raft<T> {
         self.prs().group_commit()
     }
 
+    /// Checks whether the raft group is using group commit and consistent
```
BusyJay, line 557: Why move it around?

src/raft.rs
```diff
@@ -884,16 +994,132 @@ impl<T: Storage> Raft<T> {
...
+                return;
+            }
+        }
+    }
+
     /// Sends RPC, with entries to all peers that are not up-to-date
     /// according to the progress recorded in r.prs().
     pub fn bcast_append(&mut self) {
         let self_id = self.id;
+        let leader_group_id = self
```
BusyJay, line 1058: Can this be queried only when follower replication is enabled?

src/raft.rs
```diff
@@ -884,16 +994,132 @@ impl<T: Storage> Raft<T> {
...
+            return;
+        }
+
+        // If follower replication is enabled, MsgAppends sent to the same broadcast group
+        // will be merge into a MsgGroupBroadcast.
+        //
+        // Messages that needs to be forwarded are stored in hashmap temporarily,
+        // and they are grouped by broadcast_group_id of progress.
+        // Messages in msg_group will be pushed to message queue later.
+        let mut msg_group: HashMap<u64, Vec<Message>> = HashMap::default();
```
BusyJay, line 1081: Better cache msg_group to avoid allocation.

src/raft.rs
```diff
@@ -884,16 +994,132 @@ impl<T: Storage> Raft<T> {
...
+                            .entry(pr.broadcast_group_id)
+                            .or_default()
+                            .push(msg);
+                    }
+                }
+            });
+
+        // Merge messages in the same broadcast group and send them.
+        for (group_id, mut group) in msg_group.drain() {
+            // Double check: do not need to forward messages in leader's broadcast group.
```
BusyJay, line 1110: It's meaningless.

src/raft.rs
```diff
@@ -884,16 +994,132 @@ impl<T: Storage> Raft<T> {
...
+                    ..Default::default()
+                };
+                forwards.push(forward);
+            } else {
+                mark = idx;
+            }
+        }
+        // Attach forward information to MsgGroupbroadcast and send it.
+        group[mark].set_forwards(forwards.into());
+        for msg in group {
```
BusyJay, line 1046: You mean `self.msgs.push(group[mark])`?

src/raft.rs
```diff
@@ -884,16 +994,132 @@ impl<T: Storage> Raft<T> {
         self.r.send_append_aggressively(to, pr, &mut self.msgs)
     }
 
+    // Find an appropriate agent.
+    // If found an appropriate one, mark the corresponding message's type as
+    // MsgGroupBroadcast and return true. If not found, return false.
+    fn select_agent_for_bcast_group(&self, msgs: &mut [Message]) -> bool {
```
BusyJay, line 1000: Why not return the index of the message? Changing the type in this function is quite confusing.

src/raft.rs
```diff
@@ -884,16 +994,132 @@ impl<T: Storage> Raft<T> {
...
+        for msg in msgs {
+            let peer_id = msg.to;
+            let is_voter = self.prs().conf().voters().contains(peer_id);
+            // Agent must be voter and recently active.
+            if !is_voter || !self.is_recent_active(peer_id) {
```
BusyJay, line 1005: These information can be obtained when the message is generated.

src/raft.rs
LintianShi: It is a mis-operation. I will fix it.

src/raft.rs
LintianShi: I will remove it.

src/raft.rs
LintianShi: I think it is a better choice.

src/raft.rs
LintianShi: There are two conditions that we should use leader replication. One is that follower replication is disabled. The other is that follower replication is enabled but leader's group id is unknown. So I want to reuse the code of iterating progress and sending append.

BusyJay:
```
let mut use_leader_replication = self.is_leader();
let leader_group_id = 0;
if !use_leader_replication {
    leader_group_id = query;
    use_leader_replciation = leader_group_id == 0;
}
if use_leader_replication {
    return;
}   
```

src/raft.rs
LintianShi: The information of `voter` and `recent_active` is only used for agent selection. When the message is generated, we only determine whether it should be replicated by follower.
If we want to obtain these information when the message is generated, another cache is needed to record them.

BusyJay: No, what you need to do is merging `bcast_append`, `merge_msg_group` and `select_agent_for_bcast_group`.

src/raft.rs
```diff
@@ -888,12 +1002,106 @@ impl<T: Storage> Raft<T> {
...
+                // If found an appropriate agent, return the index of agent's message. Otherwise, return None.
+                agent_msg_idx = group
+                    .iter()
+                    .position(|(_, is_agent_candidate)| *is_agent_candidate);
+                need_merge = agent_msg_idx.is_some();
+            }
+            // Do not need to merge if group size is less than two. Or there is no appropriate agent.
+            if !need_merge {
+                msgs.append(&mut group.into_iter().map(|(msg, _)| msg).collect());
+                return;
```
Connor1996, line 1076: continue

src/raft.rs
```diff
@@ -888,12 +1002,106 @@ impl<T: Storage> Raft<T> {
...
+                        ..Default::default()
+                    };
+                    forwards.push(forward);
+                }
+            }
+            // Attach forward information to MsgGroupbroadcast and send it.
+            let mut agent_msg = group.swap_remove(agent_msg_idx.unwrap()).0;
+            agent_msg.set_msg_type(MessageType::MsgGroupBroadcast);
+            agent_msg.set_forwards(forwards.into());
+            msgs.push(agent_msg);
```
Connor1996, line 1097: no need to send other msgs for agent?

LintianShi: Other messages have been pushed to `self.msgs` when generated.

src/raft.rs
```diff
@@ -888,12 +1002,106 @@ impl<T: Storage> Raft<T> {
...
+            for (idx, (msg, _)) in group.iter().enumerate() {
+                // MsgAppend sent to other peers in this group only reserve basic forward information.
+                if idx != agent_msg_idx.unwrap() {
+                    let forward = Forward {
+                        to: msg.to,
+                        log_term: msg.log_term,
+                        index: msg.index,
+                        ..Default::default()
+                    };
+                    forwards.push(forward);
```
Connor1996, line 1090: what if too many forward info makes one message too large

LintianShi: Considering that there are often 3 or 5 replicas deployed on 2 or 3 az, I think one message won't take too many forward information.

src/raft.rs
LintianShi: Thank very much. I think this cause the jitters of the number of sent messages.

src/raft.rs
```diff
@@ -888,12 +1002,106 @@ impl<T: Storage> Raft<T> {
     /// according to the progress recorded in r.prs().
     pub fn bcast_append(&mut self) {
```
Connor1996, line 1003: follower replication only covers `bcast_append`, but misses somewhere calling `send_append` and `send_append_aggressively` directly

src/raft.rs
Connor1996: Oh, I was thought there would be multiple `MsgAppend` for one peer

src/raft.rs
LintianShi: When a cluster replicates data via `bcast_append`, the cluster is in a stable state with high probability.
If `send_append` is called, it is usually due to reject of `MsgAppendResp`, which means that follower's raft log is conflict with leader's. In this situation, we'd better use leader replication to fix follower's raft log as quickly as possible.

LintianShi: In previous test, we find that follower replication can reduce a lot of cross-az traffic. So I think we should sacrifice some coverage of follower replication to archive less performance regression. Though we just use follower replication in the main path of log replication, it can still reduce considerable cross-az traffic.
Maybe we can expand follower replication in the future. For example, we can get snapshot from followers.

src/raft.rs
```diff
@@ -260,6 +266,9 @@ pub struct RaftCore<T: Storage> {
 
     /// Max size per committed entries in a `Read`.
     pub(crate) max_committed_size_per_ready: u64,
+
+    // Message group cache for follower replication.
+    msg_group: HashMap<u64, Vec<(Message, bool)>>,
```
BusyJay, line 271: `Vec` is more efficient than `HashMap`.

src/tracker.rs
```diff
@@ -190,11 +190,11 @@ impl AckedIndexer for ProgressMap {
 /// which could be `Leader`, `Follower` and `Learner`.
 #[derive(Clone, Getters)]
 pub struct ProgressTracker {
-    progress: ProgressMap,
+    pub(crate) progress: ProgressMap,
 
     /// The current configuration state of the cluster.
     #[get = "pub"]
-    conf: Configuration,
+    pub(crate) conf: Configuration,
```
BusyJay, line 197: Doesn't the above line states it's public accessible already?

src/raft.rs
```diff
@@ -888,12 +1002,106 @@ impl<T: Storage> Raft<T> {
...
+                            .entry(pr.broadcast_group_id)
+                            .or_default()
+                            // The agent must be a voter and active recently.
+                            .push((msg, pr.recent_active && conf.voters().contains(peer_id)));
+                    }
+                }
+            });
+
+        // Merge messages in the same broadcast group and send them.
+        for (_, mut group) in core.msg_group.drain() {
```
BusyJay, line 1063: Can merge be done in L1056?

src/raft.rs
```diff
@@ -888,12 +1002,106 @@ impl<T: Storage> Raft<T> {
...
+        let prs = &mut self.prs.progress;
+        let conf = &self.prs.conf;
+        prs.iter_mut()
             .filter(|&(id, _)| *id != self_id)
-            .for_each(|(id, pr)| core.send_append(*id, pr, msgs));
+            .for_each(|(id, pr)| {
+                let mut tmp_msgs = Vec::default();
+                // Let messages be pushed into tmp_vec firstly.
+                core.send_append(*id, pr, &mut tmp_msgs);
+                for msg in tmp_msgs {
```
BusyJay, line 1043: How about
```
if pr.broadcast_group_id == leader_group_id || !pr.is_replicating() {
    msgs.extend(tmp_msgs);
    return;
}
let is_voter = is_voter(pr);
for msg in tmp_msgs {
    if msg.get_msg_type() != MsgAppend {
    } else {
    }
}
```

src/tracker.rs
LintianShi: Using `conf()`, there will be a conflict between mutable borrow of `self.prs.progress` and immutable borrow of `self.prs.conf`.

src/raft.rs
BusyJay: Not just query, but also update and memory footprint.

src/raft.rs
LintianShi: I think it might be impossible because the broadcast group is fixed after all progresses are iterated. The merge should start after the broadcast group is fixed.
If we group progress by `broadcast_group_id` first and then iterate them, it is feasible.

src/raft.rs
Connor1996: you should improve the test to cover this

src/raft.rs
LintianShi: Done. Use  `Vec<(u64, Vec<(Message, bool))>` instead of `HashMap<u64, Vec<(Message, bool)>>`.

src/raft.rs
LintianShi: Now async-fetch is supported.

PR #486, opened 2022-08-29
Interface of querying the information of progress

LintianShi:
Add `RawNode` some interfaces of querying the information of progress, in order to select appropriate agent for follower replication.

Signed-off-by: LintianShi <lintian.shi@pingcap.com>


LintianShi:
@Connor1996 PTAL

PR #493, opened 2022-11-08
Introduce enum priority

Connor1996:
Signed-off-by: Connor1996 <zbk602423539@gmail.com>

u64 based priority is not handy, when you want to lower or raise the priority of one peer, you have to adjust all the peers'.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,6 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::raft::Priority;
+
 pub use super::read_only::{ReadOnlyOption, ReadState};
 use super::util::NO_LIMIT;
 use super::{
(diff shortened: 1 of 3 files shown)


Connor1996:
Test would be fixed later if the change is acceptable.

BusyJay:
This is a breaking change and if users have adjusted the priority, this can lead to expected behavior during rolling update.

Connor1996:
Do you have any better idea? If the priority field in kvproto is i64, everything would work fine, but it's also breaking

BusyJay:
You can change the priority in `Config` to i64 and add a new priority field defined as s64 to pb definition. If new field is not set, then use the old field; otherwise use the new field. When broadcasting request vote, both fields are set if the priority is larger than 0.

This change is not a fully compatible change, but it reduce the impact that *only new priority is not recognized by the old nodes*.

PR #496, opened 2022-12-16
Update to Prost 0.11

agourlay:
This PR update to Prost 0.11.

Protobuf-build has been already updated https://github.com/tikv/protobuf-build/pull/63

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,10 @@ jobs:
             rust: nightly
     steps:
     - uses: actions/checkout@v2
+    - name: install Protoc
+      uses: arduino/setup-protoc@v1
+      with:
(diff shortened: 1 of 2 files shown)


BusyJay (approved):

BusyJay:
protoc is also missing

agourlay:
I need to find a way to install protoc across all OS configurations

BusyJay:
How about using package manager instead? I'm worried about the permissions of secrets.

On Windows, it can be installed by `choco install protoc`, on MacOS it can be done by `brew install protobuf`, on Linux, it can be done by `sudo apt-get install protobuf-compiler`.

agourlay:
@BusyJay No worries, I was able to use each package managers instead.

BusyJay (approved):
Thank you!

PR #499, opened 2023-01-11
Make request_snapshot more safer

ethercflow:
### Change

Modify `request_snapshot` API, remove parameter: `request_index: u64`, use `last_index` as `request_index`, when sending request snapshot. 

### Reason

When a follower uses this interface to actively request a snapshot from the leader, it is not safe to pass an arbitrary index, as the leader will continue to send AppendRPC before receiving the snapshot request, this gap log follower is difficult to handle. Eg: it may panic at https://github.com/tikv/raft-rs/blob/5ce52b480065ff31ecef16b9b77c7c3b7c57c8c7/src/raw_node.rs#L520-L529. So the safest way is to use `last_index` as `request_index`, and the follower rejects `appendRPC`.

close tikv/raft-rs#498

Signed-off-by: Wenbo Zhang <ethercflow@gmail.com>

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -2416,7 +2416,7 @@ impl<T: Storage> Raft<T> {
         } else if self.leader_id == INVALID_ID {
             info!(
                 self.logger,
-                "drop request snapshot because of no leader";
+                "no leader; dropping request snapshot";
                 "term" => self.term,
             );
         } else if self.snap().is_some() {
@@ -2439,7 +2439,7 @@ impl<T: Storage> Raft<T> {
             }
             info! {
                 self.logger,
-                "drop reqeust snapshot because of mismatch term";
+                "mismatched term; dropping request snapshot";
                 "term" => self.term,
                 "last_term" => request_index_term,
             };
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -2439,9 +2439,9 @@ impl<T: Storage> Raft<T> {
             }
             info! {
                 self.logger,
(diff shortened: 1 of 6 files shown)


ethercflow:
@Connor1996 @BusyJay PTAL, thanks!

Connor1996 (approved):
Please describe the reason

src/raft.rs
```diff
@@ -2837,6 +2842,7 @@ impl<T: Storage> Raft<T> {
         m.reject_hint = self.raft_log.last_index();
         m.to = self.leader_id;
         m.request_snapshot = self.pending_request_snapshot;
+        m.log_term = self.pending_request_snapshot_idx_term;
```
BusyJay, line 2845: `log_term` is supposed to be aligned with `reject_hint` when `reject` is true.

Connor1996 (dismissed):
LGTM

Connor1996 (approved):
LGTM

src/raft.rs
```diff
@@ -2407,7 +2407,7 @@ impl<T: Storage> Raft<T> {
     }
 
     /// Request a snapshot from a leader.
-    pub fn request_snapshot(&mut self, request_index: u64) -> Result<()> {
+    pub fn request_snapshot(&mut self) -> Result<()> {
```
BusyJay, line 2410: Should check if it has log of current term and test against it.

BusyJay (commented):
Rest LGTM

BusyJay (approved):
Rest LGTM

PR #513, opened 2023-07-03
feature: validate term value

### LINKED ISSUE #511: timing-based panic during prevote when using priority
**Describe the bug**
Nodes often crashing (`panic`) during first few moments of Raft, when using distinct `priority` values per node together with `pre_vote = true`.
The least-prioritized node never crashes.

When I looked into it, I found that it's because during the pre-voting, we reject pre-voting requests from lower-priority nodes:
https://github.com/tikv/raft-rs/blob/82d704cdc3d93258be1f45efd715b95930764d7f/src/raft.rs#L1467C65-L1467C98

We then try to send the rejection back to the proposing node, including our term as the message's term:
https://github.com/tikv/raft-rs/blob/82d704cdc3d93258be1f45efd715b95930764d7f/src/raft.rs#L1494C21-L1498C58

But during that send, term is checked to be != 0, resulting in a panic instead of sending the proposal rejection:
https://github.com/tikv/raft-rs/blob/82d704cdc3d93258be1f45efd715b95930764d7f/src/raft.rs#L617C9-L640C14

**Expected behavior**
PreVoting rejection and not `panic()`.

GuyLewin:
Fixes #511

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -380,6 +380,12 @@ impl<T: Storage> Raft<T> {
         if raft_state.hard_state != HardState::default() {
             r.load_state(&raft_state.hard_state);
         }
+        if r.term == 0 {
+            fatal!(
+                r.logger,
+                "Invalid term value: 0"
+            );
+        }
         if c.applied > 0 {
             r.commit_apply(c.applied);
         }
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -419,6 +419,8 @@ impl MemStorage {
         // In practice, we choose the second way by assigning non-zero index to first index. Here
         // we choose the first way for historical reason and easier to write tests.
         core.raft_state.conf_state = ConfState::from(conf_state);
+        // Initialize term with a valid value (anything except 0)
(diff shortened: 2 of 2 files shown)


src/raft.rs
```diff
@@ -380,6 +380,9 @@ impl<T: Storage> Raft<T> {
         if raft_state.hard_state != HardState::default() {
             r.load_state(&raft_state.hard_state);
         }
+        if r.term == 0 {
```
BusyJay, line 383: Should be `!r.prs.is_empty() && r.term == 0`.

GuyLewin:
@BusyJay do you agree with my change in `src/storage.rs`?:
https://github.com/tikv/raft-rs/blob/b0a374060334a35442e7da1f3769599de7a040ed/src/storage.rs#L422-L423 ?

If so - it breaks a lot of tests but I can go over them and fix them. I just don't want to do it if it's the incorrect fix since it'll require some manual work.

BusyJay:
MemStorage is supposed to be used for tests, change it or not is OK to me.

PR #520, opened 2023-08-02
chore: readability for take read state

tisonkun:
It's more clear than `mem::swap` cause it doesn't depend on that `rd.read_state` happens to be empty.

Semantically we take the read state from node to ready, not swap.


gengliqi (approved):

tisonkun:
trivial to merge

Thanks for your review!

PR #522, opened 2023-08-17
Suggestion: Add `CustomDeserializer` trait for making debugging `raft-rs` logs much easier

jopemachine:
As I mentioned in the raft-rs Slack channel before, I think it would be very useful if users could see their byte slice information like Entry's `context`, `data`.

I think it could be possible by exposing the `CustomDeserializer` trait in `raft-rs`.

In this PR, I implement this thinking roughly, and I can see they are working as expected through [ratify](https://github.com/lablup/raftify) which uses `raft-rs` under the hood.

You can see the logged `Message` includes `context`, `data` as deserialized strings, not as byte slices in the below screenshot. 

I'm just wondering if there are some `raft-rs` maintainers who agree with this idea and if not, why not.

Although this PR is currently just some sample codes to show ideas and working code, if there's a chance this PR will be merged, I'm willing to work on making improvements.

(Otherwise, I think I have to keep the cloned separate repository for `raftify` which would be troublesome for me...)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,13 +19,13 @@ members = ["proto", "harness", "datadriven"]
 default = ["protobuf-codec", "default-logger"]
 # Enable failpoints
 failpoints = ["fail/failpoints"]
-protobuf-codec = ["raft-proto/protobuf-codec", "bytes"]
+protobuf-codec = ["raft-proto/protobuf-codec"]
 prost-codec = ["raft-proto/prost-codec"]
 default-logger = ["slog-stdlog", "slog-envlogger", "slog-term"]
 
 # Make sure to synchronize updates with Harness.
 [dependencies]
-bytes = { version = "1", optional = true }
+bytes = "1.0"
 fxhash = "0.2.1"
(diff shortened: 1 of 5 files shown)


BusyJay:
Can you describe more details on what's the behavior without this patch and why it's bad?

jopemachine:
> Can you describe more details on what's the behavior without this patch and why it's bad?

I think nobody wants to see byte slice in their raft-rs logs like the below screenshot.

스크린샷 2023-08-17 오후 11 23 45

Instead, I think user must have to pass their own deserialization method like below screenshot.

스크린샷 2023-08-17 오후 11 25 51

And I think this PR could accomplish this by exposing the `CustomDeserializer` trait.

If user don't implement that trace, the byte slices will still be printed as byte slices.

But if user want to see deserialized string in their raft-rs logs, they can make it through `CustomDeserializer`.

jopemachine:
So, the behavior without the patch is showing raw byte slice in logs, and I think it could be very bad for debugging and figuring out what's happening inside...

PR #537, opened 2024-03-11
allow raft apply committed logs before they are persisted

glorv:
ref: tikv/tikv#16457

Design doc: https://github.com/tikv/rfcs/pull/112

As `committed` means more than quorum node are persisted, which no data loss even if all other node are down. So in this situation, it is safe to apply this log even if it is still not persisted. 

This PR introduces a new config `applied_unpersisted_log_limit` that allows return unpersisted raft log in `light_ready`. This is one step to optimize the tail latency that one slow node can significantly impact the overall latency. By setting a proper value for `applied_unpersisted_log_limit` we can avoid the in memory raft entries consumes too much memory which may lead to OOM.

After this change, if `applied_unpersisted_log_limit` is > 0, then it is possible: 1) applied > persisted. 2) applied > committed.(Only can happen at restart). So we loose some check since they are not always true anymore.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -101,7 +101,7 @@ pub struct Config {
 
     /// Maximum raft log number that can be applied after commit but before persist.
     /// The default value is 0, which means apply after both commit and persist.
-    pub applied_unpersisted_log_limit: u64,
+    pub apply_unpersisted_log_limit: u64,
 }
 
 impl Default for Config {
@@ -124,7 +124,7 @@ impl Default for Config {
             priority: 0,
             max_uncommitted_size: NO_LIMIT,
             max_committed_size_per_ready: NO_LIMIT,
-            applied_unpersisted_log_limit: 0,
+            apply_unpersisted_log_limit: 0,
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -98,6 +98,10 @@ pub struct Config {
 
     /// Max size for committed entries in a `Ready`.
     pub max_committed_size_per_ready: u64,
+
+    /// Maximum raft log number that can be applied after commit but before persist.
+    /// The default value is 0, which means apply after both commit and persist.
+    pub applied_unpersisted_log_limit: u64,
 }
 
 impl Default for Config {
@@ -120,6 +124,7 @@ impl Default for Config {
             priority: 0,
             max_uncommitted_size: NO_LIMIT,
             max_committed_size_per_ready: NO_LIMIT,
+            applied_unpersisted_log_limit: 0,
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -101,7 +101,7 @@ pub struct Config {
 
     /// Maximum raft log number that can be applied after commit but before persist.
     /// The default value is 0, which means apply after both commit and persist.
-    pub apply_unpersisted_log_limit: u64,
+    pub max_apply_unpersisted_log_limit: u64,
 }
 
 impl Default for Config {
@@ -124,7 +124,7 @@ impl Default for Config {
             priority: 0,
             max_uncommitted_size: NO_LIMIT,
             max_committed_size_per_ready: NO_LIMIT,
-            apply_unpersisted_log_limit: 0,
+            max_apply_unpersisted_log_limit: 0,
         }
     }
 }
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -382,7 +382,9 @@ impl<T: Storage> Raft<T> {
             r.load_state(&raft_state.hard_state);
         }
         if c.applied > 0 {
-            r.commit_apply(c.applied);
+            // at initialize, it is possible that applied_index > committed_index,
+            // so we should skip this check at `commit_apply`.
+            r.commit_apply_internal(c.applied, true);
         }
         r.become_follower(r.term, INVALID_ID);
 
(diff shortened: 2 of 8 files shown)


glorv:
@Connor1996 @gengliqi  @overvenus @tonyxuqi @BusyJay  PTAL, thank you~

cfzjywxk:
Perhaps we could organize and link a top-down design document or RFC to illustrate the specific changes in tikv/raftstore, interface impacts, etc., after removing the implicit code constraints related to apply entry, as well as the corresponding design in raft-rs.

I remember @gengliqi  has some relevant design documents before?

src/raft.rs
```diff
@@ -366,6 +367,7 @@ impl<T: Storage> Raft<T> {
                 max_committed_size_per_ready: c.max_committed_size_per_ready,
             },
         };
+        r.raft_log.apply_unpersisted_log_limit = c.apply_unpersisted_log_limit;
```
Connor1996, line 370: pass it through `RaftLog::new`?

src/raft_log.rs
```diff
@@ -1185,6 +1202,91 @@ mod test {
                 );
             }
         }
+
+        let ents = [
```
Connor1996, line 1206: Do we need a case that applied > last index after restart, and make sure the init process fails?

glorv: It is possible applied > last_index and the it won't let init process fail. Call `next_entries` will just return None. Test case added.

tonyxuqqi:
@glorv We should have a RFC document to summarize the design.

src/config.rs
```diff
@@ -98,6 +98,10 @@ pub struct Config {
 
     /// Max size for committed entries in a `Ready`.
     pub max_committed_size_per_ready: u64,
+
+    /// Maximum raft log number that can be applied after commit but before persist.
+    /// The default value is 0, which means apply after both commit and persist.
+    pub apply_unpersisted_log_limit: u64,
```
tonyxuqqi, line 104: Will it support dynamic config update?

glorv: Yes, we provide a public function `Raft::set_apply_unpersisted_log_limit` to change this value in realtime.

src/raft_log.rs
```diff
@@ -310,7 +316,11 @@ impl<T: Storage> RaftLog<T> {
         if idx == 0 {
             return;
         }
-        if idx > cmp::min(self.committed, self.persisted) || idx < self.applied {
+        // Do not check idx with committed or persisted index here becase when `apply_unpersisted_log_limit` > 0:
+        // 1. then it is possible idx > persisted.
+        // 2. when the application restart after applied but before committed entried(and committed index) is persisted
+        //    then it is also possible idx > committed.
+        if idx < self.applied {
```
gengliqi, line 323: At least `idx` should be greater than or equal to `self.committed`.

src/raft_log.rs
```diff
@@ -48,14 +49,18 @@ pub struct RaftLog<T: Storage> {
...
     ///
-    /// Invariant: persisted < unstable.offset && applied <= persisted
+    /// Invariant: persisted < unstable.offset
     pub persisted: u64,
 
     /// The highest log position that the application has been instructed
     /// to apply to its state machine.
     ///
-    /// Invariant: applied <= min(committed, persisted)
+    /// Invariant: applied <= min(committed, persisted + `apply_unpersisted_log_limit`)
```
gengliqi, line 58: `apply_unpersisted_log_limit` is a config that can be changed so it does not belong to `Invariant`.

src/raft_log.rs
glorv: This is for the initialization phase, currently it call applied_to to set the applied_index. I will change it to keep this check for normal path.

src/raft_log.rs
glorv: I change to code to let initialization not call `applied_to` so we can keep the check.

glorv:
@Connor1996 @gengliqi  PTAL again, thanks~

Connor1996 (approved):
LGTM

src/raft_log.rs
```diff
@@ -42,20 +43,26 @@ pub struct RaftLog<T: Storage> {
     /// on a quorum of nodes.
     ///
     /// Invariant: applied <= committed
+    /// NOTE: this invariant can be break after restart if `max_apply_unpersisted_log_limit` is 0,
```
gengliqi, line 46: is greater than 0?

src/raft_log.rs
```diff
@@ -320,6 +328,11 @@ impl<T: Storage> RaftLog<T> {
                 self.persisted,
             )
         }
+        self.applied_to_unchecked(idx);
+    }
+
+    #[inline]
+    pub(crate) fn applied_to_unchecked(&mut self, idx: u64) {
```
gengliqi, line 335: Seems `pub` is enough?

glorv: This is an internal function only used for initialization, we don't want it be called by end user.

src/raft_log.rs
glorv: fixed

src/raft_log.rs
```diff
@@ -42,20 +43,26 @@ pub struct RaftLog<T: Storage> {
...
 
     /// The highest log position that the application has been instructed
     /// to apply to its state machine.
     ///
-    /// Invariant: applied <= min(committed, persisted)
+    /// Invariant: applied <= min(committed, persisted) iff `max_apply_unpersisted_log_limit` is 0.
     pub applied: u64,
+
+    /// the maximum log gap between persisted_index and applied_index.
+    /// Caller should ensure the value won't lead to the upper bound overflow.
```
gengliqi, line 64: The comment about overflow may be unnecessary. Alternatively, it can be written in `Config`.
```suggestion
    /// The maximum log gap between persisted and applied.
```

gengliqi:
I think we need to add more restrictions to these logs that can be applied before persisting so that correctness errors can be found. 
A log can be applied before persisting only when a log with the same term has been committed and persisted.
After restarting, we can check if the logs between committed and applied have the same term as the last committed entry. If not the same, there must be something wrong and the panic must happen immediately.

gengliqi (approved):

gengliqi:
I will propose another pr to implement this restriction.

PR #539, opened 2024-03-21
Provide formal specification and verification using TLA+ for Raft

wego1236:
We've made a formal specification for Raft in paper "In Search of an Understandable Consensus Algorithm" and a certain scale of model checking to verify the correctness of raft-rs(currently still running). There may exist differences between our specification and Raft in paper or in implementation. Please let us know flaws of specification if you have any question.

### DIFF
--- a/tla/.gitignore
+++ b/tla/.gitignore
@@ -1,0 +1,2 @@
+states
+*.jar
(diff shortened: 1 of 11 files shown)


Connor1996 (commented):
Thanks for your great job. 

We have a dedicated repo for TLA+ spec, and there is already one spec for Raft https://github.com/pingcap/tla-plus/blob/master/Raft/Raft.tla, though just for Raft consensus algorithm itself. 

I think your work would be a great complement for that. How about moving this to that repo and adding a link in raft-rs?

wego1236:
Okay, I'll go now.

PR #542, opened 2024-03-28
Rollback impl AssertSend

tisonkun:


src/raft_log.rs
```diff
@@ -58,12 +57,10 @@ pub struct RaftLog<T: Storage> {
     pub applied: u64,
 }
 
-impl<T> ToString for RaftLog<T>
```
BusyJay, line 61: This is a breaking change. Better to implement both.

tisonkun: I guess no?

```rust
#[cfg(not(no_global_oom_handling))]
#[stable(feature = "rust1", since = "1.0.0")]
impl<T: fmt::Display + ?Sized> ToString for T {
    // A common guideline is to not inline generic functions. However,
    // removing `#[inline]` from this method causes non-negligible regressions.
    // See <https://github.com/rust-lang/rust/pull/74852>, the last attempt
    // to try to remove it.
    #[inline]
    default fn to_string(&self) -> String {
        let mut buf = String::new();
        let mut formatter = core::fmt::Formatter::new(&mut buf);
        // Bypass format_args!() to avoid write_str with zero-length strs
        fmt::Display::fmt(self, &mut formatter)
            .expect("a Display implementation returned an error unexpectedly");
        buf
    }
}
```

Or I can disable the clippy rule here for now.

BusyJay: I see.

BusyJay (approved):

tisonkun:
Merging ...

PR #543, opened 2024-04-01
fix applied upper bound

glorv:
#537 introduce the new config `max_applied_unpersisted_log_limit` that changes applied upper bound to `min(committed, persisted + max_applied_unpersisted_log_limit)`, but since `max_applied_unpersisted_log_limit` can be changed dynamically, we can't directly use it as the upper bound of `applied_to`, so this PR loose the upper bound to `committed`.

### DIFF
--- a/src/raft_log.rs
+++ b/src/raft_log.rs
@@ -322,14 +322,15 @@ impl<T: Storage> RaftLog<T> {
         if idx == 0 {
             return;
         }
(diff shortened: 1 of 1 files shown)


glorv:
@gengliqi  @Connor1996  PTAL

Connor1996 (approved):
LGTM

gengliqi (approved):

PR #546, opened 2024-05-11
include leader and current rule info in vote log

glorv:
This can help us determine a peer's current state easier when investigate vote and leader changes related issues.


glorv:
waiting for #547 to fix the nightly build first.

BusyJay (approved):

PR #557, opened 2024-11-14
raft: next index shall be larger than match index

wego1236:
Leader shall always replicate its log from a index larger than corresponding match index. Note that this is not a correctness issue, but rather an optimization in case of message reordering.

Related: https://github.com/tikv/raft-rs/issues/555


BusyJay:
Thanks, seems the test case is missing.

wego1236:
Do I need to add test here?

wego1236:
This test case needs to be added to harness/tests/integreation_cases/test_raft.rs right?

BusyJay:
Yes.

wego1236:
I may need some help, the test case I wrote modelled after etcd doesn't achieve the same results as they do, this test case may need your help to complete it. When I tried it, I found that in some cases etcd and raft-rs produced different results, so I may not be able to simply copy the corresponding test cases.

wego1236:
I'm not familiar with rust, maybe for me this test case needs to know a bit too much, if it's easier for you, could you help me to modify it a bit

harness/tests/integration_cases/test_raft.rs
```diff
@@ -5851,3 +5851,57 @@ fn test_switching_check_quorum() {
...
+    assert_eq!(resp2.index, 2);
+
+    // r2 handles the first MsgApp and responses to r1.
+    // And r1 updates match index accordingly.
+    let _ = r2.step(req1);
+    let m = expect_one_message(&mut r2);
+    assert_eq!(m.reject, false);
+    assert_eq!(m.index, 2);
+    let _ = r1.step(m);
+    let _ = expect_one_message(&mut r1);
```
chagelo, line 5895:
```suggestion

```
You may need to remove this line.

wego1236: thanks a lot, i'll fix it now

wego1236:
Is there anything else I need to do?

BusyJay (approved):
LGTM

BusyJay:
Thank you!

PR #560, opened 2025-01-17
change progress status to replicate when handling heartbeat response

LuDiting:
In the current implementation, if there is a lagging follower node, the corresponding progress state in the Leader node is Probe. This state will not be released in handle_append_response until next new write.

However, there is a situation where when there is an unreachable node and the upper-level application calls report_unreachable, this node will be marked as Probe by the Leader. When this node rejoins the cluster, it is still in Probe state. If the cluster does not have any new writes during this period, this node is actually Replicate. But its Probe state will not be updated until next new write.

This has no effect on consistency, but some operations based on the progress state of the node may be involved in some upper-level applications. Therefore, it is believed that an operation can be added in handle_heartbeat_response to set the progress of the node that keeps up with the progress to replicate state.


BusyJay:
Make sense. I see upstream has already fix this issue in https://github.com/etcd-io/raft/pull/52, how about just porting the patch instead?

PR #561, opened 2025-02-20
reset max_apply_unpersisted_log_limit in become_follower

glorv:
This PR reset the `max_apply_unpersisted_log_limit` value when peer role demote from leader to follower. Because the user can only notice role change event after call `ready()` and the committed entries are collected in the `ready()` so we have to do this here to ensure we only enable `max_apply_unpersisted_log_limit` on raft leader. 

NOTE: In theory, the affect of `max_apply_unpersisted_log_limit` is not related the raft role. We limit it to only leader to prevent potencial corn cases. We may want to remove this restriction in the future. In that case, we should revert this PR then.

See https://github.com/tikv/tikv/issues/17868 for more details.

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -1882,6 +1882,9 @@ fn test_leader_stepdown_when_quorum_lost() {
 
     assert_eq!(sm.state, StateRole::Follower);
     // check after become follower, the limit is reset.
+    // Currently we only support return committed but not persisted raft
+    // entries on raft leader. Due to the limitation of current interface,
+    // we hard code this logic in raft-rs.
     assert_eq!(sm.raft_log.max_apply_unpersisted_log_limit, 0);
 }
 
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -1874,12 +1874,15 @@ fn test_leader_stepdown_when_quorum_lost() {
 
     sm.become_candidate();
     sm.become_leader();
+    sm.raft_log.max_apply_unpersisted_log_limit = 100;
 
     for _ in 0..=sm.election_timeout() {
         sm.tick();
     }
 
     assert_eq!(sm.state, StateRole::Follower);
+    // check after become follower, the limit is reset.
+    assert_eq!(sm.raft_log.max_apply_unpersisted_log_limit, 0);
 }
 
 #[test]
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -824,7 +824,7 @@ impl<T: Storage> RaftCore<T> {
                     aggressively: !allow_empty,
                 }),
             );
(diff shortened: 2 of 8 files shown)


glorv:
@Connor1996 @overvenus @gengliqi  PTAL, thanks~

Connor1996 (commented):
Can we add a test to cover the case?

glorv:
@Connor1996 @hhwyt PTAL again

Connor1996 (approved):
LGTM

glorv:
@overvenus PTAL

src/raft.rs
```diff
@@ -1152,6 +1152,13 @@ impl<T: Storage> Raft<T> {
         let from_role = self.state;
         self.state = StateRole::Follower;
         self.pending_request_snapshot = pending_request_snapshot;
+        // TODO: In theory, it's better to let the user control this after
+        // leadership changes, but bacuase the user only know this info after
+        // call `ready()` but the committed entries is fetched within `ready()`,
```
hbisheng, line 1157:
```suggestion
        // leadership changes, but because the user only know this info after
        // calling `ready()` but the committed entries is fetched within `ready()`,
```

hbisheng:
Would you mind putting down some context in the PR description? Otherwise it might not be obvious to future readers that this change is part of https://github.com/tikv/tikv/pull/18241.

harness/tests/integration_cases/test_raft.rs
```diff
@@ -1874,12 +1874,15 @@ fn test_leader_stepdown_when_quorum_lost() {
...
     sm.become_candidate();
     sm.become_leader();
+    sm.raft_log.max_apply_unpersisted_log_limit = 100;
 
     for _ in 0..=sm.election_timeout() {
         sm.tick();
     }
 
     assert_eq!(sm.state, StateRole::Follower);
+    // check after become follower, the limit is reset.
```
overvenus, line 1884: Please add some comments explaining why the reset is necessary.

glorv: Done. PTAL again, thanks

src/raft.rs
```diff
@@ -1152,6 +1152,13 @@ impl<T: Storage> Raft<T> {
         let from_role = self.state;
         self.state = StateRole::Follower;
         self.pending_request_snapshot = pending_request_snapshot;
+        // TODO: In theory, it's better to let the user control this after
+        // leadership changes, but because the user only know this info after
+        // calling `ready()` but the committed entries is fetched within `ready()`,
+        // so we hard code this logic here currently. We may need to remove
+        // this hard code when we want to also support apply unpersisted log
+        // on follower.
+        self.raft_log.max_apply_unpersisted_log_limit = 0;
```
overvenus, line 1161: Please update docs of this behavior change.

overvenus (commented):
Rest LGTM

PR #566, opened 2025-08-22
extend commit group to learners

lance6716:
extend commit group to learners. When make sure the group commit rule "raft log should be replicated to at least 2 group", also consider learners.

This feature is used to implement replication across DC by raft learner. Different with existing commit group which check followers, learner has less impact on leader&follower group like a TiFlash replica. It won't affect votes and won't elect itself as leader. Adding commit group for learner can switch between RPO=0 and better performance replication.

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5165,6 +5165,252 @@ fn test_group_commit() {
 }
 
 #[test]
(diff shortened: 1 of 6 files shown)


BusyJay:
I don't see why this feature matters. The point of group commit is to ensure latest changes are replicated to different AZs, so if one AZ is down, the new leader can still have all the changes. So what't the point of considering learner here?

PR #568, opened 2025-10-22
Add readability to example

nibblerenush:
1. We get rid of the magic number `1` and thus show that the node number must be the same in `MemStorage::new_with_conf_state` and `let cfg = Config { id: ...`
2. Adding an call `cfg.validate().unwrap()` as it is done in the documentation https://docs.rs/raft/latest/raft/


nibblerenush:
@BusyJay PTAL

PR #582, opened 2026-03-18
support updating raft-max-inflight-msgs setting

glorv:
The current impl only provide function `Raft::adjust_max_inflight_msgs` that changes the inflight cap of a specific peer. So even when we set all the peers' cap with this function, any incoming peer's cap will still be set with the initial inflight cap. This is not what we want when we want to update the global setting which should update the max_inflight_msgs for all existing and incoming peers.

So this PR add a new function `set_raft_max_inflight_msgs` to support updating the max_inflight setting.

### DIFF
--- a/src/raft.rs
+++ b/src/raft.rs
@@ -2963,4 +2963,50 @@ impl<T: Storage> Raft<T> {
             pr.ins.set_cap(cap);
         }
     }
+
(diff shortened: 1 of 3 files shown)


Connor1996 (approved):
LGTM

PR #584, opened 2026-03-25
fix: step down when leader is removed from config

Qian-Cheng-nju:
Hi, thanks for raft-rs!

I noticed an unimplemented code path in post_conf_change() flagged by TODO comments and put together a fix.

When a leader is removed from the voter set, post_conf_change() sets promotable=false but returns without stepping down. The removed leader continues sending heartbeats, which suppress elections on the remaining voters. This blocks the cluster from electing a new leader until the old one is shut down or partitioned.

The fix makes the removed leader transition to follower immediately. The included test (test_bug_removed_leader_does_not_step_down) demonstrates the issue with a 3-node cluster.

## Summary by CodeRabbit

* **Bug Fixes**
  * Leaders removed from the voter set now immediately step down to follower, stopping heartbeat emission and preventing stale leadership behavior.

* **Tests**
  * New integration test verifies that a removed leader transitions to follower, becomes non-promotable, and ceases sending heartbeat messages after the removal.

### DIFF
--- a/harness/tests/integration_cases/test_raft.rs
+++ b/harness/tests/integration_cases/test_raft.rs
@@ -5908,3 +5908,64 @@ fn test_log_replication_with_reordered_message() {
     // r1 shall re-send MsgApp from match index even if resp2's reject hint is less than matching index.
     assert_eq!(r1.prs().get(2).unwrap().matched, m.index)
 }
+
+// Test that a leader removed from the voter set steps down immediately.
+// Previously, post_conf_change() set promotable=false but returned early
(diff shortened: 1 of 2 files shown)


BusyJay:
Thanks, how about porting https://github.com/etcd-io/raft/pull/79 so that we will keep sync with upstream?

Qian-Cheng-nju:
Thanks for the pointer! I've updated this PR to port etcd/raft#79 — the step-down behavior is now opt-in via a step_down_on_removal config option(default false) for backwards compatibility. Tests cover both the enabled and default cases.

PR #589, opened 2026-05-11
try to shrink unstable entries buffer to avoid memory leak

glorv:
The issue is https://github.com/tikv/tikv/issues/19593 . The unstable entries can consume too much memory (in this case > 20GB) when there are about ~250k raft groups. So this PR try to shrink the entry buffer to avoid memory leak.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced memory usage by opportunistically shrinking retained capacity for log entry buffers after clears, truncations, or restores to avoid holding large unused allocations.

* **New Features**
  * Added a public API to aggressively release internal entry buffers when they are empty.

* **Tests**
  * Added test coverage validating the memory-reduction behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/raft-rs/pull/589)

### DIFF
--- a/src/log_unstable.rs
+++ b/src/log_unstable.rs
@@ -22,7 +22,8 @@ use slog::Logger;
 
 const SHRINK_KEEP_CAPACITY: usize = 64;
 const SHRINK_EMPTY_CAPACITY_THRESHOLD: usize = 256;
-const SHRINK_RATIO: usize = 4;
+const SHRINK_RATIO: usize = 2;
+const SHRINK_RESERVED_SIZE_RATIO: usize = 2;
 
 /// Unstable contains "unstable" log entries and snapshot state that has
 /// not yet been written to Storage.
@@ -97,6 +98,19 @@ impl Unstable {
         }
     }
 
+    /// Release the backing buffer of `entries` to further reduce memory usage.
+    ///
+    /// This is only suitable when the raft group is idle and unlikely to
+    /// propose new entries soon. Frequent release and reallocation of the
+    /// buffer can hurt performance.
+    pub fn release_entry_buffer(&mut self) {
+        if self.entries.is_empty() && self.entries.capacity() > 0 {
+            self.entries = vec![];
+        }
+    }
+
+    /// Try to shrink the capacity of `entries` when the entry count drops
+    /// to reduce memory usage.
     fn maybe_shrink_entries(&mut self) {
         let len = self.entries.len();
         let cap = self.entries.capacity();
@@ -104,7 +118,7 @@ impl Unstable {
             return;
         }
 
-        let target = len.max(SHRINK_KEEP_CAPACITY);
+        let target = (len * SHRINK_RESERVED_SIZE_RATIO).max(SHRINK_KEEP_CAPACITY);
         if cap >= target * SHRINK_RATIO {
             self.entries.shrink_to(target);
         }
@@ -239,6 +253,7 @@ mod test {
     use crate::eraftpb::{Entry, Snapshot, SnapshotMetadata};
     use crate::log_unstable::{
         Unstable, SHRINK_EMPTY_CAPACITY_THRESHOLD, SHRINK_KEEP_CAPACITY, SHRINK_RATIO,
+        SHRINK_RESERVED_SIZE_RATIO,
     };
     use crate::util::entry_approximate_size;
 
@@ -528,10 +543,14 @@ mod test {
             .map(|i| new_entry(i as u64 + 5, 1))
             .collect::<Vec<_>>();
         u.entries_size = u.entries.iter().map(entry_approximate_size).sum::<usize>();
-        let half = u.entries.len() / SHRINK_RATIO;
-        u.entries.truncate(half);
+        let shrunk_len = u.entries.len() / (SHRINK_RATIO * SHRINK_RESERVED_SIZE_RATIO);
+        u.entries.truncate(shrunk_len);
         u.entries_size = u.entries.iter().map(entry_approximate_size).sum::<usize>();
         u.maybe_shrink_entries();
-        assert!(u.entries.capacity() <= half.max(SHRINK_KEEP_CAPACITY) * SHRINK_RATIO);
+        assert!(
+            u.entries.capacity()
+                <= (shrunk_len * SHRINK_RESERVED_SIZE_RATIO).max(SHRINK_KEEP_CAPACITY)
+                    * SHRINK_RATIO
+        );
     }
 }
--- a/src/log_unstable.rs
+++ b/src/log_unstable.rs
@@ -20,6 +20,10 @@ use crate::eraftpb::{Entry, Snapshot};
 use crate::util::entry_approximate_size;
 use slog::Logger;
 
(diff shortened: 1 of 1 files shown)


glorv:
@BusyJay @overvenus @gengliqi @Connor1996 PTAL

LykxSassinator (approved):
LGTM.

And the review comment from Codex can be reviewed from the following comment:

### Deep Review

#### Problem Summary
- This branch tries to reduce retained heap usage in `Unstable.entries` by shrinking the backing `Vec<Entry>` after large clears or truncations. The production change is in `src/log_unstable.rs`, and the stated intent comes from commit `7a5e599` ("try to shrink unstable entries buffer to avoid memory leak").

#### Solution Walkthrough
- `src/log_unstable.rs:23` adds three heuristics:
  - `SHRINK_KEEP_CAPACITY = 64`
  - `SHRINK_EMPTY_CAPACITY_THRESHOLD = 256`
  - `SHRINK_RATIO = 4`
- `src/log_unstable.rs:100` adds `maybe_shrink_entries()`. It inspects `entries.len()` and `entries.capacity()`, skips small buffers (`cap <= 256`), and otherwise calls `Vec::shrink_to(target)` when the spare capacity is at least 4x the post-truncate target.
- `src/log_unstable.rs:129` calls the helper after `stable_entries()` clears the entire unstable buffer. That means a large fully-persisted unstable log now keeps roughly 64 slots instead of the previous peak capacity.
- `src/log_unstable.rs:166` calls the helper in `restore()`, so snapshot restore also drops a previously oversized unstable-entry buffer.
- `src/log_unstable.rs:186` and `src/log_unstable.rs:196` call the helper in both overwrite paths of `truncate_and_append()`:
  - full replacement (`after <= self.offset`)
  - partial truncation of the suffix (`after > self.offset`)
- `src/log_unstable.rs:507` adds `test_maybe_shrink_entries()` to exercise the new behavior.

#### Findings (ordered by severity)
- Medium: `test_maybe_shrink_entries()` asserts guarantees that `Vec::shrink_to()` does not make. `src/log_unstable.rs:525` requires `capacity() == SHRINK_KEEP_CAPACITY`, and `src/log_unstable.rs:535` requires the post-shrink capacity to stay under `target * SHRINK_RATIO`. Rust only guarantees that `shrink_to(min)` leaves capacity at least `max(len, min)`; allocators are free to keep more spare capacity. This makes the new test allocator-dependent and capable of failing even when the implementation is correct.
- Medium: the new shrink call is placed before the immediate re-append in the hot `truncate_and_append()` rewrite path (`src/log_unstable.rs:186`, `src/log_unstable.rs:188`, `src/log_unstable.rs:196`, `src/log_unstable.rs:199`). For large conflict rewrites, the old code could reuse the existing buffer; the new code may first reallocate downward and then grow again on `extend_from_slice`, adding an extra allocation and copy exactly on the follower append path. I do not see a benchmark or measurement in this branch showing that the retained-capacity savings outweigh that new allocator churn.

#### Costs and Negative Impacts
- Correctness: The core log-state transitions are unchanged; the main correctness risk I found is in the new test's incorrect assumptions about `Vec` capacity semantics.
- Security: None apparent.
- Compatibility: No wire or API compatibility change is introduced.
- Robustness: The production heuristic depends on allocator behavior for how much memory is actually returned, so the change improves memory retention opportunistically rather than deterministically.
- Cognitive Load: `src/log_unstable.rs` now carries three tuning constants and a shrink heuristic that needs justification and maintenance.
- CPU: Possible extra allocation/copy work on `truncate_and_append()` rewrite paths because shrink happens before the new entries are appended.
- Memory: Retained `Vec<Entry>` capacity should drop after large clears/truncations, but this only affects the outer `Vec` buffer. It does not change the lifetime of any still-retained entries.
- Log Volume: None.

#### Engineering Rules Check
- No `AGENTS.md` file was present in the repository root, so there were no repo-local engineering rules to validate against.

#### Questions and Assumptions
- I used `origin/master...HEAD` as the review base because `origin/HEAD` resolves to `origin/master` in this repository.
- I assumed the branch intent is to reduce retained `Vec<Entry>` capacity, not to fix a true leak of entry payload buffers.
- Local validation was partially blocked by the toolchain environment: `cargo test --lib test_maybe_shrink_entries --offline -- --exact` failed while compiling `proc-macro2 v1.0.51` with `error[E0635]: unknown feature 'proc_macro_span_shrink'` under `rustc 1.76.0-nightly (06e02d5b2 2023-12-09)`.

#### Suggested Tests / Validation
- Relax `src/log_unstable.rs:525` and `src/log_unstable.rs:535` to assert behavior that matches the `Vec::shrink_to()` contract, not exact capacity values.
- Add a targeted benchmark or allocation measurement for the two `truncate_and_append()` rewrite paths to justify the shrink-before-append tradeoff.
- Add explicit shrink-behavior coverage for `restore()` and the `after <= self.offset` branch of `truncate_and_append()`, since those are production call sites of the new helper.

coocood (approved):

hhwyt (approved):

Connor1996 (approved):
LGTM

src/log_unstable.rs
```diff
@@ -93,6 +97,19 @@ impl Unstable {
...
+    fn maybe_shrink_entries(&mut self) {
+        let len = self.entries.len();
+        let cap = self.entries.capacity();
+        if cap <= SHRINK_EMPTY_CAPACITY_THRESHOLD {
+            return;
+        }
+
+        let target = len.max(SHRINK_KEEP_CAPACITY);
+        if cap >= target * SHRINK_RATIO {
+            self.entries.shrink_to(target);
```
BusyJay, line 123: You can shrink to `target * 2` in this case. So further entries won't trigger reallocation immediately.

BusyJay (commented):
Any benchmark result?

src/log_unstable.rs
```diff
@@ -20,6 +20,10 @@ use crate::eraftpb::{Entry, Snapshot};
 use crate::util::entry_approximate_size;
 use slog::Logger;
 
+const SHRINK_KEEP_CAPACITY: usize = 64;
+const SHRINK_EMPTY_CAPACITY_THRESHOLD: usize = 256;
```
BusyJay, line 24: So the mem size of an entry is at least 64 byte, with this threshold, a raft group will need 16KiB memory. With 250k raft peer, it becomes 3.8GiB. Way smaller, but still a lot. Maybe for hibernated peers, you can just replace it with an empty vector, which allocates 0 memory.

BusyJay (approved):

gengliqi (approved):
LGTM

glorv:
@BusyJay sorry we have accidentally merged this PR while you comments were not fully resolved.

I tried to do a benchmark (in tikv with benchbot with  this https://github.com/tikv/tikv/pull/19594 code) yesterday. But I found there are some environment issues that caused the result not stable. I'm trying to re-propose a new round of bench and update the result assp.

glorv:
> Any benchmark result?

image

The benchmark is done with benchbot on this https://github.com/tikv/tikv/pull/19594 code. The result show that this is no noticeable performance impact after this change.

PR #593, opened 2026-09-16
raw_node: reset commit_since_index on lower snapshot arrival

dec256:
## Description
This PR resolves an assertion failure panic in `RawNode::ready()` during crash recovery.

### Root Cause
When a node restarts following an ungraceful shutdown, `commit_since_index` in `RawNode` is initialized from the node's local persisted state (e.g., `config.applied = 500`). If a leader subsequently sends a snapshot with a metadata index lower than `commit_since_index` (e.g., index `450`), `RawNode` hit an assertion check:

```rust
assert!(self.commit_since_index <= rd.snapshot.get_metadata().index);
```

Because a Raft snapshot represents a complete state machine truncation and log reset, enforcing monotonic growth of commit_since_index across process restarts and snapshot installation is invalid and causes unnecessary node panics.

### Fix
Removed the overly strict assert!. When a snapshot arrives, commit_since_index is updated directly to rd.snapshot.get_metadata().index, safely resetting index tracking for subsequent ready() cycles.

### Testing
Added regression test test_raw_node_snapshot_lower_commit_since_index in harness/tests/integration_cases/test_raw_node.rs.

Addresses Issue 19421

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved snapshot recovery when a restored snapshot has a lower index than the previously applied state.
  - Prevented unnecessary failures during recovery when snapshot and machine state differ.
  - Ensured the next ready state correctly reflects the restored snapshot’s index.


BusyJay:
Why a leader will send follower a snapshot that is before its applied index? Current flow should only replica logs in such case.

dec256:
when a lagging node recovers state from a peer whose snapshot lag trails the node's local persisted tracking

PR #594, opened 2026-09-17
fix: allow leader to stepdown gracefully and also step down as follower

syndbg:
## Implementation

2-node cluster, leader transfers away (hashicorp/raft does this before removing a peer). Transferee's election stalls or result never comes back. Old leader's own election timer fires MsgHup, wins again, undoes the transfer. Can loop forever: only 2 voters, nothing else to prefer.

Add Raft::set_campaign_paused / RawNode::set_campaign_paused: caller suppresses self-campaign after observing step-down, clears it once the reason ends. Voting, replication, apply unaffected. Reset on term change.

Does not fix a real network partition: if the two nodes can't talk at all, it's not safe to restore quorum, otherwise dual-leader situation might happen.
This only covers a lost vote or stalled election while both nodes are still reachable.

Step down as follower fix duplicates #584, kept in this PR since both close leader-give-up-power gaps: one forced (removed from voters), one voluntary (transfer).

## Theory

The Raft paper describes this in https://raft.github.io/raft.pdf#page=11.

```
The second issue is that the cluster leader may not be
part of the new configuration. In this case, the leader steps
down (returns to follower state) once it has committed the
Cnew log entry. This means that there will be a period of
time (while it is committingCnew) when the leader is managing a cluster that does not include itself; it replicates log
entries but does not count itself in majorities. The leader
transition occurs when Cnew is committed because this is
the first point when the new configuration can operate independently (it will always be possible to choose a leader
from Cnew). Before this point, it may be the case that only
a server from Cold can be elected leader.
```

But this campaigning is not really covered in the original paper. Diego Ongaro's (Raft author) PhD thesis (https://liqul.github.io/blog/assets/raft-thesis.pdf) covers the more practical side of why we need campaigning. Section `4.2.2 Removing the current leader` covers this and clearly states the issues with it.

```
If the existing leader is asked to remove itself from the cluster, it must step down at some point. One
straightforward approach is to use the leadership transfer extension described in Chapter 3: a leader
that is asked to remove itself would transfer its leadership to another server, which would then carry
out the membership change normally.
```

```
We initially developed a different approach for Raft, in which the existing leader carries out
the membership change to remove itself, then it steps down. This puts Raft in a somewhat awkward
mode of operation while the leader temporarily manages a configuration in which it is not a member.
We initially needed this approach for arbitrary configuration changes (see Section 4.3), where the old
configuration and the new configuration might not have any servers in common to which leadership
could be transferred. The same approach is also viable for systems that do not implement leadership
transfer.
```

```
In this approach, a leader that is removed from the configuration steps down once the Cnew entry
is committed. If the leader stepped down before this point, it might still time out and become leader
again, delaying progress. In an extreme case of removing the leader from a two-server cluster, the
server might even have to become leader again for the cluster to make progress; see Figure 4.6.
Thus, the leader waits until Cnew is committed to step down. This is the first point when the new
configuration can definitely operate without participation from the removed leader: it will always be
possible for the members of Cnew to elect a new leader from among themselves. After the removed
leader steps down, a server in Cnew will time out and win an election.
```

```
This small availability gap should be tolerable, since similar availability gaps arise when leaders fail.
This approach leads to two implications about decision-making that are not particularly harmful
but may be surprising. First, there will be a period of time (while it is committing Cnew) when a
leader can manage a cluster that does not include itself; it replicates log entries but does not count
itself in majorities. Second, a server that is not part of its own latest configuration should still start
new elections, as it might still be needed until the Cnew entry is committed (as in Figure 4.6). It does
not count its own vote in elections unless it is part of its latest configuration.
```


syndbg:
Note to reviewers, I've ran:
1. cargo test --all ✅ 
2. cargo bench --all -- --test ✅ 
3. cargo clippy --all --all-targets -- -D clippy::all
There were two pre-existing to this PR issues in `src/quorum/majority.rs`. I haven't fixed them in this PR, since they're off-topic.
4. cargo fmt --all -- --check ✅

syndbg:
Not sure who is actively maintaining it, but @BusyJay perhaps you're the right person to ping? 😉

